PR #7, opened 2015-12-09
Send 'X-Ajax-Browser-Auth' header on requests

oskarhane:
- To tell server that we want to be treated
  with special response headers.

### DIFF
--- a/app/scripts/services/RequestInterceptor.coffee
+++ b/app/scripts/services/RequestInterceptor.coffee
@@ -26,6 +26,7 @@ angular.module('neo4jApp.services')
     (AuthDataService) ->
       interceptor =
         request: (config) ->
+          config.headers['X-Ajax-Browser-Auth'] = true
           isLocalRequest = yes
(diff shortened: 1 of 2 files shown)


oskarhane:
Would @cleishm and @jakewins comment on this, please?  
It's this card: https://trello.com/c/cOaZgCt4

cleishm:
Is there any way to automatically test this kind of thing? The next step would be to change the server to send a known WWW-Auth header if it _doesnt_ see this X-AJAX-BROWSER-AUTH header. That'll break the browser if the header is ever missing, so a regression test would be nice.

oskarhane:
@cleishm The test in this PR will check that the header is there.  
All requests are intercepted and get the header added, and the test illustrates that this works when POST:ing Cypher queries.  

I'm not sure what other test you mean.

cleishm:
Ooops. I didn't notice that was a test. Just looked like more coffeescript to me ;-)

LGTM

PR #10, opened 2016-02-08
Handle modified rows and graph format

oskarhane:
- When returning a deleted node/rel, read metadata and remove from result rendering in viz for 'graph' format


oskarhane:
Here's the server change: https://github.com/neo4j/neo4j/pull/6354

PR #146, opened 2016-04-22
Remove ng-hide for bolt protocol repsonses

eve-bright:


pe4cey:
This exposes an issue with csv export when using bolt. The file treats values as a string rather than number/boolean

PR #152, opened 2016-05-04
On connect, exec command from server config

oskarhane:
Read `browser.post_connect_cmd` from server and
execute if it exists.

When picking first frame we still don't have access to this server config so that can't be a candidate on `pickFirstFrame()` but this will come after.


Mats-SX:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hsgRuZjQ&#x2F;81-have-browser-read-server-setting-browser-post-connect-cmd-for-initial-command"> Have Browser read server setting browser.post_connect_cmd for initial command</a>

pe4cey:
Do we want to clear all other frames if `browser.post_connect_cmd` has a value?

oskarhane:
I see it as an additional command and opening some frames and after connect (1-2 secs later) closing them could be confusing to users.

pe4cey:
👍 
LGTM

PR #153, opened 2016-05-04
Expose URL param API to populate editor

oskarhane:
This change exposes an API though URL params making it easier to publish links to prepopulate the editor.

The URL param API looks like this:
`cmd` - the command to use. Could be `play` or `help` or anything we decide to support.
`arg` - Zero or more of these args. 

Current implementation:   
 - `cmd` - This command will be prefixed by the current users config `cmdchar` (which defaults
to `:`). The special case here being `cmd=cypher`. In that case, the command is stripped and not prefixed by anything.  

- `arg` - They will be inserted in the order they appear in the URI and separated by a space.

This does **NOT** execute the command automatically. It adds content to the editor, the user must execute the command manually.

Examples:
Link to external guide: `http://localhost:7474/?cmd=play&arg=http://guides.neo4j.com/reco`
Link to cypher query: `http://localhost:7474/?cmd=cypher&arg=RETURN&arg=1`

### DIFF
--- a/app/scripts/controllers/Main.coffee
+++ b/app/scripts/controllers/Main.coffee
@@ -170,6 +170,17 @@ angular.module('neo4jApp.controllers')
 
   .run([
     '$rootScope'
-    ($scope) ->
+    'Utils'
+    'Settings'
+    'Editor'
+    ($scope, Utils, Settings, Editor) ->
       $scope.unauthorized = yes
+
+      if cmdParam = Utils.getUrlParam('cmd', window.location.href)
+        if cmdParam[0] is 'cypher'
+          cmdCommand = ''
+        else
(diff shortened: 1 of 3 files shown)


akollegger:
This is great stuff. Perhaps we can start slowly, though, by supporting just `play` until we've properly thought through the best way to handle security, warnings, and acknowledgements.

Mats-SX:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;x0vdx8ZP&#x2F;82-add-cmd-xxx-args-arg-url-parameters-for-neo4j-browser"> Add ?cmd=xxx&amp;args[]=arg url parameters for Neo4j browser</a>

oskarhane:
Supported commands for `cmd` are now limited to `play`

PR #191, opened 2016-08-08
Add link to show available procedures to unknown procedure message

pe4cey:


app/scripts/services/Frame.coffee
```diff
@@ -66,6 +66,8 @@ angular.module('neo4jApp.services')
             @templateUrl = intr.templateUrl
             @startTime = (new Date).getTime() unless @startTime
             @pinTime = 0
+            @procedureNotFound = (error) ->
+              return true if error && error.search('ProcedureNotFound') != -1
```
oskarhane: Details but to make it more coffee-scripty: `return yes if error and error.search('ProcedureNotFound') isnt -1`

PR #199, opened 2016-09-05
Bolt discovery

pe4cey:


app/scripts/init/statusCheck.coffee
```diff
@@ -27,6 +27,18 @@ angular.module('neo4jApp').run([
...
+#      .then(
+#        (r) ->
+#          r = r.data
+#          if r.bolt? and Settings.boltHost is ""
+#            $scope.boltHost = r.bolt.replace('bolt://', '')
+#          else
+#            $scope.boltHost = $location.host())
+#      .catch(
+#        (r)->
+#          $scope.boltHost = $location.host())
```
oskarhane: Why is the above commented out? Should it be removed?

PR #225, opened 2016-09-15
Lazy load ascii-table for frame responses

pe4cey:


app/scripts/controllers/CypherResult.coffee
```diff
@@ -63,19 +63,22 @@ angular.module('neo4jApp.controllers')
       $scope.tab = 'table' unless $scope.availableModes.indexOf($scope.tab) >= 0
 
     $scope.loadAscii = () ->
+      console.log 'loadAscii'
```
akollegger: Leftover debug?

jexp:
Could this go into 3.0.x ?

oskarhane:
Could you rebase this @pe4cey?

oskarhane:
Sure, we will back port this @jexp, just not right now.

pe4cey:
Rebased and pushed

oskarhane:
If text view is preselected it does not render the asciitable.

To reproduce:  

1. Run any query
2. Select text view
3. Run new query
4. Text view in new frame is empty

oskarhane:
Also, it seems like the coffeescript compiler on build servers is complaining about some indentation.

PR #226, opened 2016-09-16
Expand/collapse cypher result

pe4cey:
expand-collapse-frame


oskarhane:
Do you think it would be trivial to make this functionality a bit smarter so we can keep more results in the stream without performance being effected by style changes etc?  
i.e. don't ever call `graphView.update()` in collapsed state etc

PR #227, opened 2016-09-16
Add cluster role to db info drawer

eve-bright:


app/scripts/controllers/Auth.coffee
```diff
@@ -104,5 +105,5 @@ angular.module('neo4jApp.controllers')
 
       setPolicyMessage()
       connectIfNoAuthorizationRequired()
-
+      $scope.clusterRole = $rootScope.neo4j.clusterRole
```
oskarhane: Is this variable used anywhere?

app/scripts/controllers/Main.coffee
```diff
@@ -70,6 +70,13 @@ angular.module('neo4jApp.controllers')
             fetchJMX()
           ).then( ->
             featureCheck()
+          ).then( ->
+            if Features.usingCoreEdge
```
oskarhane: Where / how do we set this?

oskarhane: Found the answer to this  myself, looks good.

app/views/partials/drawer-database.jade
```diff
@@ -37,6 +37,9 @@
         a.value(exec-topic='server user list') server user list
   h5 Database
   ul
+    li.pair(ng-if="neo4j.clusterRole")
+      .key Role in cluster:
```
oskarhane: This wraps into two lines in the drawer, do you think it can be shorter so it stays on one line?

Issue #119, opened 2016-04-22
Syntax highlighting of "id"


oskarhane:
_From @jakewins on December 9, 2013 10:37_

From @akollegger @alexaverbuch 

Properties named "id" are syntax highlighted in black, whereas all other properties are in green

![neo4jbrowser__id__highlight](https://f.cloud.github.com/assets/53756/1502259/10e904d2-4898-11e3-849a-31a4dce86577.png)

_Copied from original issue: neo4j/neo4j#1665_

PR #237, opened 2016-09-21
remove dot check to allow syntax highlighting for properties

sendilkumarn:
#119 

I think the dot check is not needed here, we can by pass it. Since the only place where we use dots is to get node's property and of course we have to allow the syntax highlighting.

P.S still at early days with graph db.

### DIFF
--- a/app/scripts/codemirror-cypher.coffee
+++ b/app/scripts/codemirror-cypher.coffee
@@ -30,7 +30,7 @@ CodeMirror.defineMode "cypher", (config) ->
     if ch is "'"
       stream.match /.+?[']/
       return "string"
-    if /[{}\(\),\.;\[\]]/.test(ch)
+    if /[{}\(\),;\[\]]/.test(ch)
       curPunc = ch
       "node"
     else if ch is "/" and stream.eat("/")


sendilkumarn:
@oskarhane  is this pending ?

oskarhane:
Yes, we're focusing on features and regressions at this time so we haven't had time to revise this yet.  
We'll have  looking the coming days.

sendilkumarn:
Ping

eve-bright:
Thanks for contributing. We really appreciate you taking the time to try and improve the browser. 

Unfortunately we have decided that we want to use the official cypher highlighting from CodeMirror (https://github.com/codemirror/CodeMirror/tree/master/mode/cypher) as otherwise there is a duplication of effort and inconsistencies between the two implementations. 

This means that changes to our codemirror-cypher.js file will no longer be applicable. 

As a final note - potentially the change you made is too wide reaching. While the erroneous highlighting in the edge case when the property name is 'id' is resolved by the change, it also removes the useful highlighting of the dot in all other cases. 

We will try and update CodeMirror regularly so please do submit changes to the cypher mode file there.

PR #240, opened 2016-09-21
:qs - List and kill running queries

oskarhane:
`:qs` - Query Status

Works transparently in a cluster as well as on a single server.


app/scripts/services/Server.coffee
```diff
@@ -55,50 +55,82 @@ angular.module('neo4jApp.services')
...
-        post: (path = '', data) ->
-          path = Settings.host + path unless path.indexOf(Settings.host) is 0
+        post: (path = '', data, host = Settings.host) ->
+          path = host + path unless path.indexOf(host) is 0
           $http.post(path, data, httpOptions)
 
-        put: (path = '', data) ->
-          path = Settings.host + path unless path.indexOf(Settings.host) is 0
+        put: (path = '', data, host = Settings.host) ->
+          path = host + path unless path.indexOf(host) is 0
```
sendilkumarn: Why can't we move this line into a separate method. quite redundant.

PR #246, opened 2016-09-27
Add option to inspector for 'no icon'

eve-bright:


app/scripts/controllers/Inspector.coffee
```diff
@@ -107,6 +107,11 @@ angular.module('neo4jApp.controllers')
         $scope.$emit 'close.contextMenu'
         item.style = graphStyle.changeForSelector(item.style.selector, iconCode)
 
+#      $scope.selectNoIcon = (item) ->
+#        $scope.$emit 'close.contextMenu'
+#        item.style = graphStyle.changeForSelector(item.style.selector, {'icon-code': ''})
+
+
```
oskarhane: Could the above commented out code be removed?

eve-bright: Done, force pushed - as should not have been in original commit.

PR #249, opened 2016-09-29
Ability to connect to members in a cluster added to sysinfo

eve-bright:


eve-bright:
Do not merge this until Alistair's has merged his change.

eve-bright:
Should now be working with Alistair's change...

app/scripts/controllers/SysinfoController.coffee
```diff
@@ -103,5 +105,16 @@ angular.module('neo4jApp.controllers')
       $scope.autoRefresh = !$scope.autoRefresh
       refreshLater()
 
+    $scope.isCurrentConnection = (addresses) ->
+      if Settings.useBolt
+        boltHost = Settings.boltHost || $rootScope.boltHost
```
oskarhane: It would be better to reuse https://github.com/neo4j/neo4j-browser/blob/2.0/app/scripts/services/UtilityBolt.coffee#L246 to get the Bolt host, I think.

app/scripts/controllers/SysinfoController.coffee
```diff
@@ -103,5 +105,16 @@ angular.module('neo4jApp.controllers')
...
+        boltAddressForMember = Utils.getServerAddressByProtocol('bolt', addresses)[0]
+        return boltAddressForMember && boltAddressForMember.includes(boltHost)
+      else
+        Utils.getServerAddressByProtocol($location.protocol(), addresses).includes($location.host)
```
oskarhane: Same reuse here but for REST https://github.com/neo4j/neo4j-browser/blob/2.0/app/scripts/services/UtilityREST.coffee#L219

app/scripts/controllers/SysinfoController.coffee
```diff
@@ -103,5 +105,16 @@ angular.module('neo4jApp.controllers')
       $scope.autoRefresh = !$scope.autoRefresh
       refreshLater()
 
+    $scope.isCurrentConnection = (addresses) ->
+      if Settings.useBolt
+        boltHost = Settings.boltHost || $rootScope.boltHost
+        boltAddressForMember = Utils.getServerAddressByProtocol('bolt', addresses)[0]
+        return boltAddressForMember && boltAddressForMember.includes(boltHost)
```
oskarhane: This `includes` will always be true if you're running a cluster on the same host/ip and don't have a port specified in your bolt host settings.

oskar4j 2016-10-04 at 11 24 40

oskarhane:
Oh, I forgot, I get `Internal Server Error` when disabling Bolt and no C/E box is appearing on `:sysinfo`.
Not sure why.

app/scripts/controllers/SysinfoController.coffee
eve-bright: Hmm - this is a really odd case where host has an implicit port of 7687. Not sure how we should handle this - should we look for '7687' as a known default?

eve-bright: And now that the port has been registered, it's unlikely to change... http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=7687

PR #250, opened 2016-09-30
Update neo4j-driver to 1.1.0

pe4cey:


oskarhane:
On a fresh installation, the connect frame isn't shown.
If I manually show it, nothing happens when I enter a password and press "Connect".

pe4cey:
Looks like bower didn't pick up the correct version. Bower is looking for the latest `1.1.0` and the js driver being tagged with `1.1-M02`

PR #251, opened 2016-09-30
Set all db-info scope values to empty if disconnected

eve-bright:


oskarhane:
Could you rebase this @eve-bright?

eve-bright:
Rebased

oskarhane:
Since a database can have empty labels, types and props, would it make sense to just hide those parts and display a "Not connected" message instead, if the user isn't connected?

eve-bright:
I think we would still want to clear the values to avoid them accidentally appearing when logging in as a different user.

oskarhane:
Sure thing, the change you've done look good.

PR #254, opened 2016-10-03
Read and present query timings from result when using Bolt

oskarhane:
REST don't have access to this information so the old timer is kept there.

### Statusbar output examples

Read query:  
> Started streaming 25 records after 0 ms and completed after 1 ms.

Read query, no hits:  
> Completed after 0 ms.

Query without returns:  
> Completed after 0 ms.

Write query:  
> Set 250 properties, started streaming 250 records after 7 ms and completed after 19 ms.

Profile query (just showing the resultAvailableAfter timing):  
> Cypher version: CYPHER 3.1, planner: COST, runtime: INTERPRETED. 2 total db hits in 18 ms.

Raw / Code view:  
> Started streaming 25 records after 0 ms and completed after 1 ms.

### Frame output example

Updates, no returns:  
> Set 250 properties, statement completed in 4 ms.

No updates, no returns:  
> (no changes, no records)

### DIFF
--- a/app/scripts/controllers/CypherResult.coffee
+++ b/app/scripts/controllers/CypherResult.coffee
@@ -91,7 +91,7 @@ angular.module('neo4jApp.controllers')
     $scope.resultStatistics = (frame) ->
       if frame?.response
         updatesMessages = []
-        if frame.response.table?._response?.columns?.length
+        if frame.response and expectRecords frame
           updatesMessages = $scope.updatesStatistics frame
         rowsStatistics = $scope.returnedRowsStatistics frame
         messages = [].concat(updatesMessages, rowsStatistics)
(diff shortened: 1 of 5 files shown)


oskarhane:
Don't merge yet, waiting for clarifications on what language to use.

oskarhane:
Should be fine now, please review

app/scripts/controllers/CypherResult.coffee
```diff
@@ -120,7 +120,7 @@ angular.module('neo4jApp.controllers')
 
         message = "Cypher version: #{root.version}, planner: #{root.planner}, runtime: #{root.runtime}."
         if collectHits(root)
-          message += " #{collectHits(root)} total db hits in #{frame.response.responseTime} ms."
+          message += " #{collectHits(root)} total db hits in #{frame.response.timings.resultAvailableAfter || frame.response.timings.responseTime || 0} ms."
```
nigelsmall: "Hit" vs "hits" for 1 vs many

nigelsmall:
I would consider a special-case for _zero records returned_ that simply mentions one measurement (probably the last measurement). Something like:

> Completed after 10ms.

The reason being that "started streaming 0 records" doesn't really make sense. You can't stream zero records.

On a similar note, perhaps "finished" should then be replaced with "completed" throughout for consistency.

> Started streaming 12 records after 10 ms and completed after 20 ms.

WDYT?

PR #256, opened 2016-10-03
Warning propagation

pe4cey:


app/views/frame-add-new-user.jade
```diff
@@ -42,3 +42,5 @@ div(fullscreen)
         include partials/error-status-bar-detailed
         include partials/success-status-bar
         include partials/warning-status-bar
+        footer.tight.relative(ng-show="neo4j.clusterRole")
+          small Note: User management events do not propagate when running a core-edge cluster
```
akollegger: Small note: perhaps this can be simplified to "User management events do not propagate across a cluster."

app/views/frame-connect.jade
```diff
@@ -10,6 +10,8 @@ div(ng-controller="AuthCtrl")
                 h3 Connect to Neo4j
                 p.lead
                   | Database access requires an authenticated connection.
+                p.warn(ng-show="neo4j.clusterRole && failed_connection_attempts > 2")
+                  | Note: User management events do not propagate when running a core-edge cluster. Does your user exist on this server instance?
```
akollegger: I'd suggest avoiding direct mention of "core-edge" and simplifying the language. What about:

> Recall that user management events do not propagate across a cluster. Does your user exist here?

app/views/frame-user-admin.jade
```diff
@@ -14,3 +14,5 @@ div(fullscreen)
 
         include partials/error-status-bar-detailed
         include partials/success-status-bar
+        footer.tight.relative(ng-show="neo4j.clusterRole")
+          small Note: User management events do not propagate when running a core-edge cluster
```
akollegger: Final small note:

> User management events do not propagate across a cluster.

app/views/frame-connect.jade
pe4cey: I realised that we won't know any information about the server at this point. This message will appear any time a user fails to login more than three times. Switching the sentences around to be 'Does your user exist here? Recall that user management events do not propagate across a cluster' would cover the more likely case that the user doesn't exist at all as opposed to it being a propagation issue.

PR #258, opened 2016-10-04
Disconnect on authentication error

eve-bright:


oskarhane:
Great stuff, I just found one case where the behavior was unexpected.
Your branch against Neo4j 3.0.7 SNAPSHOT.

To reproduce:  

1. Enable Bolt
1. Connect with any user in a browser-1
1. Connect with same user in another browser-2
1. Change password with procedure in browser-2
1. Run any query in browser-1
1. You get message `No operations allowed until you send an INIT message successfully.`
1. When heartbeat comes along, you get disconnected and connect frame is shown

It works as expected when Bolt is disabled.  
It's like the failing EXPLAIN requests aren't picked up in Bolt mode.

PR #273, opened 2016-10-10
Updated dark theme styling

pe4cey:
Includes: 

- `:play` frames
- `:sysinfo`
- graph results frame
- fullscreen frames

### DIFF
--- a/app/styles/themes/dark.styl
+++ b/app/styles/themes/dark.styl
@@ -28,9 +28,14 @@ body.theme-dark
   .outline
     stroke: primary-background-color
     stroke-width: 1px
+  .relationship > .outline
+    stroke: primary-text-color
+    stroke-width: 1px
   .context-menu-item
     stroke: secondary-background-color
-
+  .relationship
+    text
+      fill: primary-text-color
   a
     color: color-link
     &:hover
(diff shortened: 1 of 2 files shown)


app/styles/themes/dark.styl
```diff
@@ -21,30 +21,47 @@ tertiary-background-hover = #292e3a
...
 color-link-hover = #ffffff
-
-
 body.theme-dark
   font: 13px 'Helvetica Neue', Helvetica, Arial, sans-serif
   background: primary-background-color
   color: primary-text
-
+  .outline
+    stroke: primary-background-color
```
sendilkumarn: this should be `primary-text-color` right?

pe4cey: For the `dark` theme it make more sense to go with a darker outline.

pe4cey:
I've just pushed a commit (after looking at @sendilkumarn's comments) to keep the stroke colour the same for both nodes and relationships

sendilkumarn:
👍  I was confused by the background and the below one [ which you changed just now ] 

Table transparent, is not quite good in dark mode. Can we have something slightly darker there, if possible?

PR #295, opened 2016-10-27
Fix bolt help typo

pe4cey:


oskarhane:
Should we make this change in 1.0 as well?

pe4cey:
Changed to 1.0

PR #306, opened 2016-11-01
Customizer provides a build switch for overlaying content and styles

akollegger:
The new `--custom` switch specifies a directory containing custom content and styling. Two examples are included, named "autumn" and "presentation".


oskarhane:
I made the change you pointed out @eve-bright

PR #308, opened 2016-11-02
Limit the number of meta items shown in database sidebar drawer

oskarhane:
Add links to show more / all

more


oskarhane:
Be careful on the forward merge.

PR #317, opened 2016-11-04
Close temporary driver when testing bolt connecitons

pe4cey:


oskarhane:
Fixes https://github.com/neo4j/neo4j-browser/issues/313

PR #322, opened 2016-11-14
Add bolt+routing

eve-bright:
Blocked until JS driver release with bolt+routing. Currently using self-built js file. 

cl: Add support for the `bolt+routing` protocol when in a Causal Cluster


oskarhane:
Looks great @eve-bright, just a few content points before I try out the actual functionality and the code.

- `s/Neo4J/Neo4j` in help texts
- Remove full stops / periods at the end from list items (`:help bolt routing`)
- Highlight the protocols and the words _direct_ and _routing_ in the first paragraph in `:help bolt routing` to help users associate protocols with phrases more early on 
- I think we should make the challenges with encryption with self signed certs and `bolt+routing` in web browsers more visible. We can link to it from `:help bolt routing` as a first step but we should also think about informing the user in other ways as well (might be out of scope for this PR though)

PR #350, opened 2017-03-19
renames 'String Store Size' to 'String Store', in sysinfo - Store Sizes table

alexaverbuch:
No other store-related entries in that column have "Size" as suffix.
This commit simply makes naming of the String Store entry consistent with other store entries.


oskarhane:
Thanks!

PR #358, opened 2017-04-06
Integrate zoom for d3 visualisation

eve-bright:


src/browser/modules/D3Visualization/components/Graph.jsx
```diff
@@ -23,13 +23,32 @@ import {createGraph, mapNodes, mapRelationships, getGraphStats} from '../mapper'
...
 import '../lib/visualization/index'
 import { dim } from 'browser-styles/constants'
+import {StyledZoomHolder, StyledSvgWrapper, StyledZoomButton} from './styled'
+import { ZoomInIcon, ZoomOutIcon } from 'browser-components/icons/Icons'
 
 export class GraphComponent extends Component {
 
+  constructor (props) {
+    super(props)
+    this.state = {}
```
irfannuri: We may define state at once like this?
this.state = {
      zoomInLimitReached: true,
      zoomOutLimitReached: false
    }

src/browser/modules/D3Visualization/components/Graph.jsx
```diff
@@ -23,13 +23,32 @@ import {createGraph, mapNodes, mapRelationships, getGraphStats} from '../mapper'
...
+    this.state.zoomInLimitReached = true
+    this.state.zoomOutLimitReached = false
+  }
+
   graphInit (el) {
     this.state.el = el
```
irfannuri: I suggest we may not keep reference to graph Dom element in state but directly in the component like: 
**this.el = el** (or maybe with a better name). Keeping in state also force us to call setState in render which is not desired.

PR #373, opened 2017-04-10
Give users feedback when trying to connect and it's slow

oskarhane:
Replaces connect button with the text 'Connecting...'.
Goes back to button if connection fails.


oskarhane:
Good find @irfannuri. Updated now.

PR #376, opened 2017-04-11
If Cypher error: Show ErrorView and hide everything else in Cypher frame

oskarhane:
- Add `shouldComponentUpdate` method to not re-render when not needed. The validation check could probably be better.
- Fixed: Viz button disappeared in old frames when executing a new query.

oskar4j 2017-04-11 at 15 19 36


pe4cey:
LGTM

PR #378, opened 2017-04-12
Update to latest Nevada version (for user testing)

joknelid:


joknelid:
Need to add some more changes so style editor works

PR #380, opened 2017-04-12
Add cypher stats to cypher results frame

pe4cey:


oskarhane:
Nice! That's really great. Still missing query statistics but I'll create a separate card for it.

PR #391, opened 2017-04-18
Enable HMR

oskarhane:
Re-enables hot module replacement when in dev mode.


FylmTM:
Works like a charm. Thanks!

PR #398, opened 2017-04-19
Add style command, send Grass style to Nevada

joknelid:


oskarhane:
I've added you to the contributor whitelist now so this should now build successfully.

oskarhane:
Passed the whitelist, but failed on linter now 😄

joknelid:
@oskarhane Sorry about that, I added "test" and "expect" to the globals for standard in package.js. Not sure why it doesn't fail on the other test files though, I might be missing something?

src/shared/modules/visualization/visualizationDuck.js
```diff
@@ -38,11 +38,13 @@ function updateLabelData (state, labels) {
 }
 
 function updateStyleData (state, styleData) {
-  return Object.assign({}, state, {styleData: styleData})
+  const clonedStyle = Object.assign({}, styleData)
```
joknelid: Oskar: This is what I was trying to talk about at standup. Do I need to do a deep copy to keep the redux state immutable immutable? This works for now in practice but you wont be able to do fancy stuff like rewind/replay.

oskarhane: I think you can do it and simplifying the updateStyleData function at the same time.

```
const updateStyleData = (state, styleData) => ({...state, styleData: {...styleData}})
```

oskarhane:
Change of plans. Closing this.

PR #399, opened 2017-04-20
Update queries frame - develop status bar

irfannuri:


src/browser/modules/Stream/Queries/QueriesFrame.jsx
```diff
@@ -18,20 +18,19 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { Component } from 'preact'
```
oskarhane: I don't like that your editor makes these formatting changes. It makes it hard to find the real changes you've made.
We have no rule for this and personally I find it a lot easier to read when it has spaces.

I'm going to merge this now, but please disable this feature in your IDE/editor.

PR #405, opened 2017-04-21
Add :style command missing popup that shows current style

joknelid:
This is the same stuff my last PR minus Nevada related code

### DIFF
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -22,17 +22,19 @@ import * as frames from 'shared/modules/stream/streamDuck'
 import { getHistory } from 'shared/modules/history/historyDuck'
 import { update as updateQueryResult } from 'shared/modules/requests/requestsDuck'
 import { getParams } from 'shared/modules/params/paramsDuck'
+import { updateGraphStyleData } from 'shared/modules/grass/grassDuck'
 import { cleanHtml } from 'services/remoteUtils'
 import { hostIsAllowed } from 'services/utils'
 import remote from 'services/remote'
 import { getServerConfig } from 'services/bolt/boltHelpers'
 import { handleServerCommand } from 'shared/modules/commands/helpers/server'
 import { handleCypherCommand } from 'shared/modules/commands/helpers/cypher'
-import { unknownCommand } from 'shared/modules/commands/commandsDuck'
+import { unknownCommand, showErrorMessage } from 'shared/modules/commands/commandsDuck'
 import { handleParamCommand, handleParamsCommand } from 'shared/modules/commands/helpers/params'
 import { handleGetConfigCommand, handleUpdateConfigCommand } from 'shared/modules/commands/helpers/config'
 import { CouldNotFetchRemoteGuideError, FetchURLError } from 'services/exceptions'
-import { parseHttpVerbCommand } from 'shared/modules/commands/helpers/http'
+import { parseHttpVerbCommand, isValidURL } from 'shared/modules/commands/helpers/http'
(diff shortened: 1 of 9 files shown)


src/shared/services/commandInterpreterHelper.js
```diff
@@ -178,6 +180,42 @@ const availableCommands = [{
...
   }
 }, {
+  name: 'style',
+  match: (cmd) => /^style(\s|$)/.test(cmd),
+  exec: function (action, cmdchar, put, store) {
+    const match = action.cmd.match(/:style\s*(\S.*)$/)
+    let param = match && match[1] ? match[1] : ''
+
+    if (param === '') {
+      // Todo: show popup
```
oskarhane: Instead of nothing, could you add to the user gets some feedback. Maybe something like: https://github.com/neo4j/neo4j-browser/blob/3.0/src/shared/services/commandInterpreterHelper.js#L184

joknelid: I just thought it was a bit overkill since there is a task in trello for adding the popup. But sure I can add it

joknelid:
I'm not sure I follow this update live stuff. It seems to update as it should for me. Are you talking about ":style reset" because that seems it doesn't update?

src/shared/services/commandInterpreterHelper.js
oskarhane: You know, just in case is doesn't make it as it should :) Thanks.

oskarhane:
Yes, you're correct @joknelid, it's the reset I meant. Sorry for not mentioning that.

joknelid:
Ok. then I'll fix that, thanks for clarifying :)

oskarhane:
I'm not sure if I'm ding something wrong, but when doing:

```
:style {"node":{"diameter":"50px","color":"#000000","border-color":"#9AA1AC","border-width":"2px","text-color-internal":"#FFFFFF","font-size":"10px"},"relationship":{"color":"#A5ABB6","shaft-width":"1px","font-size":"8px","padding":"3px","text-color-external":"#000000","text-color-internal":"#FFFFFF","caption":"<type>"},"node.Movie":{"color":"#68BDF6","border-color":"#5CA8DB","text-color-internal":"#FFFFFF","caption":"{title}"}}
```
I just get lot's of js errors:

oskar4j 2017-04-25 at 10 08 36

Same thing with 

```
:style https://oskarhane-dropshare-eu.s3-eu-central-1.amazonaws.com/style-nn1fSNZkMk/style.grass
```
That has the same contents. 

Do I use the wrong format @joknelid?

joknelid:
@oskarhane I'm not entirely sure which format the command should accept.

Currently, this PR only parses the css-like format like in this file: https://raw.githubusercontent.com/dhimmel/hetionet/383152be2c07f8f5427dd7069535c82b0c5e8d85/hetnet/neo4j/graphstyle.grass.
But it seems like the old browser accepts the json format you are using as well (although I can't seem to get anything to change in the styling by using the json format, everything looks exactly the same after I apply it)

I can't find any information in the documentation about exactly what style data the command should accept (https://neo4j.com/developer/guide-neo4j-browser/#_useful_commands_and_keyboard_shortcuts doesn't really mention it) But since you tried with json, it's probably what our customers will expect so I guess I should add it?

oskarhane:
If you want to do that, it'd be fantastic @joknelid  
It's the format we store it in the the browsers localstorage, that's where I got my example from.

oskarhane:
Could you rebase this please @joknelid

joknelid:
@oskarhane Should be rebased now

oskarhane:
Awesome, thanks so much for this!

PR #407, opened 2017-04-21
Integrate Cypher Editor

FylmTM:
Hi,

This pull request adds integration of Cypher Editor (based on CodeMirror) to browser.

Features:
- Improved syntax highlighting
- Error reporting
- Autocompletion (both reference and schema based)

Syntax colors are taken from editor. Other styles are bundled in browser directly.
Autocompletion entries have custom-created SVG icons (made by me).

### How to try it out

Code: github.com/neo4j-contrib/cypher-editor

1. Get `cypher-editor` support.
2. Link `cypher-editor-support` module with `yarn link`
3. Link `cypher-editor-support` into `cypher-codemirror` with `yarn link cypher-editor-support`
4. Build `cypher codemirror` with `yarn install && yarn build`
5. Link `cypher-codemirror` module with `yarn link`
6. Get this pull request locally for `neo4j-browser
7. Link `cypher-codemirror` into `neo4j-browser` with `yarn link cypher-codemirror`


src/shared/services/commandInterpreterHelper.js
```diff
@@ -195,5 +195,6 @@ const interpret = (cmd) => {
 }
 
 export default {
-  interpret
+  interpret,
+  commands: availableCommands.map((command) => command.name)
```
oskarhane: This is unfortunately not correct. The name is just for identification for internal use and `http`, `play-remote` amongst others aren't commands. 
We could add a `commands: ['get', 'post']`, `commands: ['play']` etc. for those who are commands. 
Would that help?

oskarhane:
Cool!
I haven't had time to try it out yet but I looked through the code.

Regarding where we want to have the syntax colors I'm not sure tbh. Let's think on that for a while.
Editor/cursor/gutter styles would be helpful to have in neo4j-browser I think.

Interesting with the performance issues.
Do you have more info on that? Is it because of the editor or unrelated?

FylmTM:
Hi @oskarhane,

The performance issue is related to animation itself. 
Animation is quite clunky even on my Macbook Pro.

Reason is that when width property is animated browser is forced to do repainting and relayouting.
Because it needs to change shape and position of surrounding elements.

Also width animation is performed on cpu, AFAIK.

It can be fixed by switching to something like translateX. But in this particular case it looks awful.

In the end my thinking is that no animation is better than slow animation.

src/shared/services/commandInterpreterHelper.js
FylmTM: Yeap. In general I thing the solution should somehow retrieve actual commands from the application, rather than having them hard-coded.

On top of that it would be handy to have subcomads as well (e.g. For help). We have theoretical possibility to implement autocompletion for nested commands.

We did something like that for procedure call output parameters.

oskarhane:
Can you rebase this again @FylmTM? Sorry, we are pushing lot's of code now.

FylmTM:
@oskarhane 

I have rebased on top of latest 3.0
Also squashed all changes into single commit.
And updated PR to reflect that this is now more-or-less finished integration.

PR #409, opened 2017-04-22
Parse and style server topic attributes and prepend icons

irfannuri:


src/browser/components/Directives.jsx
```diff
@@ -34,6 +34,11 @@ const directives = [{
     return `:play ${elem.getAttribute('play-topic')}`
   }
 }, {
+  selector: '[server-topic]',
+  valueExtractor: (elem) => {
+    return `:play ${elem.getAttribute('server-topic')}`
```
oskarhane: The command should be `:server`, not `:play`.
Ref: https://github.com/neo4j/neo4j-browser/blob/2.0/app/scripts/directives/serverTopic.coffee

irfannuri: @oskarhane updated, Thanks!

PR #411, opened 2017-04-24
Set and read version from package.json and pom.xml

oskarhane:
Actions required on release is now to:

```
npm version 3.0.0-RC2
```

That command changes version in `package.json` and creates and tags a new commit with the new version.
Pushing to git repo is not part of this npm script.

This also updates `pom.xml` so we can remove a build step.

### DIFF
--- a/src/browser/modules/Sidebar/About.jsx
+++ b/src/browser/modules/Sidebar/About.jsx
@@ -18,6 +18,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { version } from '../../../../package.json'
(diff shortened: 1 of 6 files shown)


src/browser/modules/Sidebar/About.jsx
```diff
@@ -18,6 +18,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { version } from '../../../../package.json'
```
oskarhane: Good idea. 
Maybe we should update the version in the `pom.xml` file as well when doing this?

oskarhane:
Updated @pe4cey

oskarhane:
btw, we need to make sure that we build and package _after_ the version is set.

PR #412, opened 2017-04-24
Retrieve and present update statistics from bolt result

irfannuri:
- Retrieve query update statistics from bolt result (from summary.counters)
- Prepend update stats to streaming stats in Status Bar
- Present update stats in Body in Table & Ascii View if there are no rows

### DIFF
--- a/src/browser/modules/Stream/CypherFrame.jsx
+++ b/src/browser/modules/Stream/CypherFrame.jsx
@@ -137,7 +137,7 @@ class CypherFrame extends Component {
             this.changeView(viewTypes.TABLE)
           }}><TableIcon /></CypherFrameButton>
         </Visible>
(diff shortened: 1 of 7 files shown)


pe4cey:
The only issue I noticed was with the types of cypher view available. 

Here we return three views (rows, ascii & code) with the rows and ascii rendering the same thing:
views

In 2.0 we show two views (rows & code):
screen shot 2017-04-24 at 19 09 22

irfannuri:
Thanks @pe4cey, updated.

oskarhane:
Great!

One thing that I found that needs to be addressed:

1. Executing `RETURN rand()` should show a frame with ascii table view available. Just it's just rows + code atm.

irfannuri:
@oskarhane done. I had to rebase btw

PR #421, opened 2017-04-26
Fetch and store server config and info in redux

oskarhane:
Store in `meta` reducer and update on every meta data update, since server con be restarted with new settings without the browser being reloaded.

The data stored is not used for anything with this PR, but will in future features.

oskar4j 2017-04-26 at 21 01 17

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaDuck.test.js
+++ b/src/shared/modules/dbMeta/dbMetaDuck.test.js
@@ -46,28 +46,62 @@ describe('updating metadata', () => {
     expect(nextState.properties).toEqual([{val: 'prop1', context: 'mycontext'}, {val: 'prop2', context: 'mycontext'}])
   })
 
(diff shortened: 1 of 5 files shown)


src/shared/modules/dbMeta/dbMetaDuck.test.js
```diff
@@ -36,7 +36,7 @@ describe('updating metadata', () => {
       get: (val) => { return ['prop1', 'prop2'] }
     }
     const action = {
-      type: meta.UPDATE,
+      type: meta.UPDATE_META,
       meta: {records: [ returnedLabels, returnedRelationshipTypes, returnedProperies ]},
       context: 'mycontext'
     }
```
eve-bright: Should we have tests for the new actions too?

oskarhane: Added tests @eve-bright

PR #423, opened 2017-04-26
Help: case-insensitive guide lookup

FylmTM:
Small fix.

Now all variations will work properly:
```
:help match
:help MATCH
:help Match
```


oskarhane:
Thanks!

PR #432, opened 2017-04-27
Fix table styling

eve-bright:


eve-bright:
screen shot 2017-04-27 at 18 28 07
screen shot 2017-04-27 at 18 28 13

eve-bright:
Screenshots are old browser and new browser with this change. I can see two problems with them:

- There seems to be some styling of the contents in the old browser (removing quotes, (empty) instead of {} etc) which we are not doing in the new browser.

- The table is too wide to fit on the screen and overflow is hidden so there is no way to see the last few columns. This appears to have been an issue before so not a regression. 

-

eve-bright:
Additionally this affects normal cypher queries too. In the old browser, we had a different view for this. I'm not sure what the plan is with the NBN.

eve-bright:
Finally, the changes to the user stuff, I think make a very small improvement but I think we could do with styling these frames more at some point.

PR #440, opened 2017-04-28
In validate connection use return 0 so that we don't hit read issues

eve-bright:


eve-bright:
Maybe it would better to use driver.onError and driver.onCompleted?

oskarhane:
@eve-bright to get immediate answer if credentials are ok we have to run a query, I believe.

eve-bright:
@oskarhane Ah ok, cool - Think return 0 should be fine then. It successfully rejected invalid credentials but allowed all valid users.

PR #445, opened 2017-04-28
tweaked sidebar icon colors and state changes

akollegger:
Colors are now used to indicate status. They're brighter again, but not quite as neon as in prior releases. Oh, except that the Neo4j icon has a pulsing green blue which only indicates that we are awesome.


oskarhane:
Nice!

PR #451, opened 2017-04-29
Auto-pair brackets & auto trigger completion

FylmTM:


src/browser/modules/Editor/Editor.jsx
```diff
@@ -233,14 +253,17 @@ export class Editor extends Component {
         'Cmd-Up': this.historyPrev.bind(this),
         'Ctrl-Up': this.historyPrev.bind(this),
         'Cmd-Down': this.historyNext.bind(this),
-        'Ctrl-Down': this.historyNext.bind(this)
```
oskarhane: Why was this removed?

oskarhane:
Really great, I like this behavior @FylmTM.

I behaves a bit weird in a few cases though, I'm not sure if it's hard to fix.
I think the ac activation by character needs some context knowledge not to confuse users.

See these examples:

In function, variables should be on top:  
oskar4j 2017-04-29 at 20 28 23  

Only functions are suggested here. I'm not sure why:  
oskar4j 2017-04-29 at 20 28 37  

In a string characters shouldn't trigger ac:  
oskar4j 2017-04-29 at 20 33 36

src/browser/modules/Editor/Editor.jsx
FylmTM: Accident. Will get it back.

FylmTM:
@oskarhane Yeap. This is something that should be solved at cypher-editor-support.
We are continuously tuning AC over a time.
I will fix it for your examples.

FylmTM:
@oskarhane I got back removed line.
But do not merge this PR yet.

I might be rewriting brackets auto-pair, to properly support them for strings.

FylmTM:
@oskarhane Now you can check out change again.

Better to do it with latest version of cypher-editor (check out manually and link to neo4j-browser).

Now:
- Auto-pair `()[]{}''""`. It requires fix from cypher-editor to work properly for strings
- Autocompletion in string disabled
- Variables are now prioritized and only ones that are not in expression are AC

PR #452, opened 2017-04-30
Sync signin bar

irfannuri:
This PR aims to improve Browser Sync usage. If User has signed-in to Sync before they are kindly reminded to sign-in again when they open Neo4j Browser. User can cancel the notification to opt-out from sync. Any Sync sign-in afterwards would let user opt-in to Sync again. Find related screens attached.
browser_sync_3


src/browser/modules/Main/Main.jsx
```diff
@@ -57,6 +58,7 @@ const Main = (props) => {
           Connection to server lost. Reconnecting...
         </WarningBanner>
       </Visible>
+      <SyncReminderBar />
```
oskarhane: All the other ones are named `*Banner`, could you rename this to align to that please?

irfannuri: @oskarhane Updated.

oskarhane:
A few more points, on the visual part.

- The green color seems a bit too light compared to the screenshot.
  Ask Peter or Sophia for the right color code.

- The **not** on the banner is bold in this PR, but not on the design spec.

- I also think it should it should overflow with ellipsis, it looks cut off on small displays:  
  oskar4j 2017-05-01 at 21 50 23

irfannuri:
@oskarhane how about that?
screen shot 2017-05-01 at 22 27 18

oskarhane:
@irfannuri Sorry, I think all banners should be just one line.

irfannuri:
@oskarhane discussed with Sophia and this is the best so far.

PR #453, opened 2017-04-30
Scroll to the top result when stream is updated

pe4cey:


src/browser/modules/Stream/Stream.jsx
```diff
@@ -20,6 +20,7 @@
 
 import { connect } from 'preact-redux'
 import { Component } from 'preact'
+import ReactDOM from 'react-dom'
```
FylmTM: Is react-dom in package.json?
Looks like that it's not there.

FylmTM:
Just a side question - we can use react-dom?

I thought adding it to enable testing via enzyme. But after conversation I found out that React was abandoned due to license & patents issues.
And unfortuntately react-dom is under same BSD license.

oskarhane:
No, we should not use react-dom but I suspect that this works because it's gets aliased to preact-compat that has the findDomNode.
But it would be better to be more clear and import the aliased method directly.

pe4cey:
Now preactified

PR #458, opened 2017-05-01
Add z-index to drawer so that content is hidden in safari

pe4cey:
Issue: 
safari-meh

Fixed:
safari-a


oskarhane:
There's a 3-4px dark strip to the right of the drawer in safari. Could you check if it's easy to solve in this PR as well @pe4cey?

oskarhane:
I'll merge this and create a card for the part I mentioned above.

PR #459, opened 2017-05-01
Fix json download icon color

pe4cey:
- Extracts out a `<DefaultA />` styled component to be used for other links in the future


oskarhane:
JSON? 😸

PR #461, opened 2017-05-01
UDC

oskarhane:
Report approved user interactions.

Only mount report script when server config `browser.allow_outgoing_connections` is read and set to `true`.
Start reporting events only when user connects to Neo4j Browser Sync (and by doing so, approves these reports), unless the user opted out by `:config shouldReportUdc: false`.

I've made the udcDuck totally passive, just listening on what's going on in Redux. If we want to capture more events, we should send these events through Redux and keep this a consumer.


oskarhane:
Rebased @pe4cey

PR #463, opened 2017-05-01
Fix docs links

jjaderberg:
* Version docs links.
* Use literal version: `3.2` (breaks now, works for GA, update when can lookup version from server).
* Use HTTPS.
* Use full path for chunked pages.


oskarhane:
Fantastic, thanks so much @jjaderberg

PR #464, opened 2017-05-01
Empty :help => :help help

oskarhane:
Fallback on `:help help` if user doesn't specify a help topic.


pe4cey:
The links at the bottom of the help topic are not working. This is more of an issue with the `:help help` command but this exposes it too.

PR #470, opened 2017-05-02
Display integers as integers in response

oskarhane:
To not have precision loss we did `val.toString()` on `neo4j.integers` when displaying the response.
This did have the implication that integers got quoted, just as any string, which is very misleading.

The changes in this PR changes that.

The change is made that we now implement our own version of `JSON.stringify` that should be compatible, but with one important change: our version takes a modifier function that have first access to all values passing through and can choose to modify the value, or just skip and let other handle it.

Before:
oskar4j 2017-05-02 at 15 16 05

After:
oskar4j 2017-05-02 at 15 15 02

### DIFF
--- a/src/browser/modules/Stream/CypherFrame.jsx
+++ b/src/browser/modules/Stream/CypherFrame.jsx
@@ -54,6 +54,7 @@ class CypherFrame extends Component {
 
   componentWillReceiveProps (nextProps) {
     let rows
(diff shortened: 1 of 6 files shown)


pe4cey:
Seems to remove write stats from the frame body
screen shot 2017-05-02 at 20 19 58

oskarhane:
Good catch @pe4cey. Fixed now.

PR #480, opened 2017-05-03
Update Cypher Editor version

FylmTM:


pe4cey:
@FylmTM would you be able to rebase please

PR #483, opened 2017-05-04
Try to fetch guide remotely if not found locally

oskarhane:
When trying to play a local guide: `:play reco` and it isn't found, an attempt to find it on the url `browser.remote_content_hostname_whitelist + name` is made.

`browser.remote_content_hostname_whitelist` is a server configuration list of hosts (or urls:s) that the browser can fetch remote content from.

The default value for `browser.remote_content_hostname_whitelist` is `localhost,guides.neo4j.com`, so the above play command would resolve and display the reco guide hosted on https://localhost/reco, then on http://localhost/reco, then on https://guides.neo4j.com/reco and finally on http://guides.neo4j.com/reco.
It stops after the first successful fetch.

The steps followed are:

- Try locally. If found, display it
- Try remotely, by appending the guide name to each of server config list of `browser.remote_content_hostname_whitelist`
- Iterate over whitelist and if guide is found, display it and stop trying to fetch guide from the rest of the list
- If error
  - show 'Not found' frame

Some updates to on how we use `isomorphic-fetch` are added as well, so errors are thrown when we hit a non 200 status from a request.

oskar4j 2017-05-04 at 11 10 23

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
       "afterEach",
       "neo",
       "FileReader",
-      "Blob"
+      "Blob",
+      "fetch"
     ]
   },
   "repository": {},
--- a/src/browser/modules/Stream/PlayFrame.jsx
+++ b/src/browser/modules/Stream/PlayFrame.jsx
@@ -18,36 +18,74 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { Component } from 'preact'
+import { withBus } from 'preact-suber'
+import { fetchGuideFromWhitelistAction } from 'shared/modules/commands/commandsDuck'
+
 import Guides from '../Guides/Guides'
 import * as html from '../Guides/html'
 import FrameTemplate from './FrameTemplate'
 import { splitStringOnFirst } from 'services/commandUtils'
+import ErrorsView from './Views/ErrorsView'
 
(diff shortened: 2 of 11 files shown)


oskarhane:
@pe4cey Updated

oskarhane:
It'd be great to have your view on this feature and implementation @akollegger.

Should the config be moved to server config to have it centralized?
Suggestions for a better name on the configuration option would be appreciated. 🎈

akollegger:
Well, wdyt about two things:

1. use the server whitelist as a "search path" for remote content
2a. making the `browser.*` configuration served up by the server as static browser content
2b. holding onto this PR until the server-side changes have been made for 2a

oskarhane:
I like the idea of using the whitelist as a search path @akollegger.
I'm not sure I understand what you mean by 2a.

Also, there might be the case where the browser isn't connected to a server (we might not even know the server url) and what should happen in that case? That's why I made it a client side config in the first place.
Maybe just a fallback til the guides url in that case?

oskarhane:
This is now ready for review again @pe4cey (and more comments if you have any @akollegger).
Since I've changed a lot I've updated the expected behavior in the PR:s description.

oskarhane:
Rebased @pe4cey

PR #484, opened 2017-05-04
Fix parsing of kernel version

pe4cey:
It allows us to parse the version with/without the version being at the end of the returned string


src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -220,7 +220,7 @@ export const dbMetaEpic = (some$, store) =>
         .do((res) => {
           if (!res) return
           const [ kvObj, storeObj, edObj ] = res
-          const versionMatch = kvObj.KernelVersion.match(/version:\s([^,]+),/)
+          const versionMatch = kvObj.KernelVersion.match(/version:\s([^,$]+)/)
           const version = versionMatch[1]
```
oskarhane: Just to be sure, maybe null check `versionMatch` before grabbing `[1]`.

PR #486, opened 2017-05-04
Fetch internal relationships between displayed nodes (Auto complete)

irfannuri:
A setting has been added, 'Connect result nodes', replacing 'Auto complete' of the previous version.
If this setting is checked, after a cypher query result is retrieved, a second query is executed to fetch relationships between result nodes.

Movie dataset query response before:
oskar4j 2017-05-04 at 19 58 08

and after:
oskar4j 2017-05-04 at 19 53 43

### DIFF
--- a/src/browser/modules/Sidebar/Settings.jsx
+++ b/src/browser/modules/Sidebar/Settings.jsx
@@ -81,6 +81,13 @@ const visualSettings =
             displayName: 'Max Rows',
             tooltip: "Max number of rows to render in 'Rows' result view"
           }
+        },
+        {
+          autoComplete: {
+            displayName: 'Auto Complete',
+            tooltip: 'Automatically fetch relationships between displayed nodes',
+            type: 'checkbox'
+          }
         }
       ]
     }
--- a/src/browser/modules/Sidebar/Settings.jsx
+++ b/src/browser/modules/Sidebar/Settings.jsx
@@ -85,7 +85,7 @@ const visualSettings =
         {
           autoComplete: {
             displayName: 'Connect result nodes',
-            tooltip: 'Automatically fetch relationships between displayed nodes',
+            tooltip: 'If this is checked, after a cypher query result is retrieved, a second query is executed to fetch relationships between result nodes.',
             type: 'checkbox'
           }
         }
--- a/src/browser/modules/Sidebar/Settings.jsx
+++ b/src/browser/modules/Sidebar/Settings.jsx
@@ -84,7 +84,7 @@ const visualSettings =
         },
         {
           autoComplete: {
-            displayName: 'Auto Complete',
+            displayName: 'Connect result nodes',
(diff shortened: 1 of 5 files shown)


oskarhane:
I'm getting an uncaught error thrown.

1. Insert movie dataset
1. Click `Movies` in sidebar, or execute `MATCH (n:Movie) RETURN n LIMIT 25`.

Error is shown.
oskar4j 2017-05-04 at 17 52 24

src/browser/modules/Sidebar/Settings.jsx
```diff
@@ -81,6 +81,13 @@ const visualSettings =
             displayName: 'Max Rows',
             tooltip: "Max number of rows to render in 'Rows' result view"
           }
+        },
+        {
+          autoComplete: {
+            displayName: 'Auto Complete',
```
oskarhane: I think the suggestion on the example: `Connect result nodes` is a better description that makes more sense.

src/shared/modules/settings/settingsDuck.js
```diff
@@ -38,6 +38,7 @@ export const getMaxNeighbours = (state) => state[NAME].maxNeighbours || initialS
 export const getMaxRows = (state) => state[NAME].maxRows || initialState.maxRows
 export const getInitialNodeDisplay = (state) => state[NAME].initialNodeDisplay || initialState.initialNodeDisplay
 export const shouldReportUdc = (state) => state[NAME].shouldReportUdc !== false
+export const autoComplete = (state) => state[NAME].autoComplete !== false
```
oskarhane: Selectors are usually named `get*` or `should*`.

oskarhane:
The complete error I get is:

```
code: "Neo.ClientError.Statement.TypeError"
message:"Expected entity id to be an integral value"
```

src/browser/modules/Stream/Visualization.jsx
```diff
@@ -84,7 +102,31 @@ export class Visualization extends Component {
...
+          }
+        }
+      )
+    })
+  }
+
+  getInternalRelationships (existingNodeIds, newNodeIds) {
+    existingNodeIds = existingNodeIds.concat(newNodeIds)
+    const query = `
+            MATCH (a)-[r]-(b) WHERE id(a) IN {existingNodeIds}
```
oskarhane: The "new" syntax for params is `$existingNodeIds` instead of `{existingNodeIds}`.
We should probably use that.

irfannuri:
@oskarhane @pe4cey updated. I could not reproduce the error, somehow my driver version is not updated, but able to reproduce it on mark's comp. It seems we have to convert array items to neo4j int types in order to bind it to param. Instead we removed param binding and used string interpolation.

oskarhane:
Fixed. Using same query as in old codebase now since it's been looked at by AT and MH.

PR #488, opened 2017-05-04
Fix issue where stream would scroll to top on frame rerun

pe4cey:
Only scrolls to the top of the stream when a new frame is added


oskarhane:
Works in this case now, great!
But found related bug: closing a frame scrolls to top. New PR or try to fix in this one?

oskarhane:
So I guess a new card for the second issue I mentioned @irfannuri? :)

PR #489, opened 2017-05-04
Only scroll to top of stream when frame is added and toggle is on

pe4cey:
The scroll to top will only kick in once a frame is added (this handles rerun/frame removal issues we've had)

Added toggle to the sidebar
screen shot 2017-05-04 at 20 04 32


src/shared/modules/settings/settingsDuck.js
```diff
@@ -37,6 +37,7 @@ export const getBrowserSyncConfig = (state) => {
 export const getMaxNeighbours = (state) => state[NAME].maxNeighbours || initialState.maxNeighbours
 export const getMaxRows = (state) => state[NAME].maxRows || initialState.maxRows
 export const getInitialNodeDisplay = (state) => state[NAME].initialNodeDisplay || initialState.initialNodeDisplay
+export const getScrollToTop = (state) => state[NAME].scrollToTop || initialState.scrollToTop
```
irfannuri: @pe4cey I believe default value should be true to preserve previous behaviour. Because the setting is new and so gets false by default

pe4cey: Well caught @irfannuri. PR updated

PR #491, opened 2017-05-07
Fix the error occurs on auto-complete when initial Node Display threshold is exceeded

irfannuri:
This bug happens when graph object cuts the nodes off when Initial Node Display limit is exceed. Visualisation component is unaware of the final displayed nodes and tries to find relations between all retrieved nodes. Fixed by passing graph object to Visualisation component thus it request relations between only displayed nodes.


oskarhane:
LGTM

pe4cey:
Ditto

PR #494, opened 2017-05-07
Sort meta items alphabetically when fetched

oskarhane:


irfannuri:
@oskarhane this is crazy but it does not work in safari unless you rewrite it like that ' a.val < b.val ? -1 : ( a.val > b.val ? 1 : 0)'
Somehow chrome and Firefox is fine with returning it boolean but Safari is not. I guess it expects an exact return value described here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort?v=control

oskarhane:
Good catch @irfannuri!
Would you be able to make that change?
I'm not working this morning.

oskarhane:
If we're doing a release this morning and want this in, that is.

irfannuri:
Sure @oskarhane

PR #498, opened 2017-05-08
Read and act on server config retain_connection_credentials

oskarhane:
On connect (and at the regular interval) read server config `browser.retain_connection_credentials`.
If set to a falsy value, remove persisted connection credentials from web browsers localstorage. 
Keep in memory to be able to use in a Causal clustered environment where new connections might be opened to run a query on every cluster member (`:queries` for example).

Show persist status on the `:server connect` frame.

Connect when set to false:
oskar4j 2017-05-08 at 15 05 26

and true (default):
oskar4j 2017-05-08 at 15 05 02


src/shared/modules/connections/connectionsDuck.js
```diff
@@ -134,6 +141,11 @@ const mergeConnectionHelper = (state, connection) => {
   )
 }
 
+// Local vars
+let memoryUsername = ''
```
irfannuri: @oskarhane I am a bit concerned about having those as local variable, although the current lifecycle of duck files are Ok with that. We may have them in redux store but not synced with local storage.

PR #506, opened 2017-05-10
Remove unused jquery dependency

pe4cey:
Not used by any component


oskarhane:
How do we know it's unused?

pe4cey:
But how do we know that it is?

pe4cey:
It's seems it was added when we brought in react-swipe (and it doesn't require jquery). A quick it of manual test seems to show all is ok without this dependency

PR #510, opened 2017-05-12
Add a new button to edit node.

yezuoxian:
Edit button for edit node's properties. Click the edit button will explode all the properties of the node. Then click the new node can edit the node's property.
I implement this function by adding nodeEdit() in GraphEventHandler.js


akollegger:
Hi @yezuoxian !

This is pretty incredible work, considering how un-documented the code base is! Graph editing is something we've thought about for a while. Unfortunately, we've stopped development of the "neod3" library, looking forward to a complete rewrite of graph visualization similar to what we just did with  Neo4j Browser itself. So we won't merge your changes because the code is all about to be replaced. 

I'm sorry about that. 

Best,
Andreas

PR #516, opened 2017-05-15
Style table view and just show relevant information

irfannuri:
Table view lacked the styling and also exposed all properties of all objects. With this PR it is styled considering array and Neo4j graph objects; Node, Relationship and Path. Below are before / after screenshots.
screen shot 2017-05-15 at 17 29 18
screen shot 2017-05-15 at 17 28 29

### DIFF
--- a/src/browser/components/DataTables.jsx
+++ b/src/browser/components/DataTables.jsx
@@ -42,33 +42,15 @@ export const StyledTd = styled.td`
   padding: 0 16px
 `
 
-export const StyledJsonTable = styled.table`
+export const StyledJsonPre = styled.pre`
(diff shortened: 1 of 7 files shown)


oskarhane:
I actually think the JSON view is appreciated by users, especially since it hints at what type the values have. That's the feedback I've heard at least.
What about a mix, keep objects in grey boxes, but serialize them with something like `JSON.stringify(obj, null, 2).

irfannuri:
updated @oskarhane @pe4cey. I hope that pleases all.

screen shot 2017-05-17 at 11 05 50

oskarhane:
Yes, I like that!

PR #531, opened 2017-05-18
Fix for collected graph items now showing up in viz

### LINKED ISSUE #528: Collected nodes aren't shown in graph results view (3.2)
Using the movies graph, the following query will show the movie nodes in the graph result view in version 3.1.4 (and below), but will not show the collected movie nodes in 3.2.

```
match (p:Person{name:'Keanu Reeves'})-[:ACTED_IN]->(m:Movie)
return p, collect(m)
```

oskarhane:
When extracting graph items from result only the first level was checked. This fixes that so we recursively check the complete response and extract all items.

Consider the query (on the movie dataset):

```
MATCH (p:Person{name:'Keanu Reeves'})-[:ACTED_IN]->(m:Movie)
RETURN p, collect(m)
```

Before:
oskar4j 2017-05-18 at 10 19 23

After:
oskar4j 2017-05-18 at 10 18 21

Fixes #528

### DIFF
--- a/__mocks__/neo4j.js
+++ b/__mocks__/neo4j.js
@@ -18,4 +18,22 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-module.exports = {}
+var out = {
+  v1: {
+    types: {
+      Node: function Node (id) {
+        this.identity = id
+      },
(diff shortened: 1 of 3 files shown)


oskarhane:
Updated @pe4cey

PR #540, opened 2017-05-22
Add slider to text view statusbar

oskarhane:
To adjust the max column width of the response table.
Useful when copying the result to paste in into some other medium.

oskar4j 2017-05-22 at 15 52 27

### DIFF
--- a/src/browser/modules/Stream/Views/AsciiView.jsx
+++ b/src/browser/modules/Stream/Views/AsciiView.jsx
@@ -18,15 +18,66 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { Component } from 'preact'
 import asciitable from 'ascii-data-table'
+import { debounce } from 'services/utils'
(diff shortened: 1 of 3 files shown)


src/browser/modules/Stream/Views/AsciiView.jsx
```diff
@@ -18,15 +18,66 @@
...
+  componentWillReceiveProps (nextProps) {
+    this.ensureContentWidth(nextProps.rows)
+  }
+  ensureContentWidth (rows) {
+    if (rows.length !== this.props.rows.length) {
+      this.setMaxSliderWidth(asciitable.maxColumnWidth(rows))
+    }
+  }
+  componentWillMount () {
+    this.debouncedMaxColWidthChanged = debounce(this.maxColWidthChanged, 25, this)
```
irfannuri: why use debounce here @oskarhane ? It seems to work better without it, what do I miss?

oskarhane: With large result sets it's really expensive to recalculate the table 10 times per second and will probably crash the web browser.

irfannuri: Oh fair enough, thanks.

PR #541, opened 2017-05-22
Listen for config maxFrames and enforce that limit in stream

oskarhane:
Will not affect pinned frames.
I separated the epic tests from reducer tests because I think it gets messy with both of them in the same file.


pe4cey:
Should this PR cover the visual side of changing the `maxFrames` in settings?

oskarhane:
If we broaden the scope if it, sure 😃

oskarhane:
Added to sidebar @pe4cey

pe4cey:
👍

PR #548, opened 2017-05-23
Add support for handling multiple commands from browser.post_connect_cmd

oskarhane:
They are assumed to be splitted by semicolon.

I had to refactor some in the tests files to have the tests running as expected. There seems to be a bug in either `configureMockStore` or `createEpicMiddleware`.

### DIFF
--- a/src/shared/services/commandUtils.test.js
+++ b/src/shared/services/commandUtils.test.js
@@ -115,4 +115,23 @@ describe('commandutils', () => {
       expect(obj.str + ': ' + utils.isCypherCommand(obj.str, ':')).toEqual(obj.str + ': ' + obj.expect)
     })
   })
+  test('extractPostConnectCommandsFromServerConfig should split and return an array of commands', () => {
+    // Given
+    const testStrs = [
+      { str: '', expect: undefined },
+      { str: ';;;;;;;;', expect: undefined },
+      { str: ':play cypher', expect: [':play cypher'] },
+      { str: '    :play cypher     ', expect: [':play cypher'] },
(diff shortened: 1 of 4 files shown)


src/shared/services/commandUtils.test.js
```diff
@@ -115,4 +115,23 @@ describe('commandutils', () => {
...
+      { str: '', expect: undefined },
+      { str: ';;;;;;;;', expect: undefined },
+      { str: ':play cypher', expect: [':play cypher'] },
+      { str: '    :play cypher     ', expect: [':play cypher'] },
+      { str: ':play cypher;', expect: [':play cypher'] },
+      { str: ':play cypher ;', expect: [':play cypher'] },
+      { str: ';:play cypher;', expect: [':play cypher'] },
+      { str: ':play cypher; :param x: 1', expect: [':play cypher', ':param x: 1'] },
+      { str: 'RETURN 1; RETURN 3; :play start', expect: ['RETURN 1', 'RETURN 3', ':play start'] }
+    ]
```
pe4cey: Splitting on `;` means that if you have a cypher query containing `;` then that will also split. 
Test case: `{ str: 'MATCH (n: {foo: "bar;"}) RETURN n', expect: ['MATCH (n: {foo: "bar;"}) RETURN n'] }`

oskarhane:
Added basic semicolons-within-quotes detection @pe4cey.

TheRealMarcusChiu:
breaking commands on `;` renders the following to fail
`browser.post_connect_cmd=style node { diameter: 50px; color: #FFFFFF }`

PR #551, opened 2017-05-24
Remove unused action CLEAR_IN_CONTEXT

oskarhane:


irfannuri:
@oskarhane shall we also remove https://github.com/oskarhane/neo4j-browser/blob/3.0-remove-unused-code/src/shared/modules/stream/streamDuck.js#L32

oskarhane:
Good point, I totally missed that.

oskarhane:
Done @irfannuri

PR #569, opened 2017-05-30
Tooling upgrades (standard)

oskarhane:
The important upgrade in this PR is the upgrade of `standard` and the code changes related to that upgrade.
We did not follow the latest standardJS rules so some changes was needed. Mostly rejection of promises should be an `Error` but we used it to pass whatever.

The sensitive changes are around the rejection of bolt connections in the cases where we want to have access to the `driver` object in the `.catch` clause to force another query on the connection (in the case of forced pw change and cluster member queries (mot sure why we do this though).

All tests pass now and I've manually done some testing and it all _seems_ to be working.


pe4cey:
When attempting to connect I get an error in the connectionsDuck
```
subscribeToResult.js:42 Uncaught TypeError: Invalid attempt to destructure non-iterable instance
    at app-5292fcf….js:49372
    at connectionsDuck.js:268
    at <anonymous>
```

oskarhane:
Good catch, fixed now @pe4cey

PR #576, opened 2017-05-31
Ensure recent view is applied to Cypher frame only once

irfannuri:
This PR ensures recent view at the stream state is applied to Cypher frame only once, which avoids resetting of all open frames to the recent view


oskarhane:
I'm going to close this one since a larger refactoring is in progress. If we need a fast fix for a patch release before the refactoring is completed, we can bring this back.

Issue #587, opened 2017-06-14
Bug - Boolean values do not display in table view in Neo4j 3.2.1


InverseFalcon:
Reproducible both for boolean values:
```
RETURN 1 = 1
```
and booleans within a collection:
```
RETURN [true, false, true]
```
but display is fine when inside a map:
```
RETURN {test1:true, test2:[true, false, true]}
```

PR #588, opened 2017-06-15
Refactor CypherFrame

### LINKED ISSUE #583: Export CSV downloads only first 1000 rows of results in v3.2.1 unlike earlier versions in which they downloaded the whole result.

oskarhane:
This is an attempt to make the very imperative CypherFrame more declarative and easier to reason about.

View and their data should be contained in their own files.
All state definitions should live in their separate view.
Parent state could be used to communicate between different views and view parts.
A view is now responsible for it's own main view and statusbar. 

Makse sure that new views use `shouldComponentUpdate` to opt out of updating for props changes that aren't used.

New component `<Display>` is introduced which just hides child component if the `if` statement is false. In contrast to our `<Render>` compoent which removes children from DOM.

Utility function `shallowEquals` is also introduced for faster comparisons on objects.

This PR is not "job done" with the refactoring, this is just the first step.
Next step will probably be to let each view provide their own FrameTitlebar.

Aslo fixes #583, and #587


oskarhane:
I think this one can be looked at now @pe4cey

irfannuri:
Frame does not show any view after recentview is set to 'PLAN' and the result is either graph or table.
To reproduce: Run 'PROFILE MATCH (n:Movie) RETURN n LIMIT 25' and switch to any other view and then back to PLAN. Run any cypher without plan like 'MATCH (n:Movie) RETURN n LIMIT 25'

It may be resolved by adding also Plan condition here : https://github.com/oskarhane/neo4j-browser/blob/4f3b2ae4ec96e8f9ecbac854b21fe03c7784e56a/src/browser/modules/Stream/CypherFrame/helpers.js#L112, but I am not sure if that would be the right solution.

src/browser/modules/Stream/CypherFrame/index.jsx
```diff
@@ -0,0 +1,231 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {
+      openView: undefined,
+      exportData: null
+    }
+  }
+  changeView (view) {
+    this.setState({openView: view})
+    if (this.props.onRecentViewChanged) {
```
irfannuri: This onRecentViewChanged hook was useful in CypherFrame because it did not had access to redux or bus, but since now this is a connected component I think we can directly dispatch this action from this component.

src/browser/modules/Stream/CypherFrame/TableView.jsx
```diff
@@ -0,0 +1,147 @@
...
+const toTable = (records) => records && records.length ? bolt.recordsToTableArray(records) : undefined
+
+const renderCell = (entry) => {
+  if (Array.isArray(entry)) {
+    const children = entry.map((item, index) => <span>{renderCell(item)}{index === entry.length - 1 ? null : ', '}</span>)
+    return <span>[{children}]</span>
+  } else if (typeof entry === 'object') {
+    return renderObject(entry)
+  } else {
+    return entry
```
irfannuri: We need a `toString()` or such here to render cypher calls like `return true`

src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
```diff
@@ -179,4 +162,13 @@ const mapDispatchToProps = (dispatch) => {
   }
 }
 
-export default withBus(connect(mapStateToProps, mapDispatchToProps)(Visualization))
+export const VisualizationConnectedBus = withBus(connect(mapStateToProps, mapDispatchToProps)(Visualization))
+
+export class VisualizationStatusbar extends Component {
```
irfannuri: Is this needed?

src/browser/modules/Stream/CypherFrame/index.jsx
oskarhane: Nice catch. This should definitely live in the CypherFrame rather than the Stream.

src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
oskarhane: Not at all.

src/browser/modules/Stream/CypherFrame/TableView.jsx
oskarhane: There's a separate card for this, but maybe `JSON.stringify` will do?
oskar4j 2017-06-27 at 09 11 36

oskarhane:
I've addressed your comments now @irfannuri

PR #596, opened 2017-06-20
Automatic code formatting with prettier

oskarhane:
The [prettier](https://github.com/prettier/prettier) project is spreading like a virus, for good reasons.
I wanted to try it on our codebase (but still keep the format within the standardJS rules).
The `standard --fix` command is very limited in what it can do.
Prettier takes the AST and walks backwards to create code.
Read this blogpost for more info: http://jlongster.com/A-Prettier-Formatter

I've created a precommit hook so the code is auto formatted on every commit. 
You don't have to think about it. Write code in what ever format you want, it'll be fixed into standardJS compatible format that's easy to read. No more linter errors, ever.

In my opinion the code is **a lot** easier to read formatted this way.

Thoughts on this @pe4cey @irfannuri?

### DIFF
--- a/src/browser/modules/Editor/Editor.jsx
+++ b/src/browser/modules/Editor/Editor.jsx
@@ -22,11 +22,17 @@
 import { Component } from 'preact'
 import { connect } from 'preact-redux'
 import { withBus } from 'preact-suber'
-import { executeCommand, executeSystemCommand } from 'shared/modules/commands/commandsDuck'
+import {
+  executeCommand,
+  executeSystemCommand
+} from 'shared/modules/commands/commandsDuck'
 import * as favorites from 'shared/modules/favorites/favoritesDuck'
 import { SET_CONTENT, FOCUS, EXPAND } from 'shared/modules/editor/editorDuck'
 import { getHistory } from 'shared/modules/history/historyDuck'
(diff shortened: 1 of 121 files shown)


src/browser/modules/Editor/Editor.jsx
```diff
@@ -321,22 +354,22 @@ export class Editor extends Component {
...
+            hoverIcon='&quot;\58&quot;'
+            icon='&quot;\73&quot;'
           />
           <EditorButton
             onClick={() => this.clearEditor()}
             disabled={this.getEditorValue().length < 1}
             title='Clear'
-            hoverIcon='"\e005"'
-            icon='"\5e"'
+            hoverIcon='&quot;\e005&quot;'
```
oskarhane: I suspected this would be a problem, but it seems to be working fine.

pe4cey: That's certainly more uglier than before. It does however feel more explicit about how the string is going to be encoded

oskarhane:
The reason we need standard to fix things after prettier is only for adding spaces before parens on function declarations.
I just realized we can't have standard in the precommit, if it fails we can't commit code 😂
So it seems impossible to use standard and prettier together in an automatic way at the moment.
Otoh, if we want to move to a format other than standard, it's super easy.
I just like the way prettier formats the code and therefore it would be nice to use it, but not necessary.

PR #600, opened 2017-06-22
Honor url whitelist when fetching remote grass files

pe4cey:
The url param when loading remote grass using `:style` needs to be present in the whitelist returned by the neo4j server

### DIFF
--- a/src/shared/modules/commands/helpers/grass.test.js
+++ b/src/shared/modules/commands/helpers/grass.test.js
@@ -36,18 +36,14 @@ describe('Grass remote fetch', () => {
   test('should not fetch from url not in the whitelist', () => {
     const whitelist = 'foo'
     const urlNotInWhitelist = 'bar'
-    return fetchRemoteGrass(urlNotInWhitelist, whitelist)
-    .catch((error) => {
-      expect(error).not.toBe(null)
+    return expect(fetchRemoteGrass(urlNotInWhitelist, whitelist)).rejects.toMatchObject({
+      message: 'Hostname is not allowed according to server whitelist'
     })
   })
   test('should fetch from url in the whitelist', () => {
     const whitelist = 'foo'
     const urlInWhitelist = 'foo'
-    return fetchRemoteGrass(urlInWhitelist, whitelist)
-    .then((res) => {
-      expect(res).toBe(urlInWhitelist)
-    })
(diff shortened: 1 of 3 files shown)


src/shared/modules/commands/helpers/grass.test.js
```diff
@@ -18,60 +18,85 @@
...
+    get: (url) => {
+      return new Promise((resolve, reject) => {
+        resolve(url)
+      })
+    }
+  }
+})
+
+describe('Grass remote fetch', () => {
+  test('should not fetch from url not in the whitelist', () => {
```
oskarhane: This test won't work if the function doesn't reject. Try changing `urlNotInWhitelist = 'foo'`.

Since Jest 20 you can return promises and expect it to reject. There's one missing peace though, you cannot match on error message. The workaround is to match an object and looks like this:

```
return expect(fetchRemoteGrass(urlNotInWhitelist, whitelist)).rejects.toMatchObject({
  message: 'Hostname is not allowed according to server whitelist'
})
```

https://facebook.github.io/jest/docs/expect.html#resolves and https://github.com/facebook/jest/issues/3601

src/shared/modules/commands/helpers/grass.test.js
```diff
@@ -18,60 +18,85 @@
...
+    .catch((error) => {
+      expect(error).not.toBe(null)
+    })
+  })
+  test('should fetch from url in the whitelist', () => {
+    const whitelist = 'foo'
+    const urlInWhitelist = 'foo'
+    return fetchRemoteGrass(urlInWhitelist, whitelist)
+    .then((res) => {
+      expect(res).toBe(urlInWhitelist)
```
oskarhane: Same as above, but using `.resolves`: https://facebook.github.io/jest/docs/expect.html#resolves

src/shared/services/commandInterpreterHelper.js
```diff
@@ -207,7 +207,7 @@ const availableCommands = [{
       if (!param.startsWith('http')) {
         param = 'http://' + param
       }
-      remote.get(param)
+      fetchRemoteGrass(param)
```
oskarhane: Should the whitelist be passed as a param here?

pe4cey:
@oskarhane updated

oskarhane:
lgtm

PR #601, opened 2017-06-26
Update ducks so 'hydrated' is not stored as part of redux state

pe4cey:
- Stops {hydrated:true} appearing in ':params'

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaDuck.js
+++ b/src/shared/modules/dbMeta/dbMetaDuck.js
@@ -20,7 +20,6 @@
 
 import Rx from 'rxjs/Rx'
 import bolt from 'services/bolt/bolt'
-import { hydrate } from 'services/duckUtils'
 import { getJmxValues, getServerConfig, isConfigValFalsy } from 'services/bolt/boltHelpers'
 import {
   CONNECTED_STATE,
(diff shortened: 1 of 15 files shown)


src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -155,7 +156,7 @@ const initialState = {
  * Reducer
  */
 export default function meta (state = initialState, action) {
-  state = hydrate(initialState, state)
+  state = (hydrated) ? state : { ...initialState, ...state }
```
oskarhane: I think `hydrated` be set to `true` here, as it is in all other duck files.

oskarhane:
I know we discussed this earlier and we weren't happy with the solution.
wdyt about ditching the local `hydrated` flag and use the [`APP_START`](https://github.com/neo4j/neo4j-browser/blob/3.0/src/browser/index.jsx#L77) action to hydrate with?
It's supposed to be the first "real" action dispatched.

Then the reducers would just have this:

```javascript
import { APP_START } from 'shared/modules/app/appDuck'

if (action.type === APP_START) {
  state = { ...initialState, ...state }
}
```

If you happy with this approach @pe4cey, I can make the changes.

pe4cey:
Some thing like that makes sense. Do we ever have a situation where `APP_START` wouldn't fire but we would want the new state to be hydrated?

oskarhane:
Not that I can think of. 
In tests maybe?
wdyt, should I make the changes @pe4cey?

PR #609, opened 2017-06-29
Don't add repeated executed commands to history

oskarhane:
Fix regression where repeated commands were added to history multiple times.

oskar4j 2017-06-29 at 07 34 48


ikwattro:
Thanks amigos !

PR #612, opened 2017-07-04
Upgrade firebase dependency to 4.1.x

pe4cey:


oskarhane:
It does not seem to push changes after this update @pe4cey. It just keeps saying `Synced X minutes ago`.
But, after updating it even further, to `firebase@4.2.0` it works again. 
So I'll push the extra bump and merge this.

PR #619, opened 2017-08-01
Fix display of Neo4j integers in Cypher result frames

### LINKED ISSUE #618: Bug - Browser displays incorrect long values in 3.2.1 and 3.2.2
This is related to observations from [this Stack Overflow question](https://stackoverflow.com/questions/44943239/neo4j-using-long-as-id/44949146#44949146), which can be used to replicate the problem.

The browser is using the javascript driver, which likely means [proper numeric value handling](https://github.com/neo4j/neo4j-javascript-driver#read-integers) isn't being performed, resulting in displaying wrong long values in results displays for the affected versions when the high portion of a 64-bit value is non-zero.

Requesting a fix ASAP, we shouldn't leave this bug in for another version.

oskarhane:
This regression was introduced with the cypher frame refactoring.
Rather than filling up `boltMappings` with more functions I decided to place the mappings and transforms close to where they are used.
Opens for code duplication, but makes the code more safe to make changes to.

oskarhane-mbpt 2017-08-01 at 10 55 35

Fixes #618

### DIFF
--- a/__mocks__/neo4j.js
+++ b/__mocks__/neo4j.js
@@ -18,10 +18,18 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+function integerFn (val) {
+  this.val = val
+}
+integerFn.prototype.toString = function () {
+  return this.val.toString()
+}
+
 var out = {
   v1: {
-    isInt: function () {
-      return false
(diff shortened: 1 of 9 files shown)


pe4cey:
LGTM

PR #622, opened 2017-08-07
Add cypress based e2e tests

pe4cey:
### Running the suite

1. Have a clean instance of Neo4j running using the default config (more specifically the bolt port to be `:7687`)

2. A version of the Neo4j Browser running on `http://localhost:8080` (running `yarn start` from the root of the project will be good enough)

3. Run the test suite with `yarn run e2e`

For now we just use default cypress settings so at the end of a run there will be a video output for debugging issues

### How we find parts of the dom

The components which are used in the tests have `data-test-id` attributes attached in the rendered html. In the test we use these to select from the rendered dom `cy.get('[data-test-id="foo"')`

We make use of the `babel-plugin-jsx-remove-data-test-id` library to  strip out these when building for production.


.babelrc
```diff
@@ -6,5 +6,12 @@
     ["transform-react-jsx", {
       "pragma": "h"
     }]
-  ]
+  ],
+  "env": {
```
oskarhane: Dies this work? I think the babel config for builds are in the web pack config

oskarhane:
Can we not have it globally installed, but rather a part of our dev dependencies?
And also have a npm/yarn script to invoke it.

I'd suggest that we name the test folder something else than `cypress`. How about `e2e_test` or similar?

pe4cey:
`cypress` is the default folder that is created and ran against. I've just pushed a change to bring `cypress-cli` as a dev dependency. 

Take a look at https://github.com/neo4j/neo4j-browser/pull/622/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R11 we have to override each config since the folder rename ('cypress' -> 'e2e_tests')

oskarhane:
This is so cool. I love the video output as well.

PR #624, opened 2017-08-09
Add ability to download a query plan as png

pe4cey:
In the cypher result frame for `EXPLAN||PROFILE` query we display the `plan` view that has an interactive query plan element. 

This PR adds a button in the titlebar to download the query plan as a png file.

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/PlanView.jsx
+++ b/src/browser/modules/Stream/CypherFrame/PlanView.jsx
@@ -59,10 +59,9 @@ export class PlanView extends Component {
         })
     }
     this.ensureToggleExpand(props)
+    props.assignVisElement && props.assignVisElement(this.el, this.plan)
   }
   shouldComponentUpdate (props, state) {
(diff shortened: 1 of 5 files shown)


src/browser/modules/Stream/CypherFrame/PlanView.jsx
```diff
@@ -49,6 +49,8 @@ export class PlanView extends Component {
     this.ensureToggleExpand(props)
   }
   shouldComponentUpdate (props, state) {
+    this.props.assignVisElement && this.props.assignVisElement(this.el, this.plan)
```
oskarhane: I don't think this is the lifecycle method to use for this. `componentDidUpdate` or `componentWillReceiveProps` would be more appropriate imo.

pe4cey:
Updated

oskarhane:
If you generate multiple frames with plans in them, export as PNG will download the latest plan, even from older frames.

PR #630, opened 2017-08-18
Sort keys when inspecting nodes

jgaskins:
I have to hunt down where my properties are in the inspector a lot. I assume others do, too. Cmd-F isn't the most friendly tool in this view, so this should help improve that experience. :-)

Since there are usually very few properties shown in the inspector (a few dozen?) the performance impact should be negligible.


oskarhane:
Thanks for the contribution @jgaskins! 
I like the idea of this since it improves the user experience.

Would be able to sign our [CLA](https://neo4j.com/developer/cla/) so we can accept your PR?

jgaskins:
Sure, all signed!

oskarhane:
Thank you so much for taking the time to improve Neo4j Browser, much appreciated!

PR #634, opened 2017-08-21
Fix wrong query generated for user graph items named ‘*’

oskarhane:
Only give first items special query (match any)


src/browser/modules/DatabaseInfo/MetaItems.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+  // When
+  return mount(RelationshipItems)
+    .withProps({
+      relationshipTypes,
+      totalNumItems: relationshipTypes.length,
+      onItemClick
+    })
+    .then(wrapper => {
+      const items = wrapper.find(`[data-test-id='sidebarMetaItem']`)
```
pe4cey: I completely forgot that `data-test-id` can be used in our component testing (and not just `e2e`)

PR #636, opened 2017-08-24
Added basic auth support in connectURL

Metaur:
To have an ability to connect browser to neo4j database automatically, basic auth support for `connectURL` query parameter was added. 

Along that, some code refactoring was done to improve the work with url parsing.


oskarhane:
Thanks @Metaur,
Would be able to sign our [CLA](https://neo4j.com/developer/cla/) so we can accept your PR?

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -88,7 +92,13 @@ export const discoveryOnStartupEpic = (some$, store) => {
     })
     .mergeMap(action => {
       if (action.forceURL) {
-        store.dispatch(updateDiscoveryConnection({ host: action.forceURL }))
+        store.dispatch(
+          updateDiscoveryConnection({
+            username: action.username,
+            password: action.password,
```
oskarhane: I think we only want to update`username` and `password` if they are present in the action object. 
If not, they should be left as they were.

src/shared/services/utils.test.js
```diff
@@ -182,25 +186,35 @@ describe('utils', () => {
       port: '8080',
       pathname: '/index.html',
       search: '',
-      hash: ''
+      hash: '',
+      username: '',
+      password: ''
     })
-    expect(utils.getUrlInfo('guides.neo4j.com')).toEqual({
```
oskarhane: Can we keep this test (without protocol) since the browser will sometimes receive hostnames only from the server (when all protocols for the hostname are meant to be whitelisted)?

Issue #522, opened 2017-05-16
WebSocket connection to 'ws://localhost:7687/' failed


netpoe:
I installed neo4j with Docker and opened the browser window on `localhost:7474`

Here's my `docker-compose.yml` file:

```bash
neo4j:
  container_name: neo4j
  image: neo4j
  ports:
    - 7474:7474
    - 7867:7867
  volumes:
    - ~/Projects/Docker/neo4j/data:/data
    - ./conf/neo4j.conf:/conf/neo4j.conf
    - ~/Projects/Docker/neo4j/logs:/logs
```

* `latest` version is 3.2.0 from docker hub

I've added a `neo4j.conf` file with these settings: 

```bash
#*****************************************************************
# Neo4j configuration
#
# For more details and a complete list of settings, please see
# https://neo4j.com/docs/operations-manual/current/reference/configuration-settings/
#*****************************************************************

# The name of the database to mount
#dbms.active_database=graph.db

# Paths of directories in the installation.
dbms.directories.data=/data
#dbms.directories.plugins=plugins
#dbms.directories.certificates=certificates
dbms.directories.logs=/logs
#dbms.directories.lib=lib
#dbms.directories.run=run

# This setting constrains all `LOAD CSV` import files to be under the `import` directory. Remove or comment it out to
# allow files to be loaded from anywhere in the filesystem; this introduces possible security problems. See the
# `LOAD CSV` section of the manual for details.
dbms.directories.import=import

# Whether requests to Neo4j are authenticated.
# To disable authentication, uncomment this line
dbms.security.auth_enabled=false

# Enable this to be able to upgrade a store from an older version.
#dbms.allow_format_migration=true

# Java Heap Size: by default the Java heap size is dynamically
# calculated based on available system resources.
# Uncomment these lines to set specific initial and maximum
# heap size.
dbms.memory.heap.initial_size=512M
dbms.memory.heap.max_size=512M

# The amount of memory to use for mapping the store files, in bytes (or
# kilobytes with the 'k' suffix, megabytes with 'm' and gigabytes with 'g').
# If Neo4j is running on a dedicated server, then it is generally recommended
# to leave about 2-4 gigabytes for the operating system, give the JVM enough
# heap to hold all your transaction state and query context, and then leave the
# rest for the page cache.
# The default page cache memory assumes the machine is dedicated to running
# Neo4j, and is heuristically set to 50% of RAM minus the max Java heap size.
dbms.memory.pagecache.size=512M

[...]

PR #650, opened 2017-09-27
Pass encryption mode when force changing password

oskarhane:
This fixes an issue when using officially signed certificates on a newly installed Neo4j when setting up initial password.
For some reason browsers are ok with opening an unsecure websocket on a "secure" website if it's a manually trusted self signed certificate. But I won't open an unsecure websocket from a website on a officially signed cert.
That made this quite hard to find, but https://github.com/neo4j/neo4j-browser/issues/522#issuecomment-332502755 pointed me in the right direction.

This can be verified (that an unsecure websocket is opened) by looking in the network tab in chrome on a newly installed neo4j when setting the initial password.
With these changes, all opened websockets are secure.

Changelog: Fix setting initial password over officially trusted TLS certificates

### DIFF
--- a/src/shared/modules/cypher/cypherDuck.js
+++ b/src/shared/modules/cypher/cypherDuck.js
@@ -23,6 +23,7 @@ import Rx from 'rxjs'
 import bolt from 'services/bolt/bolt'
 import { getActiveConnectionData } from 'shared/modules/connections/connectionsDuck'
 import { getCausalClusterAddresses } from './queriesProcedureHelper'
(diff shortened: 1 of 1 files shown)


Sankarb475:
Hi @oskarhane I am getting the same error due to the same reason you have mentioned over here. But I didn't really get how to resolve this issue.
can you please in short elaborate what to do exactly?

oskarhane:
@Sankarb475 Please open an issue describing your exact problem.
This PR fixes an issue when tls mode is set to `REQUIRED`.

PR #652, opened 2017-10-04
Enable webworkers when using a non-tls connection

pe4cey:
- Allows non-blocking rendering when executing a long running cypher query

### DIFF
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -59,6 +59,7 @@ import {
   parseGrass
 } from 'shared/modules/commands/helpers/grass'
 import { shouldUseCypherThread } from 'shared/modules/settings/settingsDuck'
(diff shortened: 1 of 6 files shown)


src/shared/services/commandInterpreterHelper.js
```diff
@@ -133,11 +134,18 @@ const availableCommands = [
     match: cmd => /^cypher$/.test(cmd),
     exec: (action, cmdchar, put, store) => {
       const state = store.getState()
+
+      console.log(
```
oskarhane: *cough*

oskarhane: *cough*

pe4cey:
Updated

oskarhane:
Closing this in favour of https://github.com/neo4j/neo4j-browser/pull/670
I think we hit https://bugs.chromium.org/p/chromium/issues/detail?id=760708 when it failed.

PR #655, opened 2017-10-05
Remap query plan arguments to the root object

pe4cey:
- Re-enable skipped boltMapping tests

Before:
all-node-scan-old

After:
all-nodes-scan-new

### DIFF
--- a/src/shared/services/bolt/boltMappings.js
+++ b/src/shared/services/bolt/boltMappings.js
@@ -95,24 +95,16 @@ export function extractPlan (result, calculateTotalDbHits = false) {
     const boltPlanToRESTPlanShared = plan => {
       return {
         operatorType: plan.operatorType,
-        LegacyExpression: plan.arguments.LegacyExpression,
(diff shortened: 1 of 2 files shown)


oskarhane:
Something strange happens with the total dbHits.
On `3.0`: `Cypher version: CYPHER 3.3, planner: COST, runtime: INTERPRETED. 1443 total db hits in 2 ms.`
On this PR: `Cypher version: CYPHER 3.3, planner: COST, runtime: INTERPRETED. 00025313200506677172 total db hits in 4 ms`.

Query (with movie dataset): 

```
PROFILE MATCH (n) WHERE EXISTS(n.name) RETURN DISTINCT "node" as entity, n.name AS name LIMIT 25 UNION ALL MATCH ()-[r]-() WHERE EXISTS(r.name) RETURN DISTINCT "relationship" AS entity, r.name AS name LIMIT 25
```

pe4cey:
@oskarhane updated

PR #661, opened 2017-10-11
Don't remap escape sequences in cypher value output

pe4cey:
Fixes https://github.com/neo4j/neo4j/issues/10034


oskarhane:
LGTM, but I think it would be a good idea to add tests for `\` as well, to show that this change fixes the issue.

PR #662, opened 2017-10-12
Add button to download vis as svg

pe4cey:
The download button is visible in cypher results frame titlebar when the visualization tab is selected

scg-export


oskarhane:
When I try to export a query plan as SVG I get:

oskarhane-mbpt 2017-10-16 at 08 45 05

This query on the movie dataset. PNG works fine.

```
PROFILE MATCH p=()-[r:DIRECTED]->() RETURN p LIMIT 25
```

pe4cey:
@oskarhane updated

PR #664, opened 2017-10-13
Integrate with desktop API

oskarhane:
## New component
We added the `DesktopIntegration` component which reads initial context and sets up a listener for events that the desktop att emits.
To react on events happening you attach props to the component in the form of `EVENT_NAME` -> `onEventName`.   
Example on react on `DATABASE_STARTED` -> 

```javascript
<DesktopIntegration 
  integrationPoint={window.neo4jDesktopApi} 
  onDatabaseStarted={(event, newContext, oldConext) => alert(event.type)}  
/>
```

To get the initial context on app load, you attach the 

```javascript
onMount={(activeGraph, boltCredentials, fullContext) => alert(activeGraph.id)}
```

## Features
In web env, no changes are expected.
In desktop env, this should be the behaviour:

**When browser is opened when a graph already is running:**
  - Get host + port from desktop api and populate the connection fields with those.
  - In the case of username and pw in browsers local storage not being correct, show `:server connect` and the host should be what the active graph has.
  - In the case of username and pw being correct, connect and show start frame.

**When browser is open and connected and graph is shut down:**
  - Show connect banner right away
  - Empty metadata
  - Keep username and pw in memory

**When browser is open but not connected to a graph and a graph starts:**
  - Get credentials (host and encryption mode) from desktop api and re-use stored username and pw and try to connect.
  - If successful
    - Show `:server switch success` frame and inform about the connection switch.
    - Update metadata
  - If unsuccessful
    - Show `:server switch fail` frame and inform about the failure
    - User clicks `:server connect` link to manually enter creds.

## Worth mentions

- To our `mount` function in `testUtils.js` we can now pass a second param with props to get to test the `componentDidMount` lifecycle method.
- Environment detection added. In `appDuck.js` there are now selectors to check what environment the browser lives in. Useful for code branching.
- The plan is that the functions in `DesktopIntegration/helpers.js` should be used by outside components to extract and compare data coming from the desktop api.

oskarhane-mbpt 2017-10-13 at 13 59 05

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "homepage": "https://neo4j.com",
   "main": "dist/index.html",
   "neo4jDesktop": {
-    "apiVersion": "1.1.0"
+    "apiVersion": "^1.2.0"
   },
   "scripts": {
     "start":
--- a/package.json
+++ b/package.json
@@ -12,11 +12,13 @@
   "homepage": "https://neo4j.com",
   "main": "dist/index.html",
   "neo4jDesktop": {
-    "apiVersion": "1.0.0"
+    "apiVersion": "1.1.0"
   },
   "scripts": {
-    "start": "webpack-dashboard -t 'Neo4j Browser' -- webpack-dev-server --colors --no-info",
-    "starts": "webpack-dashboard -t 'Neo4j Browser' -- webpack-dev-server --https --colors --no-info",
+    "start":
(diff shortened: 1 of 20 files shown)


oskarhane:
Blocking this from merging now since we need to update it to adhere to the new api context format

oskarhane:
Updated to use new API version

PR #665, opened 2017-10-13
Check to see if we get a valid result before mapping firebase tokens

pe4cey:


auth_server/public/getDelegationTokenNewBrowser.html
```diff
@@ -49,16 +49,18 @@
...
+                var data = {
+                  profile: profile,
+                  data_token: res.id_token,
+                  token: hash.id_token,
+                  accessToken: hash.access_token,
+                  state: hash.state,
+                  refreshToken: hash.refresh_token
+                }
+                var out = { userData: data, type: 'auth0:delegation-token' }
+                event.source.postMessage(out, event.origin);
```
oskarhane: I think we still want to post something back so the orig window can drop the interval listener. That goes for all three files.

Issue #584, opened 2017-06-12
Table result box improvements


jdorleans:
![image](https://user-images.githubusercontent.com/1254399/27042043-d7df0c1e-4f63-11e7-8e4d-a0232a81b38e.png)

OskarDamkjaer:
Thanks for the suggestion, should be fixed by #1453 🎉

PR #673, opened 2017-11-03
#584 - Removing max width from data table view

jdorleans:
#584 
- Removing max-width from data table view since it makes really hard to visualise long length strings such as UUIDs and tokens.

### DIFF
--- a/src/browser/components/DataTables.jsx
+++ b/src/browser/components/DataTables.jsx
@@ -52,6 +52,5 @@ export const StyledJsonPre = styled.pre`
   color: ${props => props.theme.preText};
   line-height: 26px;
   padding: 2px 10px;
-  max-width: 320px;
   white-space: pre-wrap;
 `


oskarhane:
Thanks for your contribution @jdorleans!
I think we will need some kind of max-width so we don't end up with a huge horizontal scroll that makes the overview and readability even worse.
Please provide some screenshots of before/after for different kinds of outputs (of different lengths).

oskarhane:
Closing this due to inactivity

nhooey:
@jdorleans @oskarhane: Could you guys figure this one out?

Almost a year after this seemingly reasonable fix was proposed, truncated tables remain an issue.

oskarhane:
I have not looked into this anymore @nhooey. If you know of a sustainable fix, feel free to open a PR for it and we will happily accept it.
The proposed fix in this PR did not render nicely for all kinds of output lengths.

PR #677, opened 2017-11-16
Fix issue where mapping cypher result to vis would break the app

pe4cey:
This removes a conditional statement in the Explorer that tries to access an `undefined` part of the component state.

Resolves https://github.com/neo4j/neo4j-browser/issues/646


oskarhane:
Great find!

PR #678, opened 2017-11-17
Persist grass to neo4j sync

pe4cey:
Fixes a known regression where users' grass files are no longer syncs with the neo4j browser sync service.

### DIFF
--- a/src/browser/modules/Sync/BrowserSync.jsx
+++ b/src/browser/modules/Sync/BrowserSync.jsx
@@ -264,12 +264,6 @@ export class BrowserSync extends Component {
 
     return (
       <Drawer id='sync-drawer'>
-        <button
-          onClick={() =>
-            this.props.sendActionToDispatch({ type: 'grass/SYNC_GRASS' })}
-        >
-          Sync grass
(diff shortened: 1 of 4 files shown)


src/browser/modules/Sync/BrowserSync.jsx
```diff
@@ -264,6 +264,12 @@ export class BrowserSync extends Component {
 
     return (
       <Drawer id='sync-drawer'>
+        <button
+          onClick={() =>
+            this.props.sendActionToDispatch({ type: 'grass/SYNC_GRASS' })}
+        >
+          Sync grass
+        </button>
```
oskarhane: Why do we have this button here? Sync happens automatically, right?

pe4cey: 👍

pe4cey:
PR updated

oskarhane:
There seems to be an issue where style changes are spammed to the sync service, hogging down the Browser completely.
Also, sync changes doesn't seem to be picked up after 20 revisions are reached.

PR #679, opened 2017-11-22
if Relationship has the property weight, the shaft-width is adjusted …

PrinzTu:
…automaticly by  width in pixel = log2(rel.weight)

if it is not set, nothing will happen.

### DIFF
--- a/src/browser/modules/D3Visualization/graphStyle.js
+++ b/src/browser/modules/D3Visualization/graphStyle.js
@@ -526,9 +526,15 @@ export default function neoGraphStyle () {
 
     GraphStyle.prototype.forRelationship = function (rel) {
       const selector = relationshipSelector(rel)
-      return this.calculateStyle(selector)
+      let style = this.calculateStyle(selector)
+      if (rel.propertyMap) {
+        if (typeof rel.propertyMap.weight !== 'undefined') {
+          let shaftWidth = Math.log2(rel.propertyMap.weight)
+          style['props']['shaft-width'] = shaftWidth
+        }
+      }
+      return style
     }
-
     return GraphStyle
   })()
   return new GraphStyle()


oskarhane:
Thanks for this contribution @PrinzTu!
Much appreciated.

However, I don't think we want to hardcode styling to a graph item property name like that.
I'd much rather have it defined in the grass, like we do with the caption, something like:

```
"relationship": {
    "color": "#A5ABB6",
    "caption": "{title}",
    "shaft-width": "{weight}",
}
```

Does that make sense to you?

jexp:
That would be amazing @oskarhane esp. if it was available for all the GRASS properties, it would be a game changer.

PrinzTu:
That would be indeed much nicer.

 

Should be possible to do that with a view lines. Maybe it would also be usefull, to define also a funktion like

„shaft-width“: „{Math.log2(weight) }“

 

What do you think?

 

Von: Oskar Hane [mailto:notifications@github.com] 
Gesendet: Freitag, 24. November 2017 09:23
An: neo4j/neo4j-browser <neo4j-browser@noreply.github.com>
Cc: PrinzTu <matthiasprinz@gmx.de>; Mention <mention@noreply.github.com>
Betreff: Re: [neo4j/neo4j-browser] if Relationship has the property weight, the shaft-width is adjusted … (#679)

 

Thanks for this contribution @PrinzTu <https://github.com/prinztu> !
Much appreciated.

However, I don't think we want to hardcode styling to a graph item property name like that.
I'd much rather have it defined in the grass, like we do with the caption, something like:

"relationship": {
    "color": "#A5ABB6",
    "caption": "{title}",
    "shaft-width": "{weight}",
}

Does that make sense to you?

oskarhane:
I'd say that we start simple by just adding the ability and add more sophisticated features to it later.

oskarhane:
I'm still 👍 to this idea, but I'm closing it due to inactivity. Maybe we'll get time to do it ourselves sometime.

PR #689, opened 2017-12-20
Fix http request issue when using Desktop

pe4cey:
Populates the `restApi` property of the connection obj when using Neo4j Desktop. 
The value is consumed when the user passes a path to one of the http verb Browser commands. E.g. `:get /foobar` will call `${restApi}/foobar`

desktop


src/browser/modules/App/App.jsx
```diff
@@ -194,12 +194,15 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
   const setInitialConnectionData = (graph, credentials, context) => {
     const creds = getActiveCredentials('bolt', context)
     if (!creds) return // No conection. Ignore and let browser show connection lost msgs.
+    const httpCreds = getActiveCredentials('http', context)
+    const restApi = `http://${httpCreds.host}:${httpCreds.port}`
```
oskarhane: Would it be possible to use https if it's available, and http just as a fallback?

PR #709, opened 2018-03-06
Optimise fetching/reading jmx information

pe4cey:
Populates state with the latest jmx values so that we don't have to hit the server every time jmx attributes are read.

_JMX values are refreshed at the same interval as metadata requests_

Before:
before

After:
after


oskarhane:
Do you have any before / after numbers of queries run?

pe4cey:
Updated description to contain log screenshots.

pe4cey:
@oskarhane fix pushed

PR #713, opened 2018-03-12
Add page cache hit ratio to :sysinfo frame

pe4cey:
Add page cache hit ratio to :sysinfo frame.

_Restricts the value returned from `queryJmx` to two decimal places_
screen shot 2018-03-12 at 15 19 54


pe4cey:
Waiting on server change to expose the metric through jmx

PR #722, opened 2018-03-19
Add support for multi-statement execution

oskarhane:
Neo4j Browser will separate them and execute them one-by-one and create one result frame per statement.

Changes that made this possible:

- produce error frame for unknown commands instead of red banner
- upgrade of the error frame to have all error frames look the same
- lot's of changes to `cypher-codemirror` which now exports a `parse` command

oskarhane-mbpt 2018-03-19 at 22 50 57

oskarhane-mbpt 2018-03-19 at 22 49 16


oskarhane:
Don't merge, will add more tests

oskarhane:
More tests added

oskarhane:
This needs to be parsed correctly, and it's currently not.

```
:param x: {
  y: 1,
  z: 2
}
```

PR #723, opened 2018-03-21
Fix issue where action stream was being stored in memory

pe4cey:
This change replaces `zip` from the `postConnectCmdEpic` with `concat`.

The usage of _`zip` (in this epic) was storing the complete history of the action stream and causing a memory leak._

### DIFF
--- a/src/shared/modules/commands/commandsDuck.js
+++ b/src/shared/modules/commands/commandsDuck.js
@@ -153,23 +153,26 @@ export const handleCommandsEpic = (action$, store) =>
     })
 
 export const postConnectCmdEpic = (some$, store) =>
-  some$
-    .concat(some$.ofType(CONNECTION_SUCCESS), some$.ofType(UPDATE_SETTINGS))
-    .do(() => {
-      const serverSettings = getAvailableSettings(store.getState())
-      if (serverSettings && serverSettings['browser.post_connect_cmd']) {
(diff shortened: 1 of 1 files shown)


Kmaschta:
Good catch!

src/shared/modules/commands/commandsDuck.js
```diff
@@ -154,7 +154,7 @@ export const handleCommandsEpic = (action$, store) =>
 
 export const postConnectCmdEpic = (some$, store) =>
   some$
-    .zip(some$.ofType(CONNECTION_SUCCESS), some$.ofType(UPDATE_SETTINGS))
+    .concat(some$.ofType(CONNECTION_SUCCESS), some$.ofType(UPDATE_SETTINGS))
```
oskarhane: I think we want `.merge` over concat.

oskarhane:
After thinking about it, we do not want to execute this on every `UPDATE_SETTINGS`. We only want to do it once, right after we successfully connected to a database.
So the thought is correct, but we didn't know that all settings actions were stored forever.
We just want the latest one, after a `CONNECTION_SUCCESS` happened.

So the flow is:

```
CONNECTION_SUCCESS -> UPDATE_SETTINGS -> check for post_connect_cmd and execute it

```

Not sure how to write that in code yet though.

PR #730, opened 2018-04-03
Clean html to remove js injection from href

pe4cey:


src/shared/services/remoteUtils.js
```diff
@@ -21,16 +21,21 @@
 /* global btoa */
 import { getUrlInfo } from 'services/utils'
 
-export function cleanHtml (string) {
-  if (typeof string !== 'string') return string
-  string = string.replace(
+const removeJavascriptFromHref = string =>
+  string.replace(/href=".*javascript:[^"]*"/, 'href=""')
```
oskarhane: Check single quotes as well

PR #736, opened 2018-04-17
Fix Cypher warning position in Warnings View

oskarhane:
Don’t strip ‘EXPLAIN’ and use 'offset' instead of 'column'.

oskarhane-mbpt 2018-04-17 at 08 43 10


oskarhane:
Closes https://github.com/neo4j/neo4j/issues/10472

PR #739, opened 2018-04-17
Detect connection loss and display error message + show retry banner

oskarhane:
E2E test added for the error message.

This PR also fixes the ugly error message in server log when we try to connect without a scheme when checking if auth is off.
Empty username + pw ensures the same thing, but without the error message.

And one more thing in this pr: when using web workers, a connection loss wasn't showing the 'Connection to server lost, reconnecting...' banner.

oskarhane-mbpt 2018-04-17 at 12 22 51

oskarhane-mbpt 2018-04-18 at 13 49 25


oskarhane:
I brought back the `Reconnecting...` banner as part of this PR as well since it builds on top of the original changes.

Issue #705, opened 2018-02-23
Parameters are auto-converted to float


szarnyasg:
Version: 3.3.3

Running the following expression returns `3`

```
RETURN 10 / 3 AS x
```

```
╒═══╕
│"x"│
╞═══╡
│3  │
└───┘
```

However, when running the same expression with a parameter, it (presumably) does an automatic conversion to float and returns `3.33...`

```
:param {par: 10}
```

```
RETURN $par / 3 AS x
```

```
╒══════════════════╕
│"x"               │
╞══════════════════╡
│3.3333333333333335│
└──────────────────┘
```

The conversion happens around the web UI.

I have hastily written some code to demonstrate that using parameters on their own is not sufficient to get an automatic floating point conversion (neither when using the embedded database, nor the driver).

```java
GraphDatabaseSettings.BoltConnector bolt = GraphDatabaseSettings.boltConnector("0");

GraphDatabaseService graphDb = new GraphDatabaseFactory()
        .newEmbeddedDatabaseBuilder( new File("/tmp/db") )
        .setConfig("apoc.import.file.enabled", "true")
        .setConfig(bolt.type, "BOLT")
        .setConfig(bolt.enabled, "true")
        .setConfig(bolt.address, "localhost:7688")
        .newGraphDatabase();

Result result = graphDb.execute(
        "RETURN $par / 3 AS x", ImmutableMap.of("par", 10)
);
System.out.println(result.next());

Driver driver = GraphDatabase.driver("bolt://localhost:7688");
Transaction tx = driver.session().beginTransaction();
StatementResult result2 = tx.run(
        "RETURN $par / 3 AS x", ImmutableMap.of("par", 10)
);
System.out.println(result2.next());
```

This returns:

```console
{x=3}
Record<{x: 3}>
```

PR #735, opened 2018-04-16
Support new Neo4j types

oskarhane:
This PR adds support for the new Neo4j types:

```
  Point,
  Date,
  DateTime,
  Duration,
  LocalDateTime,
  LocalTime,
  Time
```

This also removes the driver mock we used to have. It's getting tiresome to keep up the interface changes. No need to mock.

It formats `point` types as seen in the screenshots. Temporal types are `.toString()` formatted.
E2E tests are added (couldn't add unit tests because detached driver didn't output something testable).

oskarhane-mbpt 2018-04-22 at 21 37 27
oskarhane-mbpt 2018-04-22 at 21 37 16

PR #740, opened 2018-04-17
Add new param syntax to evaluate against the server

pe4cey:
## Summary

Introduces additional `:param` syntax that evaluates the expression against neo4j server

## More info

The statement to the right of `=>` will be evaluated against neo4j server (as cypher) and the resulting value stored against the param name to be used in future cypher statements (e.g. RETURN `$name`). 

NOTE: Setting params using `:param foo: "bar"`is still supported  but will _not_ be executed against the neo4j server

#### Browser command - return value _(type)_
`:param foo => "bar"` ->`"bar"` _(string)_
`:param foo => 1 + 1` ->`2` _(bolt int)_
`:param foo => date()` -> `"2018-04-23"` _(bolt date string)_

Fixes https://github.com/neo4j/neo4j-browser/issues/705

### DIFF
--- a/e2e_tests/integration/index.spec.js
+++ b/e2e_tests/integration/index.spec.js
@@ -126,19 +126,6 @@ describe('Neo4j Browser', () => {
       .get('[data-test-id="sidebarMetaItem"]', { timeout: 30000 })
       .should('have.length', 18)
   })
-  it('will add parameter using `:param` command', () => {
-    // add cypher evalutated param function
-    const command = ':param foo => 1 + 1'
-    cy.executeCommand(command)
-    cy
-      .get('[data-test-id="rawParamData"]')
-      .first()
-      .should('contain', 'foo')
-    cy
-      .get('[data-test-id="rawParamData"]')
-      .first()
-      .should('contain', 2)
-  })
   it('will clear local storage when clicking "Clear local data"', () => {
     const scriptName = 'foo'
     cy.get(Editor).type(`//${scriptName}`, { force: true })
--- a/e2e_tests/integration/index.spec.js
+++ b/e2e_tests/integration/index.spec.js
@@ -126,6 +126,25 @@ describe('Neo4j Browser', () => {
       .get('[data-test-id="sidebarMetaItem"]', { timeout: 30000 })
       .should('have.length', 18)
   })
+  it('will add parameter using `:param` command', () => {
+    // add cypher evalutated param function
+    const command = ':param foo => 1 + 1'
+    cy.get(Editor).type(command, { force: true })
+    cy.get(Editor).should('have.value', command)
+    cy.get(SubmitQueryButton).click()
+    cy
+      .get('[data-test-id="frameCommand"]')
+      .first()
+      .should('contain', command)
+    cy
+      .get('[data-test-id="rawParamData"]')
+      .first()
+      .should('contain', 'foo')
+    cy
+      .get('[data-test-id="rawParamData"]')
+      .first()
+      .should('contain', 2)
+  })
   it('will clear local storage when clicking "Clear local data"', () => {
     const scriptName = 'foo'
     cy.get(Editor).type(`//${scriptName}`, { force: true })
@@ -149,24 +168,5 @@ describe('Neo4j Browser', () => {
 
     // once data is cleared the user is logged out and the connect form is displayed
     cy.get('input[data-test-id="boltaddress"]')
-  })
-  it('will add parameter using `:param` command', () => {
-    // add cypher evalutated param function
-    const command = ':param foo => 1 + 1'
-    cy.get(Editor).type(command, { force: true })
-    cy.get(Editor).should('have.value', command)
-    cy.get(SubmitQueryButton).click()
-    cy
-      .get('[data-test-id="frameCommand"]')
-      .first()
-      .should('contain', command)
-    cy
-      .get('[data-test-id="rawParamData"]')
-      .first()
-      .should('contain', 'foo')
-    cy
-      .get('[data-test-id="rawParamData"]')
-      .first()
-      .should('contain', 2)
   })
 })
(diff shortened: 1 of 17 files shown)


src/shared/modules/commands/commandsDuck.test.js
```diff
@@ -137,6 +144,12 @@ describe('commandsDuck', () => {
       const cmdString = cmd + ' x: 2'
       const id = 1
       const action = commands.executeCommand(cmdString, id)
+      bolt.routedWriteTransaction = jest.fn(() =>
```
oskarhane: This is needed here, because this should not call that function.

src/shared/modules/commands/commandsDuck.test.js
```diff
@@ -175,6 +189,7 @@ describe('commandsDuck', () => {
           { type: commands.KNOWN_COMMAND },
           replaceParams({ x: 2, y: 3 }),
           frames.add({ ...action, success: true, type: 'params', params: {} }),
+          { type: 'meta/FORCE_FETCH' },
```
oskarhane: We shouldn't force fetch on param setting

src/shared/modules/commands/helpers/params.js
```diff
@@ -42,14 +72,33 @@ export const handleParamsCommand = (action, cmdchar, put, store) => {
...
+            const result = recursivelyTypeGraphItems(obj)
+            put(update(result))
+            resolve({ result, type: 'param' })
+          })
+          .catch(e => reject(e))
       } catch (e) {
-        return reject(
-          new Error('Could not parse input. Usage: `:param "x": 2`. ' + e)
-        )
+        reject(new Error('Could not parse input. Usage: `:param "x": 2`. ' + e))
```
oskarhane: I think we should promote the new format here

src/shared/modules/commands/helpers/params.test.js
```diff
@@ -44,6 +50,8 @@ describe('commandsDuck params helper', () => {
   })
   test('handles :param "x": 2 and calls the update action creator', () => {
     // Given
+    // jest.spyOn(params, , accessType?)
```
oskarhane: not needed

src/shared/services/commandInterpreterHelper.js
```diff
@@ -90,7 +91,7 @@ const availableCommands = [
     name: 'set-params',
     match: cmd => /^params?\s/.test(cmd),
     exec: function (action, cmdchar, put, store) {
-      handleParamsCommand(action, cmdchar, put)
+      return handleParamsCommand(action, cmdchar, put, bolt)
```
oskarhane: It looks like `handleParamsCommand ` isn't using the '`bolt` argument so it can be skipped here (and also the import up top)

e2e_tests/integration/index.spec.js
```diff
@@ -108,6 +108,25 @@ describe('Neo4j Browser', () => {
       .get('[data-test-id="sidebarMetaItem"]', { timeout: 30000 })
       .should('have.length', 18)
   })
+  it('will add parameter using `:param` command', () => {
+    // add cypher evalutated param function
+    const command = ':param foo => 1 + 1'
+    cy.get(Editor).type(command, { force: true })
```
oskarhane: Nice with e2e test! 
You could use `cy.executeCommand(command)` here instead to clean up a bit.

src/browser/modules/Stream/ParamsFrame.jsx
```diff
@@ -28,7 +28,9 @@ const ParamsFrame = ({ frame, params }) => {
   const contents = (
     <PaddedDiv>
       <Render if={frame.success !== false}>
-        <pre>{JSON.stringify(frame.params, null, 2)}</pre>
+        <pre data-test-id='rawParamData'>
+          {JSON.stringify(frame.params, null, 2)}
```
oskarhane: I think we want a nicer output for the new types. Might be in a separate PR, let's chat about it.

src/shared/modules/commands/commandsDuck.test.js
```diff
@@ -43,6 +43,10 @@ import {
   replace as replaceSettings
 } from 'shared/modules/settings/settingsDuck'
 import { cleanCommand, getInterpreter } from 'services/commandUtils'
+import bolt from 'services/bolt/bolt'
```
oskarhane: Bolt should not be needed in this file since we're not testing the new format, right?

src/shared/modules/commands/helpers/params.js
pe4cey: Display both usages or only the new `=>` approach?

pe4cey:
PR updated @oskarhane

src/shared/modules/commands/commandsDuck.test.js
pe4cey: I've just added a test in the latest commit that requires this.

oskarhane:
In https://github.com/neo4j/neo4j-browser/pull/735 I've added a new function that formats all types in a nice way. It'd be great if that could be used when printing the params as well.

src/shared/modules/commands/commandsDuck.test.js
```diff
@@ -137,6 +144,43 @@ describe('commandsDuck', () => {
...
+      // When
+      store.dispatch(action)
+
+      // Then
+      // See above
+    })
+    test('does the right thing for :param x => 2', done => {
+      // Given
+      const cmd = store.getState().settings.cmdchar + 'param'
+      const cmdString = cmd + ' x: 2'
```
oskarhane: This doesn't test what the test title says it should test.

src/shared/modules/commands/helpers/params.js
oskarhane: Both, I think.

Issue #682, opened 2017-11-24
UX: Broswer window status bar not always showing all properties of a node in Graph View


legolego:
### Description
When in the Graph View, the properties of nodes (other than `<id>`) aren't always visible when clicking/hovering over them.

### Reproduce Steps
Have a graph visible in the window. The total character length of the Node property labels and values should be shortish (depends on monitor/width of window). With no Node selected, hovering over a Node with a shorter character length of property labels/values, all of them appear in the  status bar below the graph:
![neo4j-hover-visible](https://user-images.githubusercontent.com/1666188/33218239-d31a21d2-d109-11e7-853c-cbd3a38b152c.png)

but hovering over a Node with a longer character length, only the `<id>` is visible, but also the toggle to open the status bar to show all the properties:
![neo4j-hover-not visible-with toggle](https://user-images.githubusercontent.com/1666188/33218246-dda02548-d109-11e7-9337-197ea2388384.png)

If the Node with a longer property label/value length is selected (Node 1) and the mouse is hovered over a shorter one (Node 5), all of the status text is visible, but also the toggle that isn't needed:
![neo4j-hover-visible-unneeded toggle](https://user-images.githubusercontent.com/1666188/33218421-d5e8fa36-d10a-11e7-84c5-bc0df4eb0109.png)

The problem case is if a Node with a shorter property label/value length is selected (Node 5) and the mouse is hovered over a longer one (Node 1), there is no indication of property labels or values except for `<id>`. 

This looks like the node has no data.
![neo4j-hover-visible-no text or toggle](https://user-images.githubusercontent.com/1666188/33218500-2e6218aa-d10b-11e7-924b-d8d9fa95fe19.png)

It's like the toggle icon is shown based on the selected Node, regardless of the Node being hovered over. This caused me to spend quite a bit of time trying to figure out why there weren't any properties on only some of the Nodes. I finally looked at the Table view and saw everything had been there all along.

### Expected Behavior
On hovering over any Node and independent of the selected Node, I would expect as much property label/value text to be shown as will fit into the status bar, but ended with three dots/ellipsis to show continuation, and maybe more prominent toggle icon too. 

### Your Environment

| software name  | version
| ---------------|----------
|                |  
| OS             |  Windows 10
| Web browser    |  Chrome/Opera/IE
| node.js        | 
| npm            |
| yarn           |
| Neo4j Browser  |
| Neo4j          | 3.3.0 Enterprise (as a developer)

### Additional Information

[...]

PR #764, opened 2018-05-23
Fix missing inspector items when spread over more than one line

pe4cey:
Fixes issue where expanding inspector would force the status bar content out of the viewport.
See #682 

Before:
screen shot 2018-05-23 at 16 54 50

After:
screen shot 2018-05-23 at 16 51 59


oskarhane:
Nice one!

PR #765, opened 2018-05-24
Visualization: Avoid calling 'toString' on empty objects

pe4cey:
Previously the visualisation would throw an error that would cause the render to break

### DIFF
--- a/src/browser/modules/D3Visualization/components/Inspector.jsx
+++ b/src/browser/modules/D3Visualization/components/Inspector.jsx
@@ -50,7 +50,7 @@ const mapItemProperties = itemProperties =>
           {prop.key + ': '}
         </StyledInspectorFooterRowListKey>
         <StyledInspectorFooterRowListValue className='value'>
-          {prop.value.toString()}
+          {prop.value ? prop.value.toString() : prop.value}
         </StyledInspectorFooterRowListValue>
       </StyledInspectorFooterRowListPair>
     ))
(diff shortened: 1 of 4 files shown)


src/browser/modules/D3Visualization/components/Inspector.jsx
```diff
@@ -50,7 +50,7 @@ const mapItemProperties = itemProperties =>
           {prop.key + ': '}
         </StyledInspectorFooterRowListKey>
         <StyledInspectorFooterRowListValue className='value'>
-          {prop.value.toString()}
+          {prop.value ? prop.value.toString() : prop.value}
```
oskarhane: What happens if `prop.value` is `0` or `false`?
I'd suggest a stricter check: 
```js
![null, undefined].includes(prop.value) && typeof prop.value.toString === 'function'
``` 
or similar.

src/browser/modules/D3Visualization/mapper.js
```diff
@@ -20,7 +20,7 @@
 
 const mapProperties = _ => Object.assign({}, ...stringifyValues(_))
 const stringifyValues = obj =>
-  Object.keys(obj).map(k => ({ [k]: obj[k].toString() }))
+  Object.keys(obj).map(k => ({ [k]: obj[k] ? obj[k].toString() : obj[k] }))
```
oskarhane: Same as above

PR #766, opened 2018-05-24
Word break long values in db info drawer

pe4cey:
Before:
screen shot 2018-05-24 at 10 32 05

After:
screen shot 2018-05-24 at 10 31 49


oskarhane:
Simple but powerful :)

PR #781, opened 2018-06-05
Show number of nodes and relationships in db sidebar

oskarhane:
Count the number of nodes and relationships on every db-meta update.

Display like we do in the viz legend.

oskarhane-mbpt 2018-06-04 at 15 08 03

### DIFF
--- a/src/browser/modules/DatabaseInfo/DatabaseInfo.jsx
+++ b/src/browser/modules/DatabaseInfo/DatabaseInfo.jsx
@@ -49,7 +49,9 @@ export class DatabaseInfo extends Component {
       properties = [],
       userDetails,
       databaseKernelInfo,
-      onItemClick
+      onItemClick,
+      nodes,
+      relationships
     } = this.props
 
     return (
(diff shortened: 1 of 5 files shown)


jexp:
Cool! Can we do that for all the labels + rel-types?
Either with: `match (n:Label) return count(*)` queries that should all execute in single ms.
Or with a trick in one query:

```
MATCH (:Label1) RETURN {label:'Label1', count:count(*)} as info
UNION ALL
MATCH (:Label2) RETURN {label:'Label2', count:count(*)} as info
UNION ALL
MATCH ()-[:TYPE1]->() RETURN {type:'TYPE1', count:count(*)} as info
```
Started streaming 3 records in less than 1 ms and completed after 1 ms.

```
╒════════════════════════════╕
│"info"                      │
╞════════════════════════════╡
│{"label":"Label1","count":0}│
├────────────────────────────┤
│{"label":"Label2","count":0}│
├────────────────────────────┤
│{"type":"TYPE1","count":0}  │
└────────────────────────────┘
```

pe4cey:
LGTM

PR #782, opened 2018-06-11
Pool and re-use web workers

oskarhane:
This introduces a web worker pool (generic so it can be used for other expensive computations as well).
Currently the pool size for cypher queries is hard coded in https://github.com/neo4j/neo4j-browser/pull/782/files#diff-046ce42896ad79b95a44f41c7faa1f94R39

This could possibly be a user config in the future, if we see the need to for it.
If no worker is free, the jobs are queued and executed in order.

This can be tested by setting the max pool size to 1 and execute a long running query (this takes about 10 secs), like

```
MATCH (n:Movie)<--()-->(m:Movie)<--()-->(o:Movie)<--()-->(p:Movie)<--()-->(q:Movie)
WITH n,p,count(*) as c
RETURN count(*)
```

on the movie data set and directly afterwards execute a few more. These will then wait with a spinner.

In the blob directory you'll only see one ww blob.
I've set the hard coded pool size to 10 for no particular reason. I've never hit that limit if I'm not hammering the keyboard with queries.

## Try it
Open browser, open devtools -> `Application`.
Expand `Other` under `Frames`.
See all the uuid named blobs. That's ww (to verify, right click and show in network tab).

1. Reload neo4j-browser
1. Click a few meta items in sidebar (or run any queries). Wait for about 30 secs.
1. Compare results with and without the changes in this pr.

### Before

oskarhane-mbpt 2018-06-11 at 13 17 27

### After

oskarhane-mbpt 2018-06-11 at 13 15 25

changelog: Fix issue where neo4j-browser kept allocating more and more disk space


reznord:
This is a huge improvement and saves a lot of disk space

src/shared/services/bolt/bolt.js
```diff
@@ -35,8 +36,7 @@ import BoltWorkerModule from 'worker-loader?inline!./boltWorker.js'
 /* eslint-enable import/no-webpack-loader-syntax */
 
 let connectionProperties = null
-let boltWorkerRegister = {}
-let cancellationRegister = {}
+let boltWorkPool = new WorkPool(() => new BoltWorkerModule(), 10)
```
pe4cey: FYI if we choose to surface the worker pool as `:config` we will probably need to only allow worker pools `> 0` as a pool of `0` is always in a `BUSY` state

oskarhane: Yeah, 0 would be infinite imo. 
New ones won't be recreated as long as there are existing free ones anyway.

PR #784, opened 2018-06-18
Switch to React 16.4

oskarhane:
Main work was to make the JSX pragma import explicit.
And fix missing `key` props.
And some ref differences for viz, scrolling stream to top, enter in forms to get to next field etc.

Ported most of the component tests to [react-testing-library](https://github.com/kentcdodds/react-testing-library), and it's really easy to use and tests get very clean and you're not lured into testing implementation details.

I've added a bunch of extra e2e tests as well to make sure none of our commands breaks the app.

Please try to break it.
It can be found on https://dist-xjxjivnrky.now.sh. Just connect to your local db and hammer it.


oskarhane:
@akollegger Can you check if https://github.com/oskarhane/neo4j-browser/blob/403b2d955aa7d83acdffd78e49715a7a53bd9dcd/NOTICE.txt looks ok?

PR #789, opened 2018-06-25
Update to webpack 4

oskarhane:
Lot’s of changes.
Builing creates production (uglified) code and chunks.
Hot reloading works in dev mode, so it all seems fine.

Some changes to how hot reloading is done and I renamed the bolt web worker file to have it more easily loaded through regular webpack config.


pranaygp:
Hey @oskarhane

I noticed that you're running into an error with a rather cryptic error message. Just wanted you to know I'm investigating this right now from our end and would love to make sure this integration goes smoothly for you.

pranaygp:
It seemed like the deployment actually worked, but the bot was simply unable to comment back @oskarhane. You can find the deployment here: https://neo4j-browser-wyfiuvnjfg.now.sh/

Also, on inspecting your `now.json` file, I noticed your alias is set to "neo4j-browser". The configuration should be a [URL](https://zeit.co/docs/features/configuration#alias(string|array)) you'd like to auto alias to when you run `now alias`

pranaygp:
@oskarhane it's an issue on our end and we're currently investigating it so we can propagate error messages better to you.

Would you be happy to jump onto https://zeit.chat and DM me (@pranaygp) so I can help figure this out?

PR #800, opened 2018-07-23
Add `:history clear` command

pe4cey:
Remove previously executed editor statements using `:history clear` command.

history


irfannuri:
It now accepts whatever follows `:history` and lists history. For instance you can input `:history asdasd adf`. It used to present an error like `Type :help commands  for a list of available commands.`

PR #812, opened 2018-08-20
Add pagecache hits and misses to plan output

oskarhane:
Show page cache stats right next to hits / estimated row stats.
I also added a E2E test for it.

oskarhane-mbpt 2018-08-20 at 10 41 11

_(Sorry for the formatting changes, I guess prettier got updated.)_

### DIFF
--- a/src/browser/external/neoPlanner.js
+++ b/src/browser/external/neoPlanner.js
@@ -164,8 +164,8 @@ neo.queryPlan = function(element) {
         (left =
           (left1 =
             operator.Expressions != null
-             ? operator.Expressions
-             : operator.Expression != null
+              ? operator.Expressions
+              : operator.Expression != null
                 ? operator.Expression
                 : operator.LegacyExpression != null
                   ? operator.LegacyExpression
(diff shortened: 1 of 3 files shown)


src/browser/external/neoPlanner.js
```diff
@@ -179,6 +179,19 @@ neo.queryPlan = function(element) {
       details.push({ className: 'padding' })
     }
 
+    if (operator.PageCacheHits || operator.PageCacheMisses) {
```
pe4cey: Aren't `PageCacheHits` and `PageCacheMisses` (if present) allow ways come together in the result?
If one is present but the other isn't how does `formatNumber` bellow handle `undefined`?

src/browser/external/neoPlanner.js
```diff
@@ -179,6 +179,19 @@ neo.queryPlan = function(element) {
...
+      details.push({
+        className: 'pagecache-hits',
+        key: 'pagecache hits',
+        value: formatNumber(operator.PageCacheHits)
```
pe4cey: ^

src/browser/external/neoPlanner.js
oskarhane: Yes, they come together.
Those are also zero checks since the result sends `PageCacheHits: 0, PageCacheMisses: 0` and in that case we don't want to show them at all. 
But if one of them aren't 0, we want to show them. 
@pe4cey.

PR #829, opened 2018-10-01
Update icons

pe4cey:
This replaces most streamline font usage with svg's

The svg's are loaded in two different ways. Both allow custom styling (mostly color) to be applied
- Main UI: Loaded using `raw-loader` & `react-svg-inline` to inline svg assets
- Guides: Embedded inline within the `.html`


oskarhane:
Good stuff!
It seems like icons easily can get stuck in the hover state if you move the mouse pointer quite fast (not unnaturally fast though, it happened unintentionally multiple times for me before I understood how to reproduce it).

See gif.

icons

PetarAlexandrov:
Looks good, Mark! I've also noticed the hover thing Oskar mentioned

PR #837, opened 2018-10-09
Introduce experimental features section in Settings

oskarhane:
This gives us a tool to rollout features that we want to get feedback on before releasing them.

At the same time as it gives users a list of available experimental features to chose from, users understand that they're experimental and can break at any time.

Users toggle the experimental features from the Settings sidebar, at the bottom.
With this PR it's hidden because there are no features to toggle.

Also note that to be able to persist the users settings we store this in the users local storage. However, men we move a feature off of the experimental list, it will still show up in users Settings sidebar if we don't check what should be there. 
That's the reason for the `cherrypickAndMerge` function in the experimental feature duck file.

This is what it looks like with a feature added:
oskarhane-mbpt 2018-10-09 at 13 22 03

To try manually, add this to the `initialState` of the experimental features reducer:

```
snake: {
    name: 'snake',
    on: true,
    displayName: 'Snake Game',
    tooltip: 'Enable the command :snake to play a game of snake'
  }
```

btw, the experimental features section in the Settings panel is feature toggled by the default entry. Non visible to change for the users though, but you can toggle it via local storage.

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -45,7 +45,8 @@ applyKeys(
   'folders',
   'grass',
   'syncConsent',
-  'udc'
+  'udc',
+  'experimentalFeatures'
 )
 
 // Create suber bus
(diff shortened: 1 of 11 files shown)


jexp:
I want the snake game :)
Does multiline editor still count as experimental?

PR #852, opened 2018-10-22
Add `Order` to plan output if it exists

oskarhane:
Only in neo4j >= 3.5.0.

## When connected to neo4j 3.4:

plan 34

## On 3.5.0 without index:

plan 35

## On 3.5.0 with index:

plan 33


sherfert:
Nice! 

I would change `Order by index` into either `Ordered by` or `Row ordering`. The word index suggests that the ordering always results from an index, but it can also be inserted after a `Sort` operation, as in your  "On 3.5.0 without index" example.

PR #872, opened 2018-11-19
Chunk explosion

oskarhane:
Not sure if good idea.


oskarhane:
Seems like e2e tests running on our CI can't load properly when so many files.
Not sure if Cypress or CI:s fault.

PR #887, opened 2018-12-14
Additional controls & visualization options

srstrickland:
Disclaimer: I'm a backend developer and this isn't my wheelhouse.  Feel free to pick out pieces of this, or none at all.  Just thought I would share some additional visualization controls I threw together to help with a project I worked on earlier this year.  Neo4j was a key piece of that puzzle, serving for me as the visualization tool to describe relationships (and the overall state of the world) that had been built by an offline spark-based graph algorithm I designed.  I found myself exploring the graph and repeating some cumbersome tasks, so I thought I would add my own helpful buttons and configuration options.  In the end, we've kept Neo4j running and it's a key diagnostic tool for us to understand the relatively complex graph processing that happens behind the scenes.  I'll try to describe all of the new features and their motivation below:

**Thumbnail Support**
Pretty self-explanatory.  In the settings drawer, the user can define the attribute that holds the URL of the thumbnail, as well as the opacity of the image:

screen shot 2018-12-14 at 9 04 52 am

The result is that nodes get an overlay of an image if the specified attribute exists and points to a valid URL:

screen shot 2018-12-14 at 9 06 01 am

**Additional Buttons: Copy Attribute & Open Link**

Two things I found myself doing quite often is copying some attribute of a node & opening an external URL, either as defined by an attribute or by some kind of template.  Both of these were rather cumbersome; click on node, expand the attributes below, select the right text, copy, then do something else.  So I added some new buttons to the donut that pops up around a node when selected (please, forgive the svg; I originally developed this against an old `master` and at that time just used unicode chars; these new buttons I just threw together, but they don't match your line style very well):

screen shot 2018-12-14 at 12 08 34 pm
screen shot 2018-12-14 at 12 08 43 pm

The behavior of these buttons is controlled by additional settings-drawer options:

screen shot 2018-12-14 at 9 14 47 am

URLS: Multiple entities may be supplied, separated by `|`, and the first one that "matches" is used.  In this case, if `url` exists as an attribute of a node, that will be used for the external-link button.  Otherwise, if `foo` exists, `http://blah.com/foo` will be opened.  If none of the attributes or templates can be fully matched, nothing happens.

Copy Attribute: Again, multiple may be supplied, but no templating is supported here.  Just attribute names, and the first one that matches gets dropped on the clipboard.

**Double-Click Behavior**

Sometimes I was doing a lot of copying or opening external URLs, so I found it useful to have a setting control what happens on doubleclick:

screen shot 2018-12-14 at 9 19 27 am

**Auto-Execution of Guide Queries**

Like I said, Neo4j is a key exploration tool for us.  I built a "guide" that basically serves as our own internal diagnostic tool.  This serves our purposes perfectly, but the default behavior of guides is to pre-fill the query bar at the top with text.  In that respect it seems like more of a tutorial vessel, which is great... but our target users aren't necessarily interested in forming their own graph queries; they just want to drop in some parameters and click "go".  So I added an option to the left side to skip the text-population phase, and simply execute guide queries immediately:

screen shot 2018-12-14 at 9 23 33 am

**Additional HTML Controls in Guides**

Again, we love neo4j guides!  But they only support text-based HTML inputs, as they currently respond only to `onKeyUp` events.  I added support for `onChange`, which I wanted specifically to support `<select>` inputs:

screen shot 2018-12-14 at 9 25 32 am

**Immediate Settings**

I found myself tweaking the new settings drawer quite a bit, and as it was, I would have to generate a new frame in the browser for them to take effect.  I wanted to just hit "reload" on any currently existing frame and have the modified settings take effect (e.g. toggle thumbnails or change my external-url-template setting).  So I changed the orchestration of the settings to make this happen.

**Minor Tweaks**

* Since thumbnail displays may result in unreadable text, give style options for controlling a `shadow-color` in `*.grass` specs
* Relationships now support outline colors

**Final Thoughts**

All this adds a fair amount of settings to the existing drawer, and may be overwhelming or due for refactoring by someone with more UI/UX background than me.  Altogether this is what it looks like as is:

screen shot 2018-12-14 at 9 34 07 am

Feel free to do with all this as you'd like; I have no expectation for all this to be accepted.  But thought I would at least share with you guys some things that I found really helpful.


oskarhane:
Thanks a lot for this contribution @srstrickland 🙇‍♂️!
There are ideas in here that we definitely like and we really appreciate you taking the time to contribute to this project.

With the holidays coming up I don't think we'll have the time to discuss and get back to you in any other way than this short response from me.

Expect us to come back to you in January when we're back.

In the meantime, please sign our [CLA](https://neo4j.com/developer/cla/) so we can start accepting code from you.

Merry Christmas, and once again, thanks!

graphadvantage:
Very nice - great job, just posted a similar pull request. Images are very helpful in understanding and driving adoption. Be great to have this as core functionality, perhaps governed by an override switch (show images Y/N) to reduce perf impact for large renders. Another nice enhancement would be to be able to point to an arbitrary icon library - there are many use-case specific icon sets (IT, Supply Chain, Transportation, eCommerce, etc) that would add additional clarity to a render.

srstrickland:
Thanks for the feedback!

I've added some comments to #895 and will happily defer to that PR.

I'd be OK with removing the button for copy, cuz it is sort of an awkward construct, but will note that most graph navigation is done via mouse (at least once something is rendered), so it's a bit jarring and hidden to move to the keyboard.  But I agree that it's a safer visual approach and probably best to be conservative here.

Regarding URL support: one thing I was trying to accomplish with this feature was based off something that elastic built for their graph product, where clickable links are generated based on attributes of a node; the user supplies the template.  Yes, you could require that all nodes prepopulate some links, but that shifts the responsibility up, might clutter the nodes (and the storage space), and only works for a small subset of use cases, like pre-defined IDs (e.g. specific ids that correlate with exactly one URL).  Think along the lines of "custom search engines" in chrome.  The user can define any URL template to use, and the format flag `%s` gets replaced by the searched phrase.  This is similar, but accessing attributes by `{name}`.  It's not very elegant or well presented in this PR, so needs some iteration.  But something to keep in mind.

OK, done with rhetoric! On to actionable stuff!

I would be happy to extract the stuff that's generally useful and non-controversial.  As you mention, this includes immediate settings.  I think it might also contain the tiny addition for guides to support `<select>` elements (and other html controls that don't involve the keyboard).  Slightly more controversial is the option for auto-executing "guide" queries.  I suppose I could open a PR for each of these to keep them isolated and narrowly focused.

johnaco:
I'd also love to see this implemented in base product.  My requirement was even simpler - if a selected node/rel has properties that are a URL, that the property be clickable to navigate to the URL.
But with the approach above, even nicer to allow 'custom actions' to a node/rel.

oskarhane:
Closing this due to inactivity, but not saying we don't want some of the features as described in https://github.com/neo4j/neo4j-browser/pull/887#pullrequestreview-191585407

Issue #889, opened 2018-12-18
`:help keys` unreadable with Dark Theme


rickhg12hs:
### Description
Using Dark Theme, text unreadable in `:help keys`.

### Reproduce Steps

1. Select `Dark Theme`
2. Execute `:help keys`
3. Observe Output

### Expected Behavior

Readable text.

### Current Behavior

Unreadable text.  Text color same as background color?

### Can you propose a solution?

Use suitable text colors for a dark theme.

### Your Environment

| software name  | version
| ---------------|----------
|                |  
| Fedora Linux             |  4.18.19-100.fc27.x86_64
| Google Chrome    |  71.0.3578.98 (Official Build) (64-bit)
| Neo4j Browser  | 3.2.13
| Neo4j          | 3.5.0 (community)

### Additional Information

neo4j was installed from [yum repo](http://yum.neo4j.org/stable/).

I'm also wondering why the repo neo4j browser version is not current.

oskarhane:
Thanks for this report @rickhg12hs. We'll get on it!

PR #891, opened 2019-01-03
Fix/help keys dark mode

linuslundahl:
* Fixes help keys in dark mode.
* Adjusts help keys layout.
* Fixes Open Sans not loading correctly.

fixes https://github.com/neo4j/neo4j-browser/issues/889


src/browser/styles/open-sans.css
```diff
@@ -2,63 +2,63 @@
   font-family: 'Open Sans';
   font-style: normal;
   font-weight: 300;
-  src: local('Open Sans'), local('OpenSans'), url('fonts/OpenSans-Light.ttf') format('truetype');
+  src: local('Open Sans Light'), local('OpenSansLight'), url('fonts/OpenSans-Light.ttf') format('truetype');
```
oskarhane: Nice find!

linuslundahl:
The Open Sans files had different permissions. Gave them the same permissions as the rest of the fonts to see if that was a factor for Open Sans not loading correctly. Then just left all files with the same permission for consistency.

Before:
before

After:
after

PR #894, opened 2019-01-08
Generate dist/manifest.json

oskarhane:
Use specified fields from `package.json` to reduce duplicate content.


build_scripts/generate-manifest-helpers.js
```diff
@@ -0,0 +1,48 @@
...
+  try {
+    const obj = JSON.parse(fs.readFileSync(file, 'utf8'))
+    return obj
+  } catch (e) {
+    throw new Error('Could not load or parse file: ' + file + '. Error: ' + e)
+  }
+}
+
+function buildTargetObject (data, dataProp) {
+  let out = {}
```
linuslundahl: This should be a `const` since the type never changes.

build_scripts/generate-manifest-helpers.js
```diff
@@ -0,0 +1,48 @@
...
+  const keys = data[dataProp] || []
+  keys.forEach(key => (out[key] = data[key]))
+  return out
+}
+
+function mergeObjects (source, target) {
```
linuslundahl: cleaner way of merging objects `const mergeObjects = (source, target) => ({...target, ...source});`

PR #895, opened 2019-01-08
image support for Neo4j browser

graphadvantage:
overlay images on nodes - useful for displaying graphs of products or other entities with diverse imagery.


oskarhane:
Thanks so much for this PR @graphadvantage!
I do like the option of having images as node bodies, but I think the correct approach would be to have it enabled through GRaSS rather than a magic property on the nodes.

On the roadmap we have a plan for this which is roughly:

A property named `fill` on a node styling declaration can have a mapping to a node property.
That property is treated as an url pointing to an image.

i.e. 

```
node {
  color: #68BDF6;
  fill: "{image_url}";
  caption: "{name}";
}

```
If a `fill` is present we should hide the caption though, otherwise it's going to lead hard to read text and not look so good imo. 
To reveal the caption, the user could hover the node and the `fill` is removed and the `color` + `caption` is displayed.

If you're willing to make those changes and sign our [CLA](https://neo4j.com/developer/cla/) we would happily accept it.

graphadvantage:
Hi

That sounds like it would work, if I understand correctly:

node {
  color: #68BDF6;
  fill: "{image_url}";
  caption: "{name}";
}

where image_url is an arbitrary node property.

I don't mind the captions over the images, and you could also make that configurable

node {
  color: #68BDF6;
  fill: "{image_url}";
  caption: "{name}";
  hide_caption: false
}

oskarhane:
> where image_url is an arbitrary node property.

Yep, exactly.

> I don't mind the captions over the images, and you could also make that configurable

I don't think we should add behavioral configurations to the stylesheet. 
The safe (in regards to have readable and clear visualizations for a generic graph) thing to do imo is to hide and reveal on hover.
I'm open to other suggestions though, but my concern is that the viz will look messy.

src/browser/modules/D3Visualization/lib/visualization/renders/init.js
```diff
@@ -134,6 +135,75 @@ const nodeRing = new Renderer({
...
+      .append('pattern')
+      .attr({
+        id (id) {
+          return 'img-fill-' + id.id
+        },
+        patternUnits: 'userSpaceOnUse',
+        x: -39,
+        y: -39,
+        height: 156,
+        width: 156
```
srstrickland: `x`, `y`, `height`, `width` shouldn't be hard-coded here.  Calculate based on node radius instead, or it won't look right when the user configures different sizes of nodes in their GraSS.  Ran into this myself 😛 ... see [here](https://github.com/neo4j/neo4j-browser/pull/887/files#diff-da817c60492d350eb701f5f1c3e687c1R147)

src/browser/modules/D3Visualization/lib/visualization/renders/init.js
```diff
@@ -134,6 +135,75 @@ const nodeRing = new Renderer({
...
+      })
+
+      .append('image')
+      .attr('xlink:href', link => link.propertyMap.image_url)
+      .attr({
+        x: 0,
+        y: 0,
+        type: 'image/png',
+        height: 78,
+        width: 78
```
srstrickland: same as above, don't hard-code here (see [here](https://github.com/neo4j/neo4j-browser/pull/887/files#diff-da817c60492d350eb701f5f1c3e687c1R154))

srstrickland: handy to throw in a (GraSS-controlled) setting for thumbnail opacity here, as in [here](https://github.com/neo4j/neo4j-browser/pull/887/files#diff-da817c60492d350eb701f5f1c3e687c1R155). but more of a feature request than an issue with this code 😛

srstrickland:
@oskarhane your intuition is right; if we do nothing then text and images do clash with each other; I ran into this issue while working on #887.  But for my use case I needed to show the text for all nodes; the images were a slight visual clue but text was still most important.  I found a solution though, to use a text-shadow.  You could hard-code it, or just leave it up to the user to configure via stylesheet.  It did a good job visually separating the elements, even when thumbnail opacity was 100% and the image is very busy.

BTW mentioned in code comments but a control for image opacity would be great, for when you want the images to be more subtle and let the background color shine through.

styk-tv:
@srstrickland @graphadvantage @oskarhane ping! 

Guys can we please merge this into mainstream?

zsz00:
Guys can we please merge this into mainstream?

oskarhane:
I'm closing this due to inactivity but not saying it's unwanted. 
Some changes are required for this to go in and it's been a year since last activity.

PR #909, opened 2019-03-11
Fix showCurrentUser procedure path and race condition for storing current user

oskarhane:


src/shared/modules/features/versionedFeatures.js
```diff
@@ -0,0 +1,48 @@
...
+
+const NEO4J_TX_METADATA_VERSION = '3.5.0-alpha01'
+const NEO4J_4_0 = '4.0.0-alpha01'
+
+export const canSendTxMetadata = state => {
+  const serverVersion = getVersion(state)
+  if (!semver.valid(serverVersion)) {
+    return false
+  }
+  if (semver.gt(serverVersion, NEO4J_TX_METADATA_VERSION)) {
```
linuslundahl: `return semver.gt(serverVersion, NEO4J_TX_METADATA_VERSION);` instead?

oskarhane: Perfect, thanks!

PR #910, opened 2019-03-11
Add tx metadata to queries that's generated by user actions

oskarhane:
Added to these

- EXPLAIN query that's called when user types
- Viz actions like auto-connect and explore neighbors 
- `:queries` frame
- `:schema` frame
- `:sysinfo`
- `:server user add` and `:server user list`


linuslundahl:
Looks good!

PR #937, opened 2019-06-19
Detect operating enviromentments more generically

oskarhane:
Also opt out of showing browser-sync sidebar in an additional operating environment.


eve-bright:
[If Browser Sync doesn't appear in Desktop any more, can we get rid of it for Cloud too?](https://trello.com/c/nkb2jT4N/24-if-browser-sync-doesnt-appear-in-desktop-any-more-can-we-get-rid-of-it-for-cloud-too)

PR #939, opened 2019-06-20
Upgrade dependencies minor versions

oskarhane:
Switch from `react-testing-library` -> `@testing-library/react` and adapt to a few breaking changes there. 
See https://github.com/testing-library/dom-testing-library/issues/260

Use core-js@3 with babel and `core-js/stable` instead of `@babel/polyfill`.
See https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babelpreset-env

To use proper react hot loader, add an alias for react-dom in webpack.
See: https://github.com/gaearon/react-hot-loader#react--dom


src/browser/modules/Stream/Extras/SnakeFrame/index.js
```diff
@@ -106,7 +106,7 @@ export class SnakeFrame extends React.Component {
         <SplashContents>
           <h2>Snake game!</h2>
           <InitialStartButton backgroundColor={foodColor} onClick={this.play}>
-            Start game!
+            Start the game!
```
linuslundahl: THE GAME!

PR #940, opened 2019-07-02
Make `:use <db-name>` case insensitive

oskarhane:
This is expected to fail the e2e test since they run against Neo4j 3.5.

### DIFF
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -184,24 +184,24 @@ const availableCommands = [
             'No multi db support detected.'
           )
         }
-        // Check if chosen db exists
-        const dbExists = getDatabases(store.getState())
-          .map(db => db.name)
-          .includes(dbName)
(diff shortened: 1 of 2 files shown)


src/shared/services/commandInterpreterHelper.js
```diff
@@ -184,24 +184,24 @@ const availableCommands = [
...
           )
         }
-        // Check if chosen db exists
-        const dbExists = getDatabases(store.getState())
-          .map(db => db.name)
-          .includes(dbName)
-        if (!dbExists) {
+        // Check if chosen db exists. Case insensitive.
+        const existingDb = getDatabases(store.getState()).find(
+          db => db.name.toLowerCase() === dbName.toLowerCase()
```
huboneo: Nitpick: You could use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) here.

PR #947, opened 2019-07-23
Use driver transaction functions and set connection timeout to 30000 if connecting to cloud

eve-bright:
Struggling to know how best to test these changes. boltConnection currently has no tests - so no obvious way to test the transaction function change. 

Similarly the connectionsDuck has tests but there is nothing actually checking what we pass to open connection. I could obviously add tests here to do that but not sure what the intended strategy is. Perhaps I should extract the connectionTimeoutHelper function and test this separately.

### DIFF
--- a/src/shared/modules/connections/connectionHelpers.js
+++ b/src/shared/modules/connections/connectionHelpers.js
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2002-2019 "Neo4j,"
+ * Neo4j Sweden AB [http://neo4j.com]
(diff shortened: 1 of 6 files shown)


eve-bright:
I decided to go ahead and extract connectionTimeoutHelper to enable it to be tested. 
I have also added tests in the connectionsDuck test to check that the timeout is being passed to openConnection.

oskarhane:
I'm closing this as it will be replaced by a new one once the generic solution is built.

PR #951, opened 2019-08-07
Import/export saved scripts

huboneo:
This PR heavily refactors the user favorites implementation, and adds support for importing/exporting saved scripts as a ZIP archive

## Key changes
- Replaces local user favorites implemtation with a new ui kit package `@relate-by-ui/saved-scripts` (https://github.com/neo4j-apps/relate-by-ui/pull/1)
- Removes old documents and folders reducers in favour of new `user-favorites`
- Folders are no longer an entity, instead just a side-effect of a script path
- Added epic that migrates old favorites to new structure, **and then clears old favorites from local and remote states**
- New favorites to browsersync
- Updated webpack conf to support symlinked npm packages (for development)

## Running this PR locally
- checkout https://github.com/neo4j-apps/relate-by-ui/pull/1
- `yarn install && yarn link` in `packages/saved-scripts`
- checkout this PR
- remove `@relate-by-ui/saved-scripts` from package.json and `yarn install`
- readd `@relate-by-ui/saved-scripts` to package.json 
-`yarn link @relate-by-ui/saved-scripts`
- `yarn start`

## Todo
- [x] Today the PR clears all firebase versions of old favorites. Change this to add one empty version but preserve earlier just in case
- [ ] I had to install `semantic-ui-css` and modify the webpack rules to get semantic UI css to work. We should discuss how to do this better
- [ ] I have tested switching between master and this branch to verify that old favorites are correctly synced. Can we do more to test these changes?
- [x] No unit tests are added here (although some are added in the relate-by-ui PR)
- [ ] Run existing tests once `@relate-by-ui/saved-scripts` is published


src/shared/modules/user-favorites/client.test.js
```diff
@@ -0,0 +1,190 @@
...
+
+      await expect(createManyUserFavorites(data)).resolves.toEqual(expected)
+      await expect(getAllUserFavorites()).resolves.toEqual(newState)
+    })
+
+    test('should add several favorites to localstate', async () => {
+      const data = [{ contents: 'bom' }, { contents: 'apa' }]
+      const expected = [
+        { contents: 'bom', id: 'apa' },
+        { contents: 'apa', id: 'apa' }
```
oskarhane: It looks kind of strange with two favorites having the same id here in the tests since the whole point of `uuid` is to generate unique id's.
Rather than mocking `uuid`, could you change the assertion to be something like `expect.any(String)` or similar.

Not a big deal, just a bit unexpected.

jest.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     '/dist/',
     '/node_modules/'
   ],
+  transformIgnorePatterns: [`/node_modules/(?!lodash-es|react-dnd|dnd-core)`],
```
oskarhane: 👍

src/browser/components/FileDrop/FileDrop.jsx
```diff
@@ -22,15 +22,20 @@ import React, { useState } from 'react'
...
 import { withBus } from 'react-suber'
 import SVGInline from 'react-svg-inline'
+import { every } from 'lodash-es'
 
 import * as editor from 'shared/modules/editor/editorDuck'
-import { addFavorite } from 'shared/modules/favorites/favoritesDuck'
+import {
+  addFavorite,
+  addManyFavorites
+} from '../../../shared/modules/user-favorites/user-favorites.duck'
```
oskarhane: Rather than traversing the file tree, could you use the `shared` alias here?

src/browser/components/FileDrop/FileDrop.jsx
```diff
@@ -40,12 +45,19 @@ import {
   StyledFileDropActionButton
 } from './styled'
 import icon from 'icons/task-list-download.svg'
+import { addScriptPathPrefix } from '@relate-by-ui/saved-scripts'
+import { BROWSER_FAVOURITES_NAMESPACE } from '../../../shared/modules/user-favorites/user-favorites.constants'
```
oskarhane: US English please :)

src/browser/components/FileDrop/file-drop.utils.js
```diff
@@ -0,0 +1,98 @@
...
+import JSZip from 'jszip'
+import {
+  addScriptPathPrefix,
+  getScriptDisplayName,
+  omitScriptPathPrefix
+} from '@relate-by-ui/saved-scripts'
+import {
+  BROWSER_FAVOURITES_NAMESPACE,
+  CYPHER_FILE_EXTENSION
+} from '../../../shared/modules/user-favorites/user-favorites.constants'
```
oskarhane: Use alias here as well

src/browser/components/FileDrop/file-drop.utils.js
```diff
@@ -0,0 +1,98 @@
...
+    )
+  )
+}
+
+/**
+ * Factory function returning a file to user script object mapper
+ * @param     {File}        file
+ * @return    {Function}            user scripts mapper
+ */
+function fileContentToFavoriteFactory (file) {
```
oskarhane: I saw that writing unit tests was on the todo-list, so this is just a reminder that we want tests for this utility function.

src/browser/modules/Sidebar/favorites.js
```diff
@@ -0,0 +1,56 @@
...
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { withBus } from 'react-suber'
+import { connect } from 'react-redux'
+
+import * as userFavorites from '../../../shared/modules/user-favorites/user-favorites.duck'
```
oskarhane: Use alias on these as well please

src/shared/modules/user-favorites/client/methods/create-many-user-favorites.js
```diff
@@ -0,0 +1,33 @@
...
+import { map } from 'lodash-es'
+
+import {
+  setUserFavoritesLocalState,
+  tryGetUserFavoritesLocalState
+} from '../../user-favorites.utils'
+
+export default async function createManyUserFavorites (data) {
+  const alreadySaved = tryGetUserFavoritesLocalState()
+  const toSave = map(data, item => ({ ...item, id: uuid() }))
```
oskarhane: If I remember correctly from us talking about this, the reason for using lodash's `map` rather that `data.map` was that lodash does the type checking for you, right?

src/shared/modules/user-favorites/client/methods/create-many-user-favorites.js
```diff
@@ -0,0 +1,33 @@
...
+
+import uuid from 'uuid/v4'
+import { map } from 'lodash-es'
+
+import {
+  setUserFavoritesLocalState,
+  tryGetUserFavoritesLocalState
+} from '../../user-favorites.utils'
+
+export default async function createManyUserFavorites (data) {
```
oskarhane: All of these client methods seems be `async` at this point, which seems unnecessary. 
I suspect this is for when / if we do any remote work in here, but can the async part be removed for now?

src/shared/modules/user-favorites/user-favorites.duck.js
```diff
@@ -0,0 +1,281 @@
+/*
```
oskarhane: Could you follow the duck file naming convention we already use here?
I like your naming, but we should be consistent.

src/shared/modules/user-favorites/user-favorites.utils.js
```diff
@@ -0,0 +1,286 @@
...
+} from './user-favorites.constants'
+import { getBrowserName } from '../../services/utils'
+
+/**
+ * Converts old user favorites into new structure
+ * @param     {Object[]}    oldFavorites
+ * @param     {Object[]}    oldFolders
+ * @return    {Object[]}                    new user favorites objects ("my scripts")
+ */
+export function mapOldFavoritesAndFolders (oldFavorites, oldFolders) {
```
oskarhane: A reminder to eventually add unit tests for these functions.

src/browser/components/FileDrop/FileDrop.jsx
huboneo: I can :) Generally I find these webpack aliases to be a bit of an anti-pattern but since this is already a norm in browser I'll update them

src/shared/modules/user-favorites/client/methods/create-many-user-favorites.js
huboneo: It's the opposite actually, I prefer lodash as I dont have to care about the type being passed. that and I personally find it easier to read than OOP

src/shared/modules/user-favorites/client/methods/create-many-user-favorites.js
huboneo: I guess we could but then we would have to make them async again later. I don't really see any downside leaving them async?

src/shared/modules/user-favorites/user-favorites.duck.js
huboneo: the reason i went use kebab-case is that it's generally recommended to use lowercase names. I'll update it :)

huboneo:
@oskarhane thank you for your feedback. I have updated the PR with your comments

src/shared/modules/user-favorites/client/methods/create-many-user-favorites.js
oskarhane: Fair enough.

src/shared/modules/user-favorites/user-favorites.duck.js
oskarhane: I agree, but we should make one big effort to rename all files at the same time.

huboneo:
Closing as we will use a smaller implementation for now

PR #954, opened 2019-08-12
Export :history

huboneo:
This PR adds support for exporting the result of `:history` as TXT.

### STR
- run `:history`
- Select "Export TXT" from export dropdown

### Screenshots
Screenshot 2019-08-13 at 10 43 14

Screenshot 2019-08-13 at 10 41 24

### DIFF
--- a/src/browser/modules/Stream/FrameTitlebar.jsx
+++ b/src/browser/modules/Stream/FrameTitlebar.jsx
@@ -88,7 +88,17 @@ class FrameTitlebar extends Component {
     const { frame } = this.props
 
     if (frame.type === 'history') {
-      const asTxt = frame.result.join(';\n\n')
+      const asTxt = frame.result
+        .map(result => {
+          const safe = `${result}`.trim()
+
+          if (safe.startsWith(':')) {
+            return safe
+          }
+
+          return safe.endsWith(';') ? safe : `${safe};`
+        })
+        .join('\n\n')
       const blob = new Blob([asTxt], {
         type: 'text/plain;charset=utf-8'
       })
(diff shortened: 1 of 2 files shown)


huboneo:
[Add download option for :history](https://trello.com/c/VpR3WHRj/29-add-download-option-for-history)

jexp:
Thanks a lot Hugo for working on this.

How about a plain text download? Just as semicolon and newline separated statements in a .cypher script file instead?

I guess CSV will be quite horrible with regard to excaped double quotes `"`?

huboneo:
@jexp Thank you for your feedback. I have updated the PR to export as TXT (:cmds are not valid cypher)

src/browser/modules/Stream/FrameTitlebar.jsx
```diff
@@ -81,28 +83,54 @@ class FrameTitlebar extends Component {
     })
     saveAs(blob, 'export.csv')
   }
+
+  exportTXT = () => {
+    const { frame } = this.props
+
+    if (frame.type === 'history') {
+      const asTxt = frame.result.join(';\n\n')
```
oskarhane: Could you check so the statement isn't already ending with a semicolon?
Also, client commands (statements starting with `:`) should / cannot end with a semicolon, only Cypher.

jexp:
Perfect thanks.

What's cool though even as :commands are not valid cypher they should still work
if you run them with cypher-shell or paste/d&d into neo4j-browser for multi-line mode.

src/browser/modules/Stream/FrameTitlebar.jsx
huboneo: done!

ryguyrg:
SWEET!!! Super helpful!!

PR #955, opened 2019-08-12
Give user more connection feedback + add connection timeout configuration

oskarhane:
This PR increases the default neo4j driver connection timeout to 30000ms and also exposes this value as a setting. Further, this PR adds a warning banner that alerts users when a server is taking more than 5000ms to respond.

### Screenshots
Setting
Screenshot 2019-08-13 at 14 10 05

Slow successful connection
bad
Really slow connection 
worse
Slow pending connection
pending


src/browser/modules/Main/main.hooks.js
```diff
@@ -0,0 +1,67 @@
...
+} from 'shared/modules/connections/connectionsDuck'
+
+/**
+ * Custom hook for detecting slow connections
+ * @param     {number}      connectionState
+ * @param     {string}      errorMessage
+ * @param     {number}      lastConnectionUpdate
+ * @return    {boolean[]}
+ */
+export function useSlowConnectionState ({
```
oskarhane: This looks good, but could you add some tests to demonstrate the expected behavior @huboneo?

src/shared/modules/connections/connectionsDuck.js
```diff
@@ -59,12 +60,14 @@ export const SWITCH_CONNECTION_FAILED = NAME + '/SWITCH_CONNECTION_FAILED'
 export const DISCONNECTED_STATE = 0
 export const CONNECTED_STATE = 1
 export const PENDING_STATE = 2
+export const CONNECTING_STATE = 'CONNECTING_STATE'
```
huboneo: @oskarhane I'm not really sure about this one. I decided to use a string to denote that it is not a driver status, but maybe that is confusing?

oskarhane: Yes, please update to follow the same pattern as the other. Thanks for pointing it out.

src/browser/modules/Main/main.hooks.js
huboneo: @oskarhane done!

huboneo:
@oskarhane I added test for the hook now. I had to create a separate test file for each use-case to avoid race conditions stemming from how JEST reads tests suites. Essentially due to `describe` being read during discovery different test groups would affect each others timers.

oskarhane:
Looks good @huboneo!
Let's wait for @eve-bright to give a 👍 before merging this.

src/browser/modules/Main/Main.jsx
```diff
@@ -72,11 +75,20 @@ const Main = React.memo(function Main (props) {
...
+          props.connectionState === PENDING_STATE && !showSlowConnectionBanner
+        }
+      >
         <WarningBanner data-testid='reconnectBanner'>
           Connection to server lost. Reconnecting...
         </WarningBanner>
       </Render>
+      <Render if={showSlowConnectionBanner}>
+        <WarningBanner data-testid='reconnectBanner'>
+          Server is taking long to respond...
```
eve-bright: I am not sure this message makes sense. 'Server is taking a long time to respond' might be better?

huboneo:
@oskarhane @eve-bright Thank you for your feedback. I updated the message as per comments

Screenshot 2019-08-15 at 13 00 37

src/browser/modules/Main/Main.jsx
huboneo: Agreed, updated!

kfreytag:
@oskarhane (cc: @eve-bright @huboneo) The user feedback @ 5 sec is really clever here, yet I wonder whether it's going to lead to support problems for DBaaS where latency / connection times can be slow. Could we do one of two things:

1/ Remove it
2/ Change it to show up at 10 seconds

kfreytag:
@oskarhane I've realized that my problem with the 5sec message is that it's presented as a WARN rather than as INFO. I think our intention here is to provide comforting feedback that things are still going on. If it were blue, I'd feel like it was helping me, making me feel comfortable that the Browser was still trying to connect. As it is, I feel like something's wrong.

Perhaps we should simply change the string below the password field that says "Connecting...". Something like "Still connecting..."?

huboneo:
@kfreytag @oskarhane I updated the PR to show an info banner after 5 seconds and warning banner first after 10 seconds

PR #956, opened 2019-08-13
Add URL param support for populating editor with Cypher and parameters

oskarhane:
Add newline for each query. 
Semi-colons are up to the user to add.

Example usages:  
`http://localhost:7474/?cmd=edit&arg=RETURN%201;&arg=RETURN%20rand();`  
`http://localhost:7474/?cmd=param&arg=x%20%3d%3E%201;`  
`http://localhost:7474/?cmd=params&arg={x:1.0,%20y:%22hello%22}`  

Since it's not limited to Cypher, but works for populating the editor with anything the command is was decided to be `edit`.

Cypher:  
Screenshot 2020-01-07 16 11 56

Param:  
oskarhane-mbpt 2019-08-13 at 15 02 56

Params:  
oskarhane-mbpt 2019-08-13 at 15 06 05

### DIFF
--- a/src/shared/modules/editor/editorDuck.js
+++ b/src/shared/modules/editor/editorDuck.js
@@ -33,7 +33,7 @@ export const NOT_SUPPORTED_URL_PARAM_COMMAND = `${NAME}/NOT_SUPPORTED_URL_PARAM_
 // Supported commands
 const validCommandTypes = {
   play: (cmdchar, args) => `${cmdchar}play ${args.join(' ')}`,
-  cypher: (_, args) => args.join('\n'),
+  edit: (_, args) => args.join('\n'),
   param: (cmdchar, args) => `${cmdchar}param ${args.join(' ')}`,
   params: (cmdchar, args) => `${cmdchar}params ${args.join(' ')}`
 }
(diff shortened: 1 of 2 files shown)


Dcanzano:
will this be included in some `:help cmd` online help

oskarhane:
I've updated this PR by adding support for 'param' and 'params' as well @huboneo.
See updated description + screenshots.

I'll add something on that note @Dcanzano, thanks for the reminder.

src/shared/modules/editor/editorDuck.js
```diff
@@ -33,7 +33,9 @@ export const NOT_SUPPORTED_URL_PARAM_COMMAND = `${NAME}/NOT_SUPPORTED_URL_PARAM_
 // Supported commands
 const validCommandTypes = {
   play: (cmdchar, args) => `${cmdchar}play ${args.join(' ')}`,
-  cypher: (_, args) => args.join('\n')
+  cypher: (_, args) => args.join('\n'),
+  param: (cmdchar, args) => `${cmdchar}param ${args.join(' ')}`,
+  params: (cmdchar, args) => `${cmdchar}params ${args.join(' ')}`
```
huboneo: do we need to worry about interoperability here? can `:play` accept params? Also what happens if provided params are malformed/invalid?

akollegger:
WDYT about using `edit` as the command, as the explicit user action is to populate the editor with the attached text?

PR #965, opened 2019-09-12
Experiment with bundles to get e2e tests to pass

oskarhane:


e2e_tests/integration/0.index.spec.js
```diff
@@ -155,7 +155,7 @@ describe('Neo4j Browser', () => {
     cy.get('[data-testid="clearLocalData"]').click()
 
     cy.get('[data-testid="drawerFavorites"]').click()
-    cy.get('[data-testid="sidebarFavoriteItem"]').should('have.length', 0)
```
huboneo: Oops, sorry about that. Good catch!

PR #969, opened 2019-09-17
Use GraphQL version of Relate-API when available

oskarhane:
Neo4j Browser should behave just as before, but using the GraphQL version rather than the injected version of the relate-api.
We're using the Apollo lib as our client.

The GraphQL queries are loaded on build so we don't need to process AST's on the client side, see https://www.apollographql.com/docs/react/recipes/webpack/.

- [x] No changes when operating standalone
- [x] Automatically connect to running graph in Neo4j Desktop
- [x] Automatically switch to connect to the active graph when user starts a different one in Neo4j Desktop and has Neo4j Browser open
- [x] ~Adapt to color theme preferences set in macOS (not working yet, might not be implemented in Neo4j Desktop, need to check)~ (not implemented on desktop side)
- [x] Update e2e tests (removed them)*

Note: We're still using one function from the injected API and that's `neo4jDesktopApi.getKerberosTicket` since it's not ported to the GraphQL API yet.

* The relate-api will provide a graphql server that can be used for E2E test for graph apps. E2E tests for the desktop env will be added again when that is available.


src/browser/components/relate-api/relate-api.utils.js
```diff
@@ -0,0 +1,236 @@
...
+  const { principals = null } = graph.connection
+  if (!principals) {
+    return null
+  }
+  if (!principals.protocols) {
+    return null
+  }
+  if (typeof principals.protocols[protocol] === 'undefined') {
+    return null
+  }
```
huboneo: these if statements could be simplified with lodash [_.has()](https://lodash.com/docs/4.17.15#has) 
```js
if (_has(graph, `connection.principals.protocols[${protocol}]`)) ...
```

src/browser/components/relate-api/relate-api.utils.js
```diff
@@ -0,0 +1,236 @@
...
+  if (
+    !principals.authenticationMethods ||
+    !principals.authenticationMethods.kerberos
+  ) {
+    return false
+  }
+  if (!principals.authenticationMethods.kerberos.enabled) {
+    return false
+  }
```
huboneo: here `_.has()` could also help

src/browser/components/relate-api/workspace.graphql
```diff
@@ -0,0 +1,76 @@
+query {
```
huboneo: (つ◕౪◕)つ━☆ﾟ.*･｡ﾟ nice!

src/browser/components/relate-api/relate-api.utils.js
```diff
@@ -0,0 +1,236 @@
+/*
```
huboneo: maybe add some unit tests to some of these utils?

src/browser/modules/App/App.jsx
```diff
@@ -136,33 +120,16 @@ export function App (props) {
 
   return (
     <ErrorBoundary>
+      <RelateApi
+        defaultConnectionData={defaultConnectionData}
+        setEnvironmentTheme={setEnvironmentTheme}
+      />
```
huboneo: Personally I dont mind this pattern, but in a way it would be clearer if `RelateApi` had children to clarify dependents. Up to you!

PR #971, opened 2019-09-27
Adapt to breaking "change password" server changes 

oskarhane:
Refactor how we change passwords + adapt to server changes.

This has a fake server response when setting the initial password until we can get the server version directly from the driver with auth.

This is expected to fail on the E2E tests.

Description:
Check server version before changing password. If multi-db, use new Cypher commands against system db.
We have two cases for changing pw: 
  1. Initial startup user needs to change the initial `neo4j` pw
  2. When using `:server change-password`

Code changes described:

1. Refactor so rather than generating the change password query in `ConnectionForm`, we generate the query back in `cypherDuck` after server version has been checked.

2. Refactor fetching server info on `dbMetaDuck` to separate epic and also pull out `updateServerInfo` function to parse server responses (to be used `handleForcePasswordChangeEpic` in `cypherDuck` above as well) to get version, edition etc.

3. Rename a terribly named `adHocSession` -> `queryAndResovle` and don't have it responsible for closing driver connections. It's used in `handleForcePasswordChangeEpic` in `cypherDuck`

4. Create new epic for validating credentials only, without replacing the base connection. It's named `validateConnectionCredentials` and lives in `connectionsDuck`

5. Small UI changes on multi-db frames after master was merged into this.


src/shared/modules/cypher/cypherDuck.js
```diff
@@ -188,8 +200,39 @@ export const handleForcePasswordChangeEpic = (some$, store) =>
...
+          // Let's establish what server version we're connected to if not in state
+          if (!getVersion(store.getState())) {
+            const versionRes = await queryAndResovle(
+              driver,
+              { ...action, query: serverInfoQuery, parameters: {} },
+              undefined
+            )
+            // It not successful, fake it for now until driver supports
+            // getting server version without auth
+            if (!versionRes.success) {
```
oskarhane: This would change when driver support something like `driver.supportsMultidb()`

oskarhane:
Thanks @huboneo!
I've updated the description to make the intent more clear + code changes.

PR #972, opened 2019-09-30
Default scheme `bolt://` -> `neo4j://` in canary

oskarhane:
"Default scheme" = The scheme added by neo4j-browser if the user didn't enter a scheme themselves.
Cases like when the connect URL is entered as `localhost`, `localhost:7687` etc.

Notes:
This would need to be changed before merging canary into master since it's now hard coded in some places.
We need to figure out what scheme to use when not connected yet before merging canary into master.

For now we'll have to live with a disconnected master / canary.

### DIFF
--- a/src/shared/modules/features/versionedFeatures.js
+++ b/src/shared/modules/features/versionedFeatures.js
@@ -80,6 +80,17 @@ export const getUsedDbName = state => {
   return getActiveDbName(state)
 }
 
+export const getDefaultBoltScheme = serverVersion => {
+  const pre4 = 'bolt://'
+  if (!semver.valid(serverVersion)) {
+    return pre4
+  }
+  if (semver.gte(serverVersion, NEO4J_4_0)) {
+    return 'neo4j://'
+  }
+  return pre4
+}
+
 export const changeUserPasswordQuery = (state, oldPw, newPw) => {
   const pre4 = {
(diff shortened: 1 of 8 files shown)


huboneo:
Tested locally.

Browser correctly defaults to `neo4j://` when no protocol present.
Screenshot 2019-09-30 at 11 43 27
Screenshot 2019-09-30 at 11 43 34

Browser no longer able to default to `bolt://` for 3.X instances
Screenshot 2019-09-30 at 11 48 20

src/shared/modules/features/versionedFeatures.js
```diff
@@ -80,6 +80,17 @@ export const getUsedDbName = state => {
   return getActiveDbName(state)
 }
 
+export const getDefaultBoltScheme = serverVersion => {
```
linuslundahl:
```
export const getDefaultBoltScheme = serverVersion =>
  semver.valid(serverVersion) && semver.gte(serverVersion, NEO4J_4_0) ? 'neo4j://' : 'bolt://'
```
Maybe?

oskarhane: Maybe. Makes it a bit harder to read, but not too bad. Merging without any changes because it follows the same pattern as other functions in that file.
But a candidate for a future refactor, for sure!

Thanks!

PR #973, opened 2019-09-30
Prioritize bold_direct over bolt in discovery

huboneo:
This PR adds support for the new `bolt_direct` field in discovery responses, prioritizing it over the old `bolt` field.


huboneo:
@oskarhane thank you for your feedback. I have updated the PR

PR #974, opened 2019-09-30
Server user add UI fixes

linuslundahl:
Fixes some style and UX issues with `server user add` and `server user list`.

Before:
image

After:
image

### DIFF
--- a/src/browser/modules/Frame/styled.jsx
+++ b/src/browser/modules/Frame/styled.jsx
@@ -87,7 +87,7 @@ export const StyledFrameMainSection = styled.div`
 `
 
 export const StyledFrameAside = styled.div`
-  flex: 1 0 25%;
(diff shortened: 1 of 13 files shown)


src/browser/modules/Frame/styled.jsx
```diff
@@ -87,7 +87,7 @@ export const StyledFrameMainSection = styled.div`
 `
 
 export const StyledFrameAside = styled.div`
-  flex: 1 0 25%;
+  flex: 0 0 25%;
```
oskarhane: This change is scary. I haven't review it visually yet so it might be fine, but this is a major change. :)

linuslundahl: It might look scary, but the Aside (which I added in the major PR we merged a few weeks ago) is not supposed to grow.

PR #979, opened 2019-10-24
Use GraphQL version of Relate-API when available (in Neo4j Desktop)

oskarhane:
Replacement of the reverted https://github.com/neo4j/neo4j-browser/pull/969

Neo4j Browser should behave just as before, but using the GraphQL version rather than the injected version of the relate-api.
We're using the Apollo lib as our client.

The GraphQL queries are loaded on build so we don't need to process AST's on the client side, see https://www.apollographql.com/docs/react/recipes/webpack/.

- [x] No changes when operating standalone
- [x] Automatically connect to running graph in Neo4j Desktop
- [x] Automatically switch to connect to the active graph when user starts a different one in Neo4j Desktop and has Neo4j Browser open
- [x] ~Adapt to color theme preferences set in macOS (not working yet, might not be implemented in Neo4j Desktop, need to check)~ (not implemented on desktop side)
- [x] Update e2e tests (removed them)*

Note: We're still using one function from the injected API and that's `neo4jDesktopApi.getKerberosTicket` since it's not ported to the GraphQL API yet.

* The relate-api will provide a graphql server that can be used for E2E test for graph apps. E2E tests for the desktop env will be added again when that is available.

~Should not be merged until Neo4j Desktop implements the relate-api correctly.~


oskarhane:
You reviewed the original PR (linked in description) for this @huboneo, could I ask you to have a look at this new one as well?

src/browser/components/relate-api/relate-api.hooks.jsx
```diff
@@ -0,0 +1,93 @@
+import { useState, useEffect } from 'react'
```
huboneo: no need to use `.jsx` for this file?

src/browser/components/relate-api/relate-api.jsx
```diff
@@ -0,0 +1,104 @@
+/*
```
huboneo: no need for `.jsx`

src/browser/modules/DBMSInfo/__snapshots__/MetaItems.test.js.snap
```diff
@@ -11,7 +11,7 @@ exports[`LabelItems renders empty 1`] = `
       Node Labels
     </h5>
     <div
-      class="Drawer__DrawerSectionBody-sc-5bct1g-5 RMcRs undefined"
+      class="undefined Drawer__DrawerSectionBody-sc-5bct1g-5 RMcRs"
```
huboneo: :/

oskarhane: ugh, ugly. Fixed.

src/browser/components/relate-api/relate-api.hooks.jsx
oskarhane: 👍

src/browser/components/relate-api/relate-api.jsx
oskarhane: 👍

PR #980, opened 2019-10-28
Retry reconnection on pw-change

oskarhane:
If the new password successfully was set when changing password, but it’s not accepted when trying to connect, retry up to 5 times before failing.

Also, when validating old credentials, don’t reconstruct the main driver instance. Create a new one and validate on that one.

### DIFF
--- a/src/shared/services/bolt/boltConnection.js
+++ b/src/shared/services/bolt/boltConnection.js
@@ -283,7 +283,7 @@ export const closeConnection = () => {
 }
 
 export const ensureConnection = (props, opts, onLostConnection) => {
-  const session = _drivers ? _drivers.getRoutedDriver().session() : false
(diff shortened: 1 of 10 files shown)


src/shared/services/bolt/boltConnection.js
```diff
@@ -283,7 +283,7 @@ export const closeConnection = () => {
 }
 
 export const ensureConnection = (props, opts, onLostConnection) => {
-  const session = _drivers ? _drivers.getDirectDriver().session() : false
+  const session = _drivers ? _drivers.getRoutedDriver().session() : false
```
huboneo: @oskarhane whats the reason for this change?

oskarhane: No reason :)
Removing that.

PR #988, opened 2019-11-13
test: add user

pontusnyman:
Tests creates an user with/without force password change
Then checks if user is added to user list.


linuslundahl:
Should test both with and without forced password change.

linuslundahl:
Good stuff! 👍

PR #990, opened 2019-11-21
User handling, bolt+routing, :schema fixes

huboneo:
This PR addresses changes to the `:schema` command, handling of depcrecated `bolt+routing` protocol,  as well as changes to user handling in anticipation of 4.0
- Fixed JS exception when calling `:schema` with created indexes
- Aliased `bolt+routing://` to `neo4j://`
- Added feature switch for new RBAC commands when on 4.0

### Screenshots
Screenshot 2019-11-21 at 12 42 58

### DIFF
--- a/src/browser/modules/User/UserAdd.jsx
+++ b/src/browser/modules/User/UserAdd.jsx
@@ -21,6 +21,7 @@ import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { withBus } from 'react-suber'
 import uuid from 'uuid'
+import { map } from 'lodash-es'
 
 import { executeCommand } from 'shared/modules/commands/commandsDuck'
 import { canAssignRolesToUser } from 'shared/modules/features/featuresDuck'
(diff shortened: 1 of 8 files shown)


src/browser/modules/User/UserAdd.jsx
```diff
@@ -337,7 +352,7 @@ export class UserAdd extends Component {
         aside={
           <FrameAside
             title='Add user'
-            subtitle='Add a user to the current databse'
+            subtitle='Add a user to the current database'
```
oskarhane: 😱

src/shared/modules/cypher/boltUserHelper.js
```diff
@@ -18,31 +18,65 @@
...
-  username,
-  password,
-  forcePasswordChange
-}) {
+export function createDatabaseUser (
+  { username, password, forcePasswordChange },
+  is40 = false
+) {
+  if (is40) {
+    return `CREATE USER ${username} SET PASSWORD "${password}" CHANGE ${
```
oskarhane: wdyt about using params when setting the password? 
I see you followed the pattern already in the existing code, but what happens if a user has a `"` in the password?

huboneo:
@oskarhane I have updated the PR with your changes. Unfortunately the new RBAC syntax only supports parametrization of password

PR #998, opened 2019-11-25
Handle initial db selection

oskarhane:
All users can get a list of all databases (incl. info on which is the default), but not all users can access all databases.

When connecting to the dbms, all users will start out using the system default db.
If they try to run cypher but don't have access, an error message is shown with an addition which a link to list all available databases to help the user find it's way.
The `"List available databases"` links to `:dbs`.
_Prior art: We do this when a users inputs a unknown procedure_

See screenshot.

oskarhane-mbpt 2019-11-25 at 20 52 14

I also reworked so we don't show error message when a user is trying to `:use xx` a db that doesn't exist but let the system handle that when the user tries to exec a command.

E2E tests added to verify the behavior.

### DIFF
--- a/src/browser/modules/DBMSInfo/DatabaseSelector.jsx
+++ b/src/browser/modules/DBMSInfo/DatabaseSelector.jsx
@@ -32,9 +32,10 @@ const Select = styled.select`
   color: ${props => props.theme.inputText};
 `
 
+const EMPTY_OPTION = 'Select db to use'
+
 export const DatabaseSelector = ({
   databases = [],
-  defaultDb = 'neo4j',
   selectedDb = '',
   onChange = () => {}
 }) => {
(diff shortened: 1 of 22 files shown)


src/browser/modules/DBMSInfo/DatabaseSelector.jsx
```diff
@@ -32,18 +32,25 @@ const Select = styled.select`
...
     return null
   }
   const selectionChange = ({ target }) => {
+    if (target.value === EMPTY_OPTION) {
+      return
+    }
     onChange(target.value)
   }
+  if (!selectedDb) {
+    databases.unshift({ name: EMPTY_OPTION, status: null })
```
huboneo: this is mutating the array, consider using an immutable alternative

oskarhane: 👍

PR #999, opened 2019-11-25
Fix errors in community edition

oskarhane:
Don’t show content on frames that can’t be used (see screenshot).
And make the dbms sidebar not crash when opened.

oskarhane-mbpt 2019-11-25 at 21 50 33


src/browser/modules/DBMSInfo/UserDetails.jsx
```diff
@@ -31,10 +31,10 @@ import { StyledTable, StyledKey, StyledValue, Link } from './styled'
 export class UserDetails extends Component {
   render () {
     const userDetails = this.props.user
+    const roles = userDetails && userDetails.roles ? userDetails.roles : []
```
huboneo: good usecase for lodash `_.get()`

PR #1003, opened 2019-12-02
Neo4j-Browser becomes unresponsive with large query results

huboneo:
This PR introduces a new configuration option in browser for truncating max field length before they are passed to rendering layers. The reason for this is that certain queries can produce a single result with a lot of values in a given field, which we had yet to account for.

### Screenshot
Screenshot 2019-12-12 at 13 01 29
Screenshot 2020-01-07 at 12 30 25

### DIFF
--- a/src/browser/modules/D3Visualization/components/Inspector.jsx
+++ b/src/browser/modules/D3Visualization/components/Inspector.jsx
@@ -40,6 +40,9 @@ import { GrassEditor } from './GrassEditor'
 import { RowExpandToggleComponent } from './RowExpandToggle'
 import ClickableUrls from '../../../components/clickable-urls'
 import numberToUSLocale from 'shared/utils/number-to-US-locale'
+import { StyledTruncatedMessage } from 'browser/modules/Stream/styled'
+import { Icon } from 'semantic-ui-react'
+import Ellipsis from 'browser-components/Ellipsis'
 
 const mapItemProperties = itemProperties =>
   itemProperties
@@ -149,6 +152,12 @@ export class InspectorComponent extends Component {
           <StyledInlineList className="list-inline">
             <StyledInspectorFooterRowListPair className="pair" key="pair">
               <StyledInspectorFooterRowListValue className="value">
+                {this.props.hasTruncatedFields && (
+                  <StyledTruncatedMessage>
+                    <Icon name="warning sign" /> Result fields have been
+                    truncated.&nbsp;
+                  </StyledTruncatedMessage>
+                )}
                 {description}
               </StyledInspectorFooterRowListValue>
             </StyledInspectorFooterRowListPair>
--- a/src/browser/modules/D3Visualization/components/Inspector.jsx
+++ b/src/browser/modules/D3Visualization/components/Inspector.jsx
@@ -154,7 +154,7 @@ export class InspectorComponent extends Component {
               <StyledInspectorFooterRowListValue className="value">
                 {this.props.hasTruncatedFields && (
                   <StyledTruncatedMessage>
-                    <Icon name="warning sign" /> Result fields have been
+                    <Icon name="warning sign" /> Record fields have been
                     truncated.&nbsp;
                   </StyledTruncatedMessage>
                 )}
--- a/src/browser/modules/Sidebar/Settings.jsx
+++ b/src/browser/modules/Sidebar/Settings.jsx
@@ -161,9 +161,9 @@ const visualSettings = [
       },
       {
         maxFieldItems: {
-          displayName: 'Property item limit',
+          displayName: 'Record field limit',
           tooltip:
-            'Max amount of items shown for list properties. When reached, lists get truncated.'
+            'Max number of items in a field per record being handled. When reached, items get truncated.'
         }
       },
       {
--- a/src/browser/modules/Sidebar/Settings.jsx
+++ b/src/browser/modules/Sidebar/Settings.jsx
@@ -161,9 +161,9 @@ const visualSettings = [
       },
       {
         maxFieldItems: {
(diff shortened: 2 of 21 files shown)


src/shared/services/bolt/bolt.js
```diff
@@ -288,6 +288,8 @@ export default {
   ) => {
     const intChecker = neo4j.isInt
     const intConverter = val => val.toString()
+    return { nodes: [], relationships: [] }
```
huboneo: Remove this, WIP

oskarhane (commented):
Some really good stuff in here.

Some comments: 

- I get somewhat inconsistent results. If you have the movies database and set the "Max field items" to `1` and execute `MATCH p=()-[r:ACTED_IN]->() RETURN p LIMIT 25` I see two nodes in the viz. 
But switching to table view I see 25 node objects.

- The limitations we already have with "Max Rows" in table view + "Initial node display" in viz, maybe those should be removed and replaced by this one (since this operates at a lower level)?

- I don't see a truncation message in the viz's statusbar.  

- Maybe not part of this PR, but just to point it out, property size is also a performance killer (try returning a node with a property of 1MB).

huboneo:
@oskarhane thank you for your feedback.
- Indeed there was a bug truncation.
- Added a message to Vis for truncated messages
- Updated setting display name
- Unsure about removing max row truncation as that is on a different axis from fields. Lets leave it as is for now?

src/browser/modules/Stream/CypherFrame/CodeView.jsx
```diff
@@ -54,16 +57,33 @@ class ExpandableContent extends Component {
...
+  render () {
+    const { request = {}, query, maxFieldItems } = this.props
     if (request.status !== 'success') return null
-    const resultJson = JSON.stringify(request.result.records, null, 2)
-    const summaryJson = JSON.stringify(request.result.summary, null, 2)
+    const resultJson = JSON.stringify(
+      request.result.records,
+      fieldLimiterFactory(maxFieldItems),
+      2
+    )
```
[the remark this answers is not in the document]
huboneo: Unfortunately not, even `JSON.stringify()` gets overwhelmed with large result fields

src/browser/modules/Stream/CypherFrame/TableView.jsx
```diff
@@ -190,14 +201,30 @@ export class TableStatusbar extends Component {
...
   }
 
   render() {
     return (
       <StyledStatsBar>
-        <Ellipsis>{this.state.statusBarMessage}</Ellipsis>
+        <Ellipsis>
+          {this.state.hasTruncatedFields && (
+            <StyledTruncatedMessage>
+              <Icon name='warning sign' /> Result fields have been
```
oskarhane: "Result fields" => "Records fields"

src/browser/modules/Sidebar/Settings.jsx
```diff
@@ -159,6 +159,13 @@ const visualSettings = [
           tooltip: "Max number of rows to render in 'Rows' result view"
         }
       },
+      {
+        maxFieldItems: {
+          displayName: 'Property item limit',
```
oskarhane: "Record field limit" describes this better imo. wdyt?

src/browser/modules/Sidebar/Settings.jsx
```diff
@@ -159,6 +159,13 @@ const visualSettings = [
...
+          tooltip:
+            'Max amount of items shown for list properties. When reached, lists get truncated.'
```
oskarhane: "Max number of fields per record being handled" os more clear. wdyt?

src/browser/modules/D3Visualization/components/Inspector.jsx
```diff
@@ -149,6 +152,12 @@ export class InspectorComponent extends Component {
           <StyledInlineList className="list-inline">
             <StyledInspectorFooterRowListPair className="pair" key="pair">
               <StyledInspectorFooterRowListValue className="value">
+                {this.props.hasTruncatedFields && (
+                  <StyledTruncatedMessage>
+                    <Icon name="warning sign" /> Result fields have been
```
oskarhane: "Result fields" => "Records fields"

oskarhane (commented):
Thanks for the update on this.

## Result bug
I still get mismatches between the different view though.
To reproduce:

1. Load the common movie dataset
1. Set the limit to 1
1. Run `MATCH p=(n:Movie)-[]-(:Person) RETURN p LIMIT 25` and compare the views. The table view only sees one item per row while viz and text show the whole path.
See:
Screenshot 2020-01-13 09 27 13

## Phrasing
I've re-tought the phrasing as well and I think it's ok to use the word "fields" as this is the first time we describe the records width to users. So sorry about the back-and-forth on that.
I've suggest what wording to use via comments.

huboneo:
@oskarhane thank you again for your feedback. I have updated the PR. Tests are still failing I believe

huboneo:
Screenshot 2020-01-13 at 10 51 47

oskarhane (commented):
Thanks for the update!

I still see different results in table vs text view. They should be the same. The viz does some magic with internal relationships, but not the different table views.

1. Load movie dataset
1. Set limit to 1
1. Run `MATCH p=()-[r:FOLLOWS]->() RETURN p LIMIT 25` and compare table and text views.

Screenshot 2020-01-13 12 23 11


👍 on the wording changes.

oskarhane:
Stable version of react-table is now out @huboneo. When yo have time, could you update this upstream and update this PR?

huboneo:
closing in favour of https://github.com/neo4j/neo4j-browser/pull/1090

PR #1004, opened 2019-12-04
Error message when declaring params on system db

oskarhane:
Show client side error message when trying to set parameters when using system database.
Links to `:help params` and `:help param` which both has more info on this..
Added E2E test for it.

Screenshot 2019-12-04 10 51 19

### DIFF
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -131,7 +131,12 @@ const availableCommands = [
     name: 'set-params',
     match: cmd => /^params?\s/.test(cmd),
     exec: function (action, cmdchar, put, store) {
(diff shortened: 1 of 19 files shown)


src/shared/services/commandInterpreterHelper.js
```diff
@@ -152,10 +157,11 @@ const availableCommands = [
           if (!action.parentId) {
             put(
               frames.add({
+                useDb: getUseDb(store.getState()),
                 ...action,
                 error: {
                   type: 'Syntax Error',
-                  message: error.message.substring('Error: '.length)
```
huboneo: 👍

oskarhane:
Let's merge into release branch instead.

PR #1006, opened 2019-12-04
Only show each database once in :dbs output

huboneo:
### Screenshot
Screenshot 2019-12-04 at 16 20 43


OskarDamkjaer:
(note: duplicate databases can be shown on clusters as they are listed once per cluster member)

PR #1008, opened 2019-12-04
Format node/relationship counts in sidebar, inspector and legend

jk05:
fixes: #953

- Adds in thousands comma separators for numbers (shown in screenshot)
- Naming suggestions very welcome as Im not sure what to call the util
- Ive called it mostly inline but happy to go with assigning it to a variable first for readability, depending on how we do things in Neo4j

Screenshot 2019-12-04 at 10 05 48

### DIFF
--- a/src/browser/modules/D3Visualization/components/Inspector.jsx
+++ b/src/browser/modules/D3Visualization/components/Inspector.jsx
@@ -39,7 +39,7 @@ import {
 import { GrassEditor } from './GrassEditor'
 import { RowExpandToggleComponent } from './RowExpandToggle'
 import ClickableUrls from '../../../components/clickable-urls'
-import numberToUSLocale from '../../../../shared/utils/number-to-US-locale'
+import numberToUSLocale from 'shared/utils/number-to-US-locale'
 
 const mapItemProperties = itemProperties =>
   itemProperties
--- a/src/browser/modules/D3Visualization/components/Inspector.jsx
+++ b/src/browser/modules/D3Visualization/components/Inspector.jsx
@@ -39,6 +39,7 @@ import {
 import { GrassEditor } from './GrassEditor'
 import { RowExpandToggleComponent } from './RowExpandToggle'
 import ClickableUrls from '../../../components/clickable-urls'
+import thousandsCommaSeparatedNumber from '../../../../shared/utils/thousands-separated-number'
 
 const mapItemProperties = itemProperties =>
   itemProperties
@@ -141,7 +142,11 @@ export class InspectorComponent extends Component {
           </StyledInlineList>
         )
       } else if (type === 'canvas') {
-        const description = `Displaying ${item.nodeCount} nodes, ${item.relationshipCount} relationships.`
+        const description = `Displaying ${thousandsCommaSeparatedNumber(
+          item.nodeCount
+        )} nodes, ${thousandsCommaSeparatedNumber(
+          item.relationshipCount
+        )} relationships.`
         inspectorContent = (
           <StyledInlineList className='list-inline'>
             <StyledInspectorFooterRowListPair className='pair' key='pair'>
--- a/src/browser/modules/D3Visualization/components/Inspector.jsx
+++ b/src/browser/modules/D3Visualization/components/Inspector.jsx
@@ -42,9 +42,8 @@ import ClickableUrls from '../../../components/clickable-urls'
 
 const mapItemProperties = itemProperties =>
   itemProperties
-    .sort(
-      ({ key: keyA }, { key: keyB }) =>
-        keyA < keyB ? -1 : keyA === keyB ? 0 : 1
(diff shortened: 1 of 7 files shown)


src/browser/modules/D3Visualization/components/Inspector.jsx
```diff
@@ -39,12 +39,12 @@ import {
 import { GrassEditor } from './GrassEditor'
 import { RowExpandToggleComponent } from './RowExpandToggle'
 import ClickableUrls from '../../../components/clickable-urls'
+import thousandsCommaSeparatedNumber from '../../../../shared/utils/thousands-separated-number'
```
jk05: I'm struggling to name this so any suggestions very welcome

oskarhane: `numberToUsLocale` maybe?

jk05: yeah thats more descriptive of what its actually doing.

src/browser/modules/DBMSInfo/MetaItems.jsx
```diff
@@ -64,7 +65,7 @@ const createItems = (
   if (showStar) {
     let str = '*'
     if (count) {
-      str = `${str}(${count})`
+      str = `${str}(${thousandsCommaSeparatedNumber(count)})`
```
jk05: I did this here as I assume we would always want to comma separate whenever `createItems` is called...

src/shared/utils/thousands-separated-number.js
```diff
@@ -0,0 +1,21 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+export default value =>
```
jk05: Im going to have another look at this, but the general thought process is for defensive purposes but I dont know how defensive we need to be here
- `parseInt` to coerce everything to numbers, but need to do the `isNaN` check for null etc

It reads a bit odd anyway but open to suggestions, I will take another look in a bit

src/shared/utils/number-to-US-locale.js
```diff
@@ -0,0 +1,19 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+export default value =>
+  (parseInt(value, 10) && parseInt(value, 10).toLocaleString('en-US')) || value
```
jk05: thought process here is to be defensive against anything where `parseInt` returns `NaN` and just return the value passed in as it's a separate issue. Any other ways to do this or suggestions or comments to be more in in line with Neo4j patterns is welcome

src/browser/modules/D3Visualization/components/Inspector.jsx
```diff
@@ -39,12 +39,12 @@ import {
 import { GrassEditor } from './GrassEditor'
 import { RowExpandToggleComponent } from './RowExpandToggle'
 import ClickableUrls from '../../../components/clickable-urls'
+import numberToUSLocale from '../../../../shared/utils/number-to-US-locale'
```
oskarhane: You should be able to import from `services/utils/number-to-US-locale`. Maybe we need to adjust our aliases if this doesn't work for you.

src/shared/utils/number-to-US-locale.js
oskarhane: That looks good to me

src/shared/utils/number-to-US-locale.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // Given
+    value = 987654312345
+
+    // When
+    returnValue = numberToUSLocale(value)
+
+    // Then
+    expect(returnValue).toBe('987,654,312,345')
+  })
```
oskarhane: Great to se tests!
I think it could be useful to add tests for numbers that you don't expect to be changed as well, like `1` and `100`.
If we ever change the implementation we know we're not introducing a regression on on-formatted numbers.

src/browser/modules/D3Visualization/components/Inspector.jsx
jk05: ah Ill give it a go, didnt see that

jexp:
would it be possible to make the formatting rendering only? so that when someone copies the number it's just the raw number?


also it would be amazing to add the counts to the other labels and rel-types too

PR #1010, opened 2019-12-05
Suggest strong passwords in change-password frame

huboneo:
This PR adds the ability for users to generate random sentence passwords to help them create secure credentials.

### Screenshots
Screenshot 2019-12-09 at 08 15 28


oskarhane (commented):
Looks good!
Really useful change actually.

The only thing I think behaves a little strange is that the "Suggest password" button changes width with the width of the suggested password field. I think it would look better if the button width was fixed.

huboneo:
@oskarhane thank you for your feedback. I've updated the PR with a new design should also address the UI bug you found.

oskarhane (approved):
🌷

PR #1014, opened 2019-12-10
Remove standardjs and lint with eslint only

oskarhane:
Keep formatting with prettier.
Upgraded a bunch of other dependencies as well.

I had to disable (or warn) on a bunch of eslint rules from the react-recommended rule suite.
Not ideal but something we can work on separately.


src/browser/AppInit.jsx
```diff
@@ -90,14 +90,14 @@ const AppInit = () => {
   return (
     <Provider store={store}>
       <BusProvider bus={bus}>
-        <React.Fragment>
+        <>
```
jk05: I know these are interchangeable, so do we generally prefer `<>` to `<React.Fragment`?

oskarhane: Yeah, the linter will fix it automatically. Do you mind that rule?

package.json
```diff
@@ -20,12 +20,12 @@
...
   "main": "dist/index.html",
   "scripts": {
-    "build": "NODE_ENV=\"production\" webpack --config ./build_scripts/webpack.config.js --progress",
+    "build": "NODE_ENV=\"production\" webpack --config ./build_scripts/webpack.config.js",
     "dev": "jest --watch",
     "e2e": "cypress run",
     "e2e-local": "CYPRESS_E2E_TEST_ENV=\"local\" cypress run",
     "e2e-local-open": "CYPRESS_E2E_TEST_ENV=\"local\" cypress open",
-    "format": "prettier-eslint 'src/**/!(*.min).js' 'src/**/*.jsx' --write",
+    "format": "prettier-eslint --write 'src/**/!(*.min).{js,jsx,css,json}' 'src/**/*.{js,jsx,css,json}' 'e2e_tests/**/*.{js,jsx,css,json}' 'scripts/**/*.{js,jsx,css,json}' 'build_scripts/**/*.{js,jsx,css,json}'",
```
nglgzz: Seems like in `src` you're formatting all not-minified files, and then all files, could it be that you meant to do the following?
```suggestion
    "format": "prettier-eslint --write 'src/**/!(*.min).{js,jsx,css,json}' 'e2e_tests/**/*.{js,jsx,css,json}' 'scripts/**/*.{js,jsx,css,json}' 'build_scripts/**/*.{js,jsx,css,json}'",
```

e2e_tests/integration/user.spec.js
```diff
@@ -18,10 +18,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* global Cypress, cy, test, expect, before */
```
jk05: so were `test, expect` being accidentally added as they were already in the eslintrc globals?

package.json
jk05: ^ good point

src/browser/AppInit.jsx
jk05: Im fine with this, I wonder what anyone else feels in terms of readability? Otherwise no problem from me

src/browser/styles/bootstrap.grid-only.min.css
```diff
@@ -7,8 +7,1100 @@
...
+ */ /*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
```
nglgzz: Yeah, this seems related to [this comment](https://github.com/neo4j/neo4j-browser/pull/1014/#discussion_r355902452). It probably wasn't meant to be formatted.

package.json
oskarhane: Thanks, fixed

src/browser/styles/bootstrap.grid-only.min.css
oskarhane: 👍

e2e_tests/integration/user.spec.js
oskarhane: yeah, I'm not sure why we needed to declare those before, probably due to an old ESLint version or something.

src/browser/AppInit.jsx
oskarhane: It's quite common so I don't think it will be a problem

jk05:
Soon as tests pass, I will approve 👍

PR #1017, opened 2019-12-10
User feedback running system commands

jk05:
Adjust the messaging shown to users with regards to system updates when running system commands

- Displays `1 system update,  no records` when `_systemUpdate` = 1
- Should display `n system update, no records`, when `_systemUpdate` > 1 (not been able to figure out to get this show in practice so suggestions welcome). I can remove this if its just deemed as unnecessary.
- Displays `no changes, no records` when !`_systemUpdate` (as before)

Ive only scoped the e2e tests within the `'serverVersion') >= 4.0`...

Screenshot 2019-12-10 at 11 34 06
Screenshot 2019-12-10 at 11 33 44

### DIFF
--- a/e2e_tests/integration/multi-db.spec.js
+++ b/e2e_tests/integration/multi-db.spec.js
@@ -38,6 +38,27 @@ describe('Multi database', () => {
     cy.connect('neo4j', password)
   })
   if (Cypress.config('serverVersion') >= 4.0) {
+    it('shows a message indicating whether system updates have occured', () => {
+      cy.executeCommand(':clear')
+
+      cy.executeCommand(':use system')
+      cy.executeCommand('CREATE DATABASE test1')
+
+      cy.wait(3000) // CREATE database can take a sec
+
+      cy.resultContains('1 system update, no records')
+
+      cy.executeCommand('STOP DATABASE test1')
+      cy.wait(1000)
+      cy.resultContains('1 system update, no records')
+
+      cy.executeCommand('STOP DATABASE test1')
+      cy.wait(1000)
+      cy.resultContains('no changes, no records')
+
+      cy.executeCommand('DROP DATABASE test1')
+      cy.executeCommand(':clear')
+    })
     it(':use command works + shows current db in editor gutter', () => {
       cy.executeCommand(':clear')
       const editor = () => cy.get('[data-testid="editor-wrapper"]')
--- a/src/browser/modules/Stream/CypherFrame/helpers.js
+++ b/src/browser/modules/Stream/CypherFrame/helpers.js
@@ -62,9 +62,7 @@ export function getBodyAndStatusBarMessages (result, maxRows) {
   let updateMessages = bolt.retrieveFormattedUpdateStatistics(result)
   let streamMessage =
     result.records.length > 0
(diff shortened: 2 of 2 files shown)


e2e_tests/integration/multi-db.spec.js
```diff
@@ -38,6 +38,27 @@ describe('Multi database', () => {
...
+      cy.executeCommand(':clear')
+
+      cy.executeCommand(':use system')
+      cy.executeCommand('CREATE DATABASE test1')
+
+      cy.wait(3000) // CREATE database can take a sec
+
+      cy.resultContains('1 system update, no records')
+
+      cy.executeCommand('STOP DATABASE test1')
```
jk05: I know this isn't very pretty - doing `STOP DATABASE` twice in order to get 0 system updates. So suggestions welcome. 

I also haven't figured out a way to get >1 system updates to test pluralisation...

src/browser/modules/Stream/CypherFrame/helpers.js
```diff
@@ -74,10 +72,15 @@ export function getBodyAndStatusBarMessages (result, maxRows) {
...
   }
 
+  const systemUpdatesValue = get(result, 'summary.counters._systemUpdates')
   const bodyMessage =
     (!updateMessages || updateMessages.length === 0) &&
     result.records.length === 0
-      ? '(no changes, no records)'
+      ? `(${(systemUpdatesValue > 0 &&
+          `${systemUpdatesValue} system update${(systemUpdatesValue > 1 &&
+            's') ||
```
jk05: can remove pluralisation if its not deemed necessary

src/browser/modules/Stream/CypherFrame/helpers.js
```diff
@@ -74,10 +72,15 @@ export function getBodyAndStatusBarMessages (result, maxRows) {
     streamMessage = streamMessage[0].toUpperCase() + streamMessage.slice(1)
   }
 
+  const systemUpdatesValue = get(result, 'summary.counters._systemUpdates')
```
jk05: I have had to pull it here similar to another section in the code, no ideal but only way I can see to get this value (suggestions welcome).

I haven't defaulted to anything as if it's undefined thats fine, and the logic should still hold...

in terms of naming, `systemUpdatesValue` or `systemUpdatesCount`?

e2e_tests/integration/multi-db.spec.js
jk05: Maybe I can add a unit test for the `/Stream/CypherFrame/helpers.js` instead for this. I will try this first and then return.

oskarhane: No, I think this is a good approach 👍

src/browser/modules/Stream/CypherFrame/helpers.js
oskarhane: I think this is the only place where it's available after serialization.

src/browser/modules/Stream/CypherFrame/helpers.js
oskarhane: 👍

PR #1020, opened 2019-12-16
Query plans view expanded by default

jk05:
WIP (looking to add a test)
- Have query plans expanded by default as opposed to collapsed now
- I chained the `setParentState` after the `extractPlan` in the `PlanView` `componentDidMount` in order to make sure `this.state.extractedPlan` is set prior to `this.toggleExpanded` is called (as it required it to be set first.). The only other way I could think of doing it was to have `this.ensureToggleExpand` always expand by default when no `_planExpand` is available on the parent component state initially but the pattern of setting parent state is preferred. Any suggestions here are welcome as I couldn't quite determine the best way to do this work.

Some example queries I tested and it seems to work ok:
```
PROFILE 
MATCH p=(n:Movie)--(m:Person)--(o:Movie)
WHERE m.born > 1000
AND length(p) > 1
MATCH (mm:Person) WHERE mm.name STARTS WITH 'T'
MATCH (mmm:Person) WHERE mmm.name STARTS WITH 'E'
RETURN p, mm, mmm
```
```
PROFILE
WITH ['Action','Drama','Mystery'] AS genreNames
UNWIND genreNames AS name
MATCH (g:Genre {name:name})
WITH g ORDER BY size( (g)<-[:BELONGS_TO]-() ) ASC
WITH collect(g) AS genres
WITH head(genres) AS first, tail(genres) AS rest
MATCH (first)<-[:BELONGS_TO]-(m:Title)-[:BELONGS_TO]->(other)
WITH m, collect(other) AS movieGenres, rest
WHERE all(g IN rest WHERE g IN movieGenres)
RETURN count(m)
```

Watch this [video](https://drive.google.com/file/d/1G1OhcDjDAsK6KV1cTE-J8UiGcwRyaQTs/view ) for a quick demo.

### DIFF
--- a/e2e_tests/integration/plan.spec.js
+++ b/e2e_tests/integration/plan.spec.js
@@ -37,6 +37,26 @@ describe('Plan output', () => {
     const password = Cypress.config('password')
     cy.connect('neo4j', password)
   })
+  it('displays the expanded details by default and displays/hides details when clicking the plan expand/collapse buttons respectively', () => {
+    cy.executeCommand(':clear')
+    cy.executeCommand(
+      'EXPLAIN MATCH (n:Person) WHERE n.age > 18 RETURN n.name ORDER BY n.age'
+    )
+
+    // initially should be expanded
+    const el1 = cy.get('[data-testid="planSvg"]', { timeout: 10000 })
+    el1.find('.detail').should('exist')
+
+    // collapse
+    cy.get('[data-testid="planCollapseButton"]', { timeout: 10000 }).click()
+    const el2 = cy.get('[data-testid="planSvg"]', { timeout: 10000 })
+    el2.find('.detail').should('not.exist')
+
+    // expand
+    cy.get('[data-testid="planExpandButton"]', { timeout: 10000 }).click()
+    const el3 = cy.get('[data-testid="planSvg"]', { timeout: 10000 })
+    el3.find('.detail').should('exist')
+  })
   if (Cypress.config('serverVersion') >= 3.5) {
     it('print Order in PROFILE', () => {
       cy.executeCommand(':clear')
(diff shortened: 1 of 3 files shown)


src/browser/modules/Stream/CypherFrame/PlanView.jsx
```diff
@@ -43,7 +43,9 @@ export class PlanView extends Component {
   }
 
   componentDidMount() {
-    this.extractPlan(this.props.result).catch(e => {})
+    this.extractPlan(this.props.result)
+      .then(() => this.props.setParentState({ _planExpand: 'EXPAND' }))
```
jk05: I explained my thought process for this in the PR description.

e2e_tests/integration/plan.spec.js
```diff
@@ -37,6 +37,26 @@ describe('Plan output', () => {
...
     cy.connect('neo4j', password)
   })
+  it('displays the expanded details by default and displays/hides details when clicking the plan expand/collapse buttons respectively', () => {
+    cy.executeCommand(':clear')
+    cy.executeCommand(
+      'EXPLAIN MATCH (n:Person) WHERE n.age > 18 RETURN n.name ORDER BY n.age'
+    )
+
+    // initially should be expanded
+    const el1 = cy.get('[data-testid="planSvg"]', { timeout: 10000 })
```
jk05: I could have set a `let el` and then reassign each time but I just followed the pattern elsewhere in the file for a new `const el`

e2e_tests/integration/plan.spec.js
```diff
@@ -37,6 +37,26 @@ describe('Plan output', () => {
...
+    el1.find('.detail').should('exist')
```
jk05: Im having to do a `.find('.detail')` as the `.detail` classes (e.g. `.detail identifiers` etc) were the only things I could identify in the dom as determining whether expanded or not. I don't really like this and see we don't really use class names in this manner elsewhere but I couldn't think of another way.

PR #1021, opened 2019-12-16
Add help cards for new administration features

nglgzz:
In this pull request I'm adding help cards related to databases and users. I will add the rest of the cards (roles and privileges) in a separate pull request in order to keep this one to a reasonable amount of changes. 

One thing to note is the change to the `ManualLink` component. I added a `minVersion` attribute to avoid getting 404s in case browser is connected to an older version of the server that doesn't contain these features.

### DIFF
--- a/src/browser/components/ManualLink.jsx
+++ b/src/browser/components/ManualLink.jsx
@@ -35,7 +35,10 @@ export function ManualLink({
   const cleanPage = page.replace(/^\//, '')
 
   let version = formatDocVersion(neo4jVersion)
-  if (minVersion && semver.cmp(neo4jVersion, '<', minVersion)) {
+  if (
+    minVersion &&
+    (!neo4jVersion || semver.cmp(neo4jVersion, '<', minVersion))
+  ) {
     version = formatDocVersion(minVersion)
   }
 
--- a/src/browser/components/ManualLink.jsx
+++ b/src/browser/components/ManualLink.jsx
@@ -20,13 +20,25 @@
 
 import React from 'react'
 import { connect } from 'react-redux'
+import semver from 'semver'
 
 import { getVersion } from 'shared/modules/dbMeta/dbMetaDuck'
 import { formatDocVersion } from 'browser/modules/Sidebar/Documents'
 
-export function ManualLink({ chapter, page, children, neo4jVersion }) {
+export function ManualLink({
+  chapter,
+  page,
+  children,
+  neo4jVersion,
+  minVersion
+}) {
   const cleanPage = page.replace(/^\//, '')
-  const version = formatDocVersion(neo4jVersion)
+
+  let version = formatDocVersion(neo4jVersion)
+  if (minVersion && semver.cmp(neo4jVersion, '<', minVersion)) {
+    version = formatDocVersion(minVersion)
+  }
+
   const url = `https://neo4j.com/docs/${chapter}/${version}/${cleanPage}`
 
   return (
--- a/src/browser/components/ManualLink.test.js
+++ b/src/browser/components/ManualLink.test.js
@@ -57,6 +57,10 @@ const tests = [
       minVersion: '3.5.0'
     },
     'https://neo4j.com/docs/driver-manual/4.0-preview/'
+  ],
+  [
+    { chapter: 'driver-manual', page: '/', minVersion: '3.5.0' },
+    'https://neo4j.com/docs/driver-manual/3.5/'
   ]
 ]
 
--- a/src/browser/components/ManualLink.test.js
+++ b/src/browser/components/ManualLink.test.js
@@ -39,6 +39,24 @@ const tests = [
   [
     { neo4jVersion: '4.0.0-beta03mr03', chapter: 'driver-manual', page: '' },
     'https://neo4j.com/docs/driver-manual/4.0-preview/'
+  ],
+  [
+    {
+      neo4jVersion: '3.4.11',
+      chapter: 'driver-manual',
+      page: '',
+      minVersion: '4.0.0'
+    },
+    'https://neo4j.com/docs/driver-manual/4.0/'
+  ],
+  [
+    {
+      neo4jVersion: '4.0.0-rc01',
+      chapter: 'driver-manual',
+      page: '',
+      minVersion: '3.5.0'
+    },
+    'https://neo4j.com/docs/driver-manual/4.0-preview/'
   ]
 ]
 
--- a/src/browser/documentation/help/create-database.jsx
+++ b/src/browser/documentation/help/create-database.jsx
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2002-2019 "Neo4j,"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react'
+import ManualLink from 'browser-components/ManualLink'
+import DbsOnSystemDb from './partials/dbs-on-systemdb'
+const title = 'CREATE DATABASE'
+const subtitle = 'Create a new database'
+const category = 'administration'
+const content = (
+  <>
+    <p>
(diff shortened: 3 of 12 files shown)


src/browser/documentation/help/create-database.jsx
```diff
@@ -0,0 +1,65 @@
...
+import React from 'react'
+import ManualLink from 'browser-components/ManualLink'
+import DbsOnSystemDb from './partials/dbs-on-systemdb'
+const title = 'CREATE DATABASE'
+const subtitle = 'Create a new database'
+const category = 'administration'
+const content = (
+  <>
+    <p>
+      Databases can be created using <code>CREATE DATABASE</code>
```
jk05: Looking at the pattern used elsewhere in this directory, the structure seems to be along the lines of 
`The <code>CREATE DATABASE</code> clause/command will...`

src/browser/components/ManualLink.jsx
```diff
@@ -20,13 +20,25 @@
 
 import React from 'react'
 import { connect } from 'react-redux'
+import semver from 'semver'
 
 import { getVersion } from 'shared/modules/dbMeta/dbMetaDuck'
 import { formatDocVersion } from 'browser/modules/Sidebar/Documents'
 
-export function ManualLink({ chapter, page, children, neo4jVersion }) {
+export function ManualLink({
```
jk05: this is more of a general question as I need to understand too - do you know why some files use arrow funcs and others function declaration syntax? I realise there are differences when it comes to this (like `this`, `arguments` and usage as a constructor) but beyond that Im not sure why we mix the two or is it not deemed to be an issue? @oskarhane maybe you can shed more light on this?

src/browser/documentation/help/show-users.jsx
```diff
@@ -0,0 +1,65 @@
...
+    <div className="links">
+      <div className="link">
+        <p className="title">Reference</p>
+        <p className="content">
+          <ManualLink
+            chapter="cypher-manual"
+            page="/administration/security/users-and-roles/#administration-security-users-show"
+            minVersion="4.0.0"
+          >
+            SHOW USER
```
jk05: `SHOW USERS`?

src/browser/documentation/help/create-database.jsx
nglgzz: You're right. I actually had a hard time figuring out what to write here and ended up copying what's in the docs. I'm still not sure if there's even the need for a description in some of these commands to be honest, as it's not giving any new information, the command itself is self explanatory.

src/browser/documentation/help/show-users.jsx
jk05: also I wonder if you can just refer to `title` here for all the files? or is that not always going to be case and/or relevant since  this isn't actually a title?

nglgzz: Good catch, thank you! 

The links sometimes point to a different page of the manual that is not necessarily the one for that command (could be a bigger section containing the docs for the command for example).

jk05: 👍 Yeah was just a thought. I guess it doesn't get used that much anyway so not a problem here. If it were something like Redux actions or something else that gets used a lot more it makes more sense as it reduces chance of typos etc

src/browser/documentation/help/drop-database.jsx
```diff
@@ -0,0 +1,65 @@
...
+import React from 'react'
+import ManualLink from 'browser-components/ManualLink'
+import DbsOnSystemDb from './partials/dbs-on-systemdb'
+const title = 'DROP DATABASE'
+const subtitle = 'Delete a database'
+const category = 'administration'
+const content = (
+  <>
+    <p>
+      Databases can be deleted using the command <code>DROP DATABASE.</code>
```
jk05: you may change the wording around but if not just be mindful of the full stop in `<code>DROP DATABASE.</code>`

src/browser/documentation/help/show-databases.jsx
```diff
@@ -0,0 +1,64 @@
...
+import ManualLink from 'browser-components/ManualLink'
+import DbsOnSystemDb from './partials/dbs-on-systemdb'
+const title = 'SHOW DATABASES'
+const subtitle = 'List all available databases'
+const category = 'administration'
+const content = (
+  <>
+    <p>
+      All available databases can be seen using the command{' '}
+      <code>SHOW DATABASES.</code>
```
jk05: full stop

src/browser/documentation/help/partials/admin-on-systemdb.jsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+import React from 'react'
+export default function AdminOnSystemDb() {
+  return (
+    <p>
+      <strong>A note on system database</strong>
+      <br />
+      Neo4j has a complex security model stored in the system graph, maintained
+      in a special database called the <code>system</code> database. All
+      administrative commands need to be executing against the{' '}
```
jk05: I feel like `executed` is better here?

nglgzz: Fixed.

src/browser/documentation/help/show-databases.jsx
nglgzz: Fixed.

src/browser/documentation/help/drop-database.jsx
nglgzz: Thank you for pointing it out! Fixed.

src/browser/documentation/help/create-database.jsx
nglgzz: I rephrased the descriptions, but I'm still not sure if all of them are needed as I mentioned above. Would love to hear @oskarhane's input on this.

src/browser/documentation/help/drop-database.jsx
```diff
@@ -0,0 +1,65 @@
...
+import React from 'react'
+import ManualLink from 'browser-components/ManualLink'
+import DbsOnSystemDb from './partials/dbs-on-systemdb'
+const title = 'DROP DATABASE'
+const subtitle = 'Delete a database'
+const category = 'administration'
+const content = (
+  <>
+    <p>
+      The <code>DROP DATABASE</code> command can be used to delete a database.
```
jk05: this is worded slightly different to the others, but I think is probably better worded in this way and in line with the rest of the help files.

Also, some places use `clause` and others `command` so maybe we need to be more exact with the wording with regards to each?

Sorry to be pedantic here; I feel like this is best answered by someone with better understanding of our wording/expectations from the help/docs.

nglgzz: This is the most helpful answer I found for the difference between `clause` and `command`. It's in the context of SQL, but I feel like it should have the same meaning in other query languages. 

https://stackoverflow.com/questions/52835714/what-is-difference-between-clause-command-statement-and-query-in-sql

jk05: Yeah I was looking at the same thing. I feel someone more knowledgeable with the docs and the terminology should be taking a look over this as we want to keep the wording accurate and consistent across all the Neo4j products.

src/browser/components/ManualLink.test.js
```diff
@@ -39,6 +39,24 @@ const tests = [
...
+  ],
+  [
+    {
+      neo4jVersion: '4.0.0-rc01',
+      chapter: 'driver-manual',
+      page: '',
+      minVersion: '3.5.0'
+    },
+    'https://neo4j.com/docs/driver-manual/4.0-preview/'
   ]
```
oskarhane: Could you add one more test, with `minVersion` defined but no `neo4jVersion` defined (simulating a browser that's not connected to a dbms yet).

src/browser/documentation/help/drop-database.jsx
oskarhane: I think it sounds like a good idea to bring in someone from docs to review the wording we use. Could you ping them and ask for some or their time @nglgzz ?

src/browser/components/ManualLink.jsx
oskarhane: Good observation @jk05. This is just due to personal preference and that we don't see it as an issue.
The main goal is for the code to be readable and easy to understand. I think in many cases there might just be code that's a bit older.
Feel free to modernize code you come across with your changes.
I see it as continuous maintenance (like moving away from deprecated lifecycle methods, porting classes to hooks etc) as I com across it.

src/browser/components/ManualLink.test.js
nglgzz: Good catch, thank you! I had to update the component as well to make this work.

src/browser/documentation/help/drop-database.jsx
davidoliverSP2: thanks @oskarhane, @jk05  & @nglgzz.

We're a little inconsistent with this, even if it's just about our definition and usage:

- **Clause** is very much used in the Cypher manual, as per https://neo4j.com/docs/cypher-manual/4.0/clauses/ 

- the newer multi-database administration section clearly prefers **command**: https://neo4j.com/docs/cypher-manual/4.0/administration/databases/#administration-databases-drop-database

I think this is a good candidate to discuss with the Terminology group, and publish our definition (both internally via a style & terminology guide, and externally as a description in the documentation)

PR #1025, opened 2019-12-20
Update help cards to point to the cypher manual

nglgzz:
The developer manual has been removed in version 3.5, and the cypher manual should be used going forward. I updated all the URLs to point to the correct version, so that we don't have any broken links (which is currently the case for 4.0). 

This change in the URL structure is backwards compatible. I tested all links with 4.0 and 3.4.

The `/schema` section has been moved to `/administration` in 4.0, and the last commit is to make sure the correct path is rendered depending on the version.

### DIFF
--- a/src/browser/components/ManualLink.jsx
+++ b/src/browser/components/ManualLink.jsx
@@ -25,6 +25,27 @@ import semver from 'semver'
 import { getVersion } from 'shared/modules/dbMeta/dbMetaDuck'
 import { formatDocVersion } from 'browser/modules/Sidebar/Documents'
 
+const movedPages = {
+  '/administration/indexes-for-search-performance/': {
+    oldPage: 'schema/index/',
+    oldContent: 'Schema indexes'
+  },
+  '/administration/constraints/': {
+    oldPage: 'schema/constraints/',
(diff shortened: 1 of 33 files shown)


src/browser/components/ManualLink.jsx
```diff
@@ -25,14 +25,40 @@ import semver from 'semver'
 import { getVersion } from 'shared/modules/dbMeta/dbMetaDuck'
 import { formatDocVersion } from 'browser/modules/Sidebar/Documents'
 
+const movedPages = {
```
oskarhane: 👍

src/browser/components/ManualLink.test.js
```diff
@@ -72,3 +72,40 @@ test.each(tests)('Render correct url for props %o', (props, expected) => {
...
+  (props, expected) => {
+    const { getByText } = render(
+      <ManualLink chapter="cypher-manual" {...props}>
+        link to manual
+      </ManualLink>
+    )
+    const url = getByText(expected.text).getAttribute('href')
+    expect(url).toEqual(expected.url)
+  }
+)
```
oskarhane: Great!

PR #1026, opened 2019-12-20
Show NaN in the table output instead of being handled displayed as null

jk05:
- Added `NaN` as a 'false positive' so we can display as it is and move on. Plus added an addition to a test.
- Currently, where we display the table output, we call `stringifyMod` with `stringModifier` as the second param (https://github.com/neo4j/neo4j-browser/blob/master/src/browser/modules/Stream/CypherFrame/TableView.jsx#L72). The issue is `stringModifier` doesn't handle `NaN` in any of the existing cases and so `undefined` is returned.
- This means we never return at this point (https://github.com/neo4j/neo4j-browser/blob/master/src/shared/services/utils.js#L371) and instead handled by this line (https://github.com/neo4j/neo4j-browser/blob/master/src/shared/services/utils.js#L375) and return 'null'

- This work is now in line with the output of cypher-shell (using queries shown on the Trello card)
Screenshot 2019-12-20 at 14 15 09
Screenshot 2019-12-20 at 14 16 27

### DIFF
--- a/src/shared/services/bolt/cypherTypesFormatting.js
+++ b/src/shared/services/bolt/cypherTypesFormatting.js
@@ -37,7 +37,7 @@ export const stringModifier = anything => {
 
 const numberFormat = anything => {
   // Exclude false positives and return early
-  if ([Infinity, -Infinity].includes(anything)) {
+  if ([Infinity, -Infinity, NaN].includes(anything)) {
     return `${anything}`
   }
   if (Math.floor(anything) === anything) {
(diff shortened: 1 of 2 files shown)


src/shared/services/bolt/cypherTypesFormatting.js
```diff
@@ -37,7 +37,7 @@ export const stringModifier = anything => {
 
 const numberFormat = anything => {
   // Exclude false positives and return early
-  if ([Infinity, -Infinity].includes(anything)) {
+  if ([Infinity, -Infinity, NaN].includes(anything)) {
     return `${anything}`
   }
   if (Math.floor(anything) === anything) {
```
jk05: this wont handle `NaN`. The main reason being any strict equality comparisons with `NaN` will always return `false` according to the JS Spec. Another option could be to add in a `isNan` here too but given that we are using the array above to exclude false positives (typeof number), then it seems like a logical place...

PR #1027, opened 2019-12-23
Add help cards for security commands

nglgzz:
Adds help cards for the following commands:

- ALTER USER
- SHOW ROLES
- [CREATE|GRANT|REVOKE|DROP] ROLE
- SHOW PRIVILEGES
- GRANT
- DENY
- REVOKE

Links to the help cards are also added into a new section of the `:help cypher` result card.

image

### DIFF
--- a/src/browser/documentation/help/alter-user.jsx
+++ b/src/browser/documentation/help/alter-user.jsx
@@ -21,13 +21,14 @@
 import React from 'react'
 import ManualLink from 'browser-components/ManualLink'
 import AdminOnSystemDb from './partials/admin-on-systemdb'
-const title = 'DROP USER'
-const subtitle = 'Delete a user'
+const title = 'ALTER USER'
+const subtitle = 'Modify a user'
 const category = 'administration'
 const content = (
   <>
     <p>
-      The command <code>DROP USER</code> can be used to delete an existing user.
+      The command <code>ALTER USER</code> can be used to modify an existing
+      user.
     </p>
     <div className="links">
       <div className="link">
@@ -35,10 +36,10 @@ const content = (
         <p className="content">
           <ManualLink
             chapter="cypher-manual"
-            page="/administration/security/users-and-roles/#administration-security-users-drop"
+            page="/administration/security/users-and-roles/#administration-security-users-alter"
             minVersion="4.0.0"
           >
-            DROP USER
+            ALTER USER
           </ManualLink>{' '}
           manual page
         </p>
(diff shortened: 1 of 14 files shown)


src/browser/documentation/help/alter-user.jsx
```diff
@@ -0,0 +1,87 @@
...
+import React from 'react'
+import ManualLink from 'browser-components/ManualLink'
+import AdminOnSystemDb from './partials/admin-on-systemdb'
+const title = 'ALTER USER'
+const subtitle = 'Modify a user'
+const category = 'administration'
+const content = (
+  <>
+    <p>
+      The command <code>ALTER USER</code> can be used to modify an existing
```
jk05: Put `command` after `<code>ALTER USER</code>` since it seems all the others created write it this way?

src/browser/documentation/help/deny.jsx
```diff
@@ -0,0 +1,88 @@
...
+            Subgraph security
+          </ManualLink>{' '}
+          manual page
+          <br />
+          <ManualLink
+            chapter="cypher-manual"
+            page="/administration/security/administration/#administration-security-administration-database-privileges"
+            minVersion="4.0.0"
+          >
+            Database administration
```
jk05: I assume there isnt a specific `DENY` section?

jk05: I see there are a few others like this too

src/browser/documentation/help/grant.jsx
```diff
@@ -0,0 +1,88 @@
...
+import ManualLink from 'browser-components/ManualLink'
+import AdminOnSystemDb from './partials/admin-on-systemdb'
+const title = 'GRANT'
+const subtitle = 'Grant privileges to roles'
+const category = 'security'
+const content = (
+  <>
+    <p>
+      The <code>GRANT</code> command allows an administrator to grant a
+      privilege to a role in order to access an entity.
```
jk05: entity seems vague. I assume its in the documentation like this too.

src/browser/documentation/help/deny.jsx
nglgzz: That's correct. `GRANT`, `DENY`, and `REVOKE` are together in two different sections. One for DB administration (limiting access to specific databases), and the other one for subgraph administration (limiting what kind of queries a role can do).

src/browser/documentation/help/grant.jsx
nglgzz: Yeah, this is coming from the documentation. From my understanding it's vague on purpose, because entity in this context includes many different things. It can be different types of operations on a database, a type of query, or even a type of query applied to a specific type of node/relationship. 

I do agree that is not so clear though. I think `access` is not really appropriate, as it's not just that, but other types of operations as well: traverse/read/write/match in the case of graphs, managing indexes&constraints, and other administration operations in the case of databases.

src/browser/documentation/help/alter-user.jsx
nglgzz: Yep, you're right. Fixed it!

src/browser/documentation/help/grant.jsx
jk05: Yeah it sounds vague but its not one I can really comment on much more. I guess if you have taken it from the documentation then at least its consistent.

PR #1028, opened 2019-12-23
e2e test added for export options

jk05:
- Add e2e tests for the correct export options to appear dependent on the Cypher frame view

- I question the need for alphabetical order mainly for the viz view, as it seems to be logical/better UX to me to have the PNG and SVG options first. It's a trivial change in any case and will be quick to implement once we decide on what to do (edit: after speaking with UX, it was deemed better to put in alphabetical order since in this case CSV/JSON will always be at the top and likely a better UX experience)

### DIFF
--- a/e2e_tests/integration/data-export.spec.js
+++ b/e2e_tests/integration/data-export.spec.js
@@ -31,69 +31,49 @@ describe('Data export', () => {
     const password = Cypress.config('password')
     cy.connect('neo4j', password)
   })
-  it('shows the correct export buttons', () => {
-    cy.executeCommand(':clear')
-    cy.executeCommand('CREATE (n:ExportTest) RETURN n')
-
-    cy.get('[data-testid="frame"]', { timeout: 10000 }).should('have.length', 1)
-
-    // viz
-    cy.get('[data-testid="frame-export-dropdown"]').trigger('mouseover')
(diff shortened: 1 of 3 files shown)


e2e_tests/integration/data-export.spec.js
```diff
@@ -31,17 +31,68 @@ describe('Data export', () => {
...
 
     cy.get('[data-testid="frame"]', { timeout: 10000 }).should('have.length', 1)
-    cy.get('[data-testid="frame-export-dropdown"]')
-      .first()
-      .trigger('mouseover')
-    cy.get('[data-testid="frame-export-dropdown"]')
-      .first()
-      .should('contain', 'Export PNG')
+
+    // viz
```
nglgzz: Since these blocks (viz, table, text, code) are testing exactly in the same way, you could [generate these tests dynamically](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Dynamically-Generate-Tests) making the file less verbose and making it obvious that they're doing the same thing.

oskarhane:
Btw, can you check for exports on the `Plan` view as well? 
So users can export the plan as a PNG.

jk05:
Ok Ill check `Plan` view. Are there any more to consider?

oskarhane:
> Ok Ill check `Plan` view. Are there any more to consider?

No, that should be the last view that has exports

Issue #905, opened 2019-02-26
After expand if we first dismiss a node then re-expand it throws error


ams1:
After expanding the child relationships of **the node** if we dismiss (remove from graph) any of the children, then double click on **the node** will reveal the dismissed child, BUT when double-clicking AGAIN on **the node** it will throw an error.

### Reproduce Steps

(given)
1. Display a single node that has child relationships (aka the node)
2. Expand the child relationships of the grandfather (to get the children)
3. Pick ANY node from the children (aka the child)
4. Dismiss (remove from graph) the child

(when)
5. Double click on the node -> it will reveal the dismissed child
6. Double click AGAIN on the node -> ERROR

### Expected Behavior

![image](https://user-images.githubusercontent.com/10154357/53415389-cfe3a580-39d9-11e9-94c4-8c948957be9c.png)

### Can you propose a solution?

As i see it, the expandedNodeMap contains duplicates after first removing the child and then expanding the node, possibly because the removeNode doesn't remove the node from the expandedNodeMap.
![image](https://user-images.githubusercontent.com/10154357/53415551-39fc4a80-39da-11e9-9533-23e4a350d677.png)

### Your Environment

| software name  | version
| ---------------|----------
|                |  
| OS             |  Windows
| Web browser    |  Chrome
| node.js        | v10.15.1
| npm            |6.4.1
| yarn           |1.13.0
| Neo4j Browser  | downloaded today 3.2.17?
| Neo4j          |

### Additional Information

**P.S.: BIG, BIG Thank you for Neo4j (!) and keep up the GREAT work you're doing!**

oskarhane:
Thanks for this report @ams1, we'll have a look at it!

PR #1031, opened 2019-12-31
Remove duplicates from the expandedNodeMap array

jk05:
fixes: #905 

- There seems to be an issue where the `expandedNodeMap` contains duplicates, so that as the code recursively loops over the `expandedNodeMap` array to remove nodes when a node is collapsed, it falls into an issue where the node is not found the second time round during the recursion and throws an error.

- I'm not sure of where to place a test for this as there didn't seem to be any e2e tests for expanding/closing nodes unless Im mistaken, so any input on that would be appreciated. The other option would be to open a unit test file but I don't want to start cluttering the code base with unnecessary unit test files if they're not needed.

**Before** - shows the duplicates in the `expandedNodeMap`:
Screenshot 2020-01-02 at 12 37 31

[Before Video](https://drive.google.com/file/d/1veHbQHBpT4GDvxop8McrQC5lk06pd42J/view)

**After** - shows no duplicates in the `expandedNodeMap`:

Screenshot 2020-01-02 at 12 38 29

[After video](https://drive.google.com/file/d/1HYt4MnB8I2bDZ5Vfia11tdZ5c9x_Uj6G/view)

### DIFF
--- a/src/browser/modules/D3Visualization/lib/visualization/components/graph.js
+++ b/src/browser/modules/D3Visualization/lib/visualization/components/graph.js
@@ -82,7 +82,7 @@ export default class Graph {
         this.nodeMap[eNode.id] = eNode
         this._nodes.push(eNode)
         this.expandedNodeMap[node.id] = this.expandedNodeMap[node.id]
-          ? this.expandedNodeMap[node.id].concat([eNode.id])
+          ? [...new Set(this.expandedNodeMap[node.id].concat([eNode.id]))]
           : [eNode.id]
       }
     }


src/browser/modules/D3Visualization/lib/visualization/components/graph.js
```diff
@@ -82,7 +82,7 @@ export default class Graph {
         this.nodeMap[eNode.id] = eNode
         this._nodes.push(eNode)
         this.expandedNodeMap[node.id] = this.expandedNodeMap[node.id]
-          ? this.expandedNodeMap[node.id].concat([eNode.id])
+          ? [...new Set(this.expandedNodeMap[node.id].concat([eNode.id]))]
```
jk05: so I have to investigate a little more but I have removed duplicates at this stage and it seems to work. I have to look into the `nodes` parameter a bit as to why it includes the multiples of the node relationships. It may be that thats where the change is required.

jk05: I have tracked back to https://github.com/neo4j/neo4j-browser/blob/master/src/shared/services/bolt/boltMappings.js#L214 and tried to put in a `reduce` here to remove duplicates but since this function is used in several places, it seems to cause other issues. So to narrow the scope to the particular problem, Im going to stick with my initial solution for now.

oskarhane: That makes sense and I agree with keeping the scope small. 

I think Hugo is tackling it at a lower level in https://github.com/neo4j/neo4j-browser/pull/1003.
So maybe his PR fixes this bug as well @jk05.
Could you have a look if that's the case?

jk05: Ill take a look 👍

jk05: I pulled it down locally and the bug still seems to be there...

jk05: so we can either can look to go ahead with this work to keep the scope smaller, or I can leave a comment on #1003 with a link to this PR (as the context of where Hugo is making the changes could potentially be a better fit than this work). What do you think @oskarhane?

oskarhane: I think this change makes sense 👍

PR #1048, opened 2020-01-17
Use best practices driver transaction functions by default

oskarhane:
To get automatic retries in cases of leader switches in clusters we want to move away from auto-committed transactions `session.run()` to transaction functions `session.xxxTransaction(tx => tx.run())`.

However, certain Cypher commands (i.e. `USING PERIODIC COMMIT`) requires an auto-committed transaction to be able function.
That's why we introduce the opt-out of transaction functions command and instead use auto-commited transactions: `:auto`.

Just prepend it to the query:  `:auto RETURN 1` and it'll use an auto-committing transaction.

### DIFF
--- a/src/shared/services/bolt/boltHelpers.js
+++ b/src/shared/services/bolt/boltHelpers.js
@@ -46,7 +46,7 @@ export const buildTxFunctionByMode = session => {
   if (!session) {
     return null
   }
-  return session && session._mode !== 'READ'
+  return session._mode !== 'READ'
     ? session.writeTransaction.bind(session)
     : session.readTransaction.bind(session)
 }
--- a/src/shared/services/bolt/boltHelpers.js
+++ b/src/shared/services/bolt/boltHelpers.js
@@ -41,3 +41,12 @@ export const isConfigValTruthy = val =>
   [true, 'true', 'yes', 1, '1'].indexOf(val) > -1
 export const isConfigValFalsy = val =>
   [false, 'false', 'no', 0, '0'].indexOf(val) > -1
+
+export const buildTxFunctionByMode = session => {
+  if (!session) {
+    return null
+  }
+  return session && session._mode !== 'READ'
+    ? session.writeTransaction.bind(session)
+    : session.readTransaction.bind(session)
+}
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -57,7 +57,8 @@ import {
   unsuccessfulCypher,
   SINGLE_COMMAND_QUEUED,
   listDbsCommand,
-  useDbCommand
+  useDbCommand,
+  implicitTxCommand
 } from 'shared/modules/commands/commandsDuck'
 import {
   getParamName,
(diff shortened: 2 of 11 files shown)


src/shared/services/commandInterpreterHelper.js
```diff
@@ -337,9 +341,25 @@ const availableCommands = [
...
+      // and attach to the actions object
+      const query = action.cmd.replace(
+        getCmdChar(state) + autoCommitTxCommand,
+        ''
+      )
+      action.query = query.trim()
+
+      // We need to find out if this is an impllicit tx or not
+      // i.e. Did we strip something off above here?
+      const autoCommit = action.query.length < action.cmd.trim().length
```
HerrEmil: You could get this value from the string.replace above
```
let autoCommit = false
const query = action.cmd.replace(
  getCmdChar(state) + autoCommitTxCommand,
  () => autoCommit = true && ''
)
```
But I'm not sure if it's more or less readable

src/shared/services/bolt/boltHelpers.js
```diff
@@ -41,3 +41,12 @@ export const isConfigValTruthy = val =>
   [true, 'true', 'yes', 1, '1'].indexOf(val) > -1
 export const isConfigValFalsy = val =>
   [false, 'false', 'no', 0, '0'].indexOf(val) > -1
+
+export const buildTxFunctionByMode = session => {
+  if (!session) {
+    return null
+  }
+  return session && session._mode !== 'READ'
```
HerrEmil: session is always truthy on this line

src/shared/services/commandInterpreterHelper.js
```diff
@@ -337,9 +341,25 @@ const availableCommands = [
...
+
+      // Since we now also handle queries with the :auto prefix, we need to strip that
+      // and attach to the actions object
+      const query = action.cmd.replace(
+        getCmdChar(state) + autoCommitTxCommand,
+        ''
+      )
+      action.query = query.trim()
+
+      // We need to find out if this is an impllicit tx or not
```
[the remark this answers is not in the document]
oskarhane: Thanks!

src/shared/services/bolt/boltHelpers.js
oskarhane: Good point 👍

src/shared/services/commandInterpreterHelper.js
oskarhane: Yeah, I do think it's more readable the way it is tbh. Interesting idea though, I've never thought of that.

KaysHaydock:
Hello, can I just check please that in v1.2.4 (Desktop) version for a MAC, I can still use USING PERIODIC COMMIT but now I just have to add :auto USING PERIODIC COMMIT in front of the query please?

oskarhane:
> Hello, can I just check please that in v1.2.4 (Desktop) version for a MAC, I can still use `USING PERIODIC COMMIT` but now I just have to add `:auto USING PERIODIC COMMIT` in front of the query please?

@KaysHaydock Correct, prepend the query with `:auto` and everything should work as before.

PR #1051, opened 2020-01-23
Setup gradual migration to Typescript

nglgzz:
These changes update the webpack, eslint, prettier, and jest configurations to be able to build, lint, format, and test Typescript alongside Javascript. 

This way we can keep working on the code-base as usual, while being able to migrate parts of it to Typescript when possible. 

I also migrated one module and its tests as an example.

For the most part I followed these docs for updating the configs: https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html


package.json
```diff
@@ -65,7 +65,11 @@
     "@hot-loader/react-dom": "^16.8.6",
     "@testing-library/react": "^9.0.1",
     "@testing-library/react-hooks": "^2.0.1",
+    "@types/jest": "^24.9.0",
```
HerrEmil: Would make sense to add type definitions for other libraries as well, e.g. `@types/react-dom`

nglgzz: Yeah makes sense. Should add them here or should I leave it to whenever they will be needed, eg. switching the first React component to TS?

oskarhane: I think we should wait until we need them, but very good point @HerrEmil

oskarhane:
Sorry for holding back on this one. 
After next browser release, I think we should merge this.

nglgzz:
I rebased on the latest master to get rid of the `yarn.lock` and `package.json` conflicts.

src/shared/utils/number-to-US-locale.test.ts
```diff
@@ -19,75 +19,75 @@
  */
 
 /* global describe, test, expect */
-import numberToUSLocale from './number-to-US-locale'
+import numberToUSLocale from './number-to-US-locale';
```
OskarDamkjaer, line 22: I think it'd be nice to leave of the semi colons for typescript as well to avoid these kinds of diffs every time we port something to ts. Thoughts?

HerrEmil: I agree, semicolons should match between js and ts

.prettierrc
```diff
@@ -7,5 +7,13 @@
   "trailingComma": "none",
   "jsxBracketSameLine": false,
   "arrowParens": "avoid",
-  "jsxSingleQuote": false
+  "jsxSingleQuote": false,
+  "overrides": [
+    {
+      "files": ["*.ts", "**/*.ts"],
+      "options": {
+        "semi": true
```
OskarDamkjaer, line 15: I'm guessing this is what adds the semicolons?

HerrEmil: Yes, and in eslint, removing now

PR #1058, opened 2020-01-30
Make node 12 the default (using .nvmrc file) build env

oskarhane:
Fix breaking build under node 12 (firebase was old).
Also update the other dependencies with minor+patch versions.


oskarhane:
We need to make sure builder servers are using `.npmrc` files before merging this.

oskarhane:
> If any script is using engines.node value in package.json, maybe that should also be updated?

Good point, I'll fix

PR #1060, opened 2020-02-05
filter example scripts by database version

### LINKED ISSUE #1033: What is related, and how -> ERROR There is no procedure with the name `db.schema` registered...
The `Favorites/Sample Scripts/Data Profiling/What is related, and how` query is not working for  neo4j server version 4.0.0-rc01.

Not sure if `db.schema()` is permanently moved to `db.schema.visualization()` on the server side.

https://github.com/neo4j/neo4j-browser/blob/38d7a31c89bae5fcc6f5ed7f39edd7b06c8fa590/src/shared/modules/favorites/staticScripts.js#L58

HerrEmil:
Fixes #1033  by adding a version range on the example queries and filtering the list by if the currently connected database is in the range.

Also added a Connect to database example for when there is no database version (when disconnected).

### DIFF
--- a/src/shared/modules/favorites/staticScripts.js
+++ b/src/shared/modules/favorites/staticScripts.js
@@ -21,89 +21,123 @@
 export const scripts = [
   {
     folder: 'basics',
-    content:
-      '// Hello World!\nCREATE (database:Database {name:"Neo4j"})-[r:SAYS]->(message:Message {name:"Hello World!"}) RETURN database, message, r'
+    content: '// Connect to DBMS\n:server connect',
+    versionRange: '0.0.0'
   },
   {
     folder: 'basics',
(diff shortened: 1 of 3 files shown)


src/shared/modules/favorites/staticScripts.js
```diff
@@ -19,113 +19,151 @@
  */
 
 export const scripts = [
+  {
+    folder: 'basics',
+    content: '// Connect to database\n:server connect',
```
oskarhane: 👍 
I think this should be "Connect to DBMS". We use "database" here and there but we should move away from that.

HerrEmil: OK, I will change the wording and add some tests :)

PR #1061, opened 2020-02-06
Fix re-run for Cypher queries to target original DB

oskarhane:
## Changes made
Wrongfully we assumed that the target DB always should be the globally selected one. This was not a valid assumption and this PR fixes that.
Now a command action object can target a specific DB.

I refactored the arguments to our `readTransaction`, `writeTransaction`, and `directTransaction` to take an options object instead of just adding more args.

Unit tests for when using web workers and not + e2e tests included.

## Reproduce the bug

1. Run cypher query (q1) on db1 `RETURN rand()`
1. Switch to a different db, db2: `:use system`
1. Click "Re-run" on result frame for q1 from step 1.
1. See it being executed on db2 rather than db1, which is unexpected behaviour.

### DIFF
--- a/src/shared/services/bolt/bolt.js
+++ b/src/shared/services/bolt/bolt.js
@@ -27,13 +27,11 @@ import { generateBoltHost } from 'services/utils'
 import {
   runCypherMessage,
   cancelTransactionMessage,
-  closeConnectionMessage,
-  CYPHER_ERROR_MESSAGE,
-  CYPHER_RESPONSE_MESSAGE,
-  POST_CANCEL_TRANSACTION_MESSAGE,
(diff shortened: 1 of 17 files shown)


src/shared/services/bolt/bolt.js
```diff
@@ -96,22 +95,26 @@ function routedWriteTransaction(input, parameters, requestMetaData = {}) {
...
+      id,
+      workFn,
+      onLostConnection
+    )
     return [id, workerPromise]
   } else {
-    return boltConnection.routedWriteTransaction(
-      input,
-      parameters,
+    console.log('dfdfdf')
```
HerrEmil: Leftover console log

PR #1067, opened 2020-03-10
Display warning in UI when database is unavailable

HerrEmil:
Display an error banner when current database is not online
error banner
I reused an existing error banner in the main component that was unused.

Also, if user tries to `:use` an unavailable database, display an error fram in same way as when trying to use a non-existing database
error frame

If the current database is not responding, queries time out after 30 seconds. This makes the dbMetaEpic slow as it has four queries running in sequence. I remove the chained mergeMaps and wrapped the queries in a forkJoin instead to run the queries in parallel.

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaDuck.js
+++ b/src/shared/modules/dbMeta/dbMetaDuck.js
@@ -312,162 +312,166 @@ export const dbMetaEpic = (some$, store) =>
           .throttle(() => some$.ofType(DB_META_DONE))
           // Server version and edition
           .do(store.dispatch({ type: FETCH_SERVER_INFO }))
-          // Labels, types and propertyKeys, and server version
-          .mergeMap(() => {
-            const state = store.getState()
-            const db = getUseDb(state)
-            if (db === 'system') {
-              store.dispatch(updateMeta([]))
-              return Rx.Observable.of(null)
-            }
-
-            return Rx.Observable.fromPromise(
-              bolt.routedReadTransaction(
-                metaQuery,
-                {},
-                {
-                  useCypherThread: shouldUseCypherThread(store.getState()),
-                  onLostConnection: onLostConnection(store.dispatch),
-                  ...getBackgroundTxMetadata({
-                    hasServerSupport: canSendTxMetadata(store.getState())
-                  })
-                }
-              )
-            ).catch(e => {
-              store.dispatch(updateMeta([]))
-              return Rx.Observable.of(null)
-            })
-          })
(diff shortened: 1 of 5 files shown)


src/shared/services/commandInterpreterHelper.js
```diff
@@ -209,20 +209,31 @@ const availableCommands = [
...
+        if (!dbMeta) {
+          throw new UseDbError({
+            code: 'NotFound',
+            message: `A database with the "${dbName}" name could not be found.`
+          })
+        }
+        if (dbMeta.status !== 'online') {
+          throw new UseDbError({
+            code: 'DatabaseUnavailable',
+            message: `Database "${dbName}" is unavailable.`
```
oskarhane: Could you add the actual status to this message as well? 
Something like `Database "${dbName}" is unavailable, it's status is "${dbMeta.status}"`

src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -312,162 +312,166 @@ export const dbMetaEpic = (some$, store) =>
...
-                  } else if (name === 'dbms.security.auth_enabled') {
-                    let authEnabled = true
-                    if (
-                      typeof value !== 'undefined' &&
-                      isConfigValFalsy(value)
-                    ) {
-                      authEnabled = false
-                    }
-                    store.dispatch(setAuthEnabled(authEnabled))
+            Rx.Observable.forkJoin([
```
oskarhane: Not your fault, but these observables are crazy hard to read. Do you think it can be refactored to smaller pieces and glued together here?
What's super important is that this doesn't stop listening for new events after errors has happened.

src/shared/services/commandInterpreterHelper.js
HerrEmil: Done

src/shared/modules/dbMeta/dbMetaDuck.js
HerrEmil: I agree, I thought a bit about how to break apart the chain and the options I could think off are either to refactor the chain to return actions instead of dispatching them, so parts can be broken out into pure functions, or to replace the large chain with several small chains, each with their own actions. I didn't do it in this branch as either of those options would be a bigger change and hard to review mixed with these changes. Maybe we should consider the big size of the epic tech debt and add a card in our backlog to break it apart?

HerrEmil: Added tech debt card to backlog

PR #1078, opened 2020-03-30
Update E2E tests for Neo4j 4.1

HerrEmil:


e2e_tests/integration/bolt.spec.js
```diff
@@ -85,7 +85,12 @@ describe('Bolt connections', () => {
         'contain',
         'noroles'
       )
-      cy.get('[data-testid="user-details-roles"]').should('contain', '-')
+      if (Cypress.config('serverVersion') === 4.0) {
```
oskarhane: I think should be on all versions except 4.1

e2e_tests/integration/bolt.spec.js
```diff
@@ -85,7 +85,12 @@ describe('Bolt connections', () => {
...
+        cy.get('[data-testid="user-details-roles"]').should('contain', '-')
+      }
+      if (Cypress.config('serverVersion') === 4.1) {
+        cy.get('[data-testid="user-details-roles"]').should('contain', 'PUBLIC')
```
oskarhane: This is a strange value for us to get back. 
Let's look into what we should show the user at some later point.

e2e_tests/integration/schema-frame.spec.js
```diff
@@ -95,10 +95,19 @@ describe('Schema Frame', () => {
       cy.executeCommand(':clear')
       cy.executeCommand(':schema')
 
-      cy.get('[data-testid="frameContents"]')
-        .should('contain', 'Constraints')
-        .and('contain', 'schematest.prop1')
-        .and('contain', 'IS UNIQUE')
+      if (Cypress.config('serverVersion') === 4.0) {
```
oskarhane: Same here I think, should be `<= 4.0`.

PR #1080, opened 2020-04-05
GraSS Style edit

jk05:
Lab Day
 - https://trello.com/c/t6S02ZIt/49-grass-editor
- currently added as an experimental feature. I have placed it in the side bar and as an edit function within the style frame itself just for comparison as people may have differing opinions on the UI aspects. Either one can then be removed from the code depending on if we prefer to go with the other.
- Mostly working ok. No tests added and no additional UI things added due to time constraints that Lia suggested (these are described on the Trello card above and can be added when needed)
- There is a slight bug when changing css on the style frame it wont show as updated on the fly on the sidebar. Its probably some issue I didnt bother to look into as its not really important - we wouldn't have both the editing on the sidebar and the frame at the same time anyway.
- Code could do with clean up and better naming

Video: https://drive.google.com/file/d/1nCu80vUAvysdf183aadZD8ClKLsEpYwI/view

Screenshot 2020-04-30 at 19 16 17
Screenshot 2020-04-30 at 19 16 10


oskarhane:
I'm closing this for now @jk05. Not that we don't want it, but it will require some work + rebasing for it to be mergeable.
Please keep your private branch for you or someone else to pickup in the future.

PR #1085, opened 2020-04-16
Aura E2E testing

HerrEmil:
Add `isAura` utility and use it to skip Aura incompatible parts of E2E tests.


HerrEmil:
Initially I tried setting up an Aura test build without touching browser code. The easiest way was to skip the tests that failed and cypress cli has an issue with the ignoreTestFiles flag that was not an issue when you set it in config. It is a heavy-handed test suite setup to skip entire files when some part of some test is incompatible with Aura, I agree that it is nicer to skip individual tests with an env var instead, I'll get started updating the tests.

oskarhane:
I've added a few changes now where we treat 'aura' as it's own edition and now we connect to aura using the `neo4j://` scheme so we can do the writing tests as well.

e2e_tests/support/defaults.js
```diff
@@ -2,24 +2,43 @@
...
-      ? Cypress.env('bolt-url')
-      : Cypress.config('boltHost') + ':' + Cypress.config('boltPort'))
+  Cypress.env('bolt-url')
+    ? Cypress.env('bolt-url')
+    : 'bolt://' + Cypress.config('boltHost') + ':' + Cypress.config('boltPort')
 )
+
+function removeUriScheme(uri = '') {
+  const SEPARATOR = '://'
+  const [_skip, ...rest] = uri.includes(SEPARATOR)
```
HerrEmil, line 40: You could skip the rest and join operation `return uri.includes(SEPARATOR) ? uri.split(SEPARATOR)[1] : uri`

oskarhane: Not safe, there can be multiple `://` in a url. ex: `http://test.com/?url=http://hello.com`

PR #1086, opened 2020-04-16
Add initalSwitchConnectionFailEpic to rootEpic

HerrEmil:
In the [autorun play commands PR](https://github.com/neo4j/neo4j-browser/pull/1072) I added the initialSwitchConnectionFailEpic epic, but forgot to combine it in the root epic so it was never active. This created the bug that the server switch fail frame was not displayed when starting browser from neo4j desktop without a started database. This PR adds the epic to rootEpic to fix this bug and adds an E2E test to verify that the server switch fail frame is displayed as intended.

### DIFF
--- a/e2e_tests/integration/desktop-env.spec.js
+++ b/e2e_tests/integration/desktop-env.spec.js
@@ -61,4 +61,26 @@ describe('Neo4j Desktop environment', () => {
       .first()
       .should('contain', 'Connection updated')
   })
+
+  it('displays disconnected banner and connection failed frame when initial state is INACTIVE', () => {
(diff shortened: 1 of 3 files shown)


e2e_tests/integration/desktop-env.spec.js
```diff
@@ -61,4 +61,26 @@ describe('Neo4j Desktop environment', () => {
       .first()
       .should('contain', 'Connection updated')
   })
+
+  it('displays disconnected banner and connection failed frame when initial state is INACTIVE', () => {
```
oskarhane, line 65: 👍

PR #1090, opened 2020-04-29
Feature relatable

huboneo:
This PR introduces a new configuration option in browser for truncating max field length before they are passed to rendering layers. The reason for this is that certain queries can produce a single result with a lot of values in a given field, which we had yet to account for.

This PR also adds `@relate-by-ui/relatable` table to improve rendering performance.

Finally this PR also adds a max limit before Viz is disabled as it cannot handle more than 1000 rendered nodes/relationships

### STR
- see trello card

### Screenshot
Screenshot 2020-04-29 at 11 36 52
Screenshot 2020-04-29 at 11 46 03
Screenshot 2020-04-29 at 11 38 46
Screenshot 2020-04-29 at 11 39 07


huboneo:
@oskarhane I have passing tests now apart from 4.1

huboneo:
@oskarhane great catch, updated!

PR #1093, opened 2020-05-04
Use public procedure to read server conf, if it exists

oskarhane:


oskarhane:
Blocked until server side changes go in

PR #1096, opened 2020-05-04
More reliable connection status check

oskarhane:
Main driver for this PR is the existing bug that can be reproduced like this:

1. Open Desktop (don't start a dbms)
1. Open Neo4j Browser and see the `:server switch fail` frame.
1. Type `:server connect` or `:server status` in Neo4j Browser. It say's you're connected.

The fix: Use the `isConnected` that we have in state and don't infer it from other state props.

New look of unconnected `:server status`:  
Screenshot 2020-05-04 15 32 54

This PR also does some user capabilities checks (can user read server config + can user list server procedures).
I've also added a `:debug` command that might be helpful when coding.
This command will get more useful in the future, but this is the introduction to it. At this point only targeted at Noe4j Browser developers.

Also, target system db (if we have multi-db support) for the "read server config" and "get existing procedures".

### DIFF
--- a/e2e_tests/integration/commands.spec.js
+++ b/e2e_tests/integration/commands.spec.js
@@ -38,6 +38,7 @@ const commands = [
   ':param x => 1',
   ':param y: 2',
   ':queries',
+  ':debug',
   ':get /',
   ':unknown',
   'RETURN 1',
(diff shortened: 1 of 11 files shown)


oskarhane:
Blocking due to lack of tests. Will add soon.

oskarhane:
Tests are now added, please review @HerrEmil when you have time.
The scenario fixed here is that when browser has connection creds stored and launches (without connecting because of server not reachable for example), `:server status` and `:server connect` would think it's connected.

PR #1103, opened 2020-05-22
Fix auth scheme error from driver when auth is disabled + reflect auth status on `:server connect` frame`

oskarhane:
+ also fix issue where usernames didn't clear when server switch auth mode during browser session (without reload)

### DIFF
--- a/src/browser/modules/Stream/Auth/ConnectionForm.jsx
+++ b/src/browser/modules/Stream/Auth/ConnectionForm.jsx
@@ -49,6 +49,7 @@ import {
   toggleSchemeRouting,
   isNonSupportedRoutingSchemeError
 } from 'services/boltscheme.utils'
+import { StyledConnectionBody } from './styled'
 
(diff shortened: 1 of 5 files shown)


src/browser/modules/Stream/Auth/ConnectionForm.jsx
```diff
@@ -260,15 +261,29 @@ export class ConnectionForm extends Component {
...
-          username={this.state.username}
+          username={this.props.activeConnectionData.username}
           storeCredentials={this.props.storeCredentials}
           hideStoreCredentials={this.state.authenticationMethod === NO_AUTH}
         />
       )
+    } else if (
+      this.props.isConnected &&
+      this.props.activeConnectionData &&
+      this.props.activeConnectionData.authEnabled === false // excplicit false = auth enabled for sure
```
HerrEmil, line 280: Comment should probably say that auth is disabled for sure

PR #1105, opened 2020-05-25
Mdx guide support

HerrEmil:
Add a check on play commands to see if a guide file is an mdx file, and if it is, parse it as mdx. The mdx file is split into slides on lines with `---`, just like [mdx-deck](https://github.com/jxnblk/mdx-deck).


package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://neo4j.com/developer/guide-neo4j-browser/",
   "main": "dist/index.html",
   "scripts": {
-    "build": "NODE_ENV=\"production\" webpack --config ./build_scripts/webpack.config.js",
+    "build": "cross-env NODE_ENV=\"production\" webpack --config ./build_scripts/webpack.config.js",
```
oskarhane, line 23: 👍

src/shared/services/commandInterpreterHelper.js
```diff
@@ -490,13 +490,19 @@ const availableCommands = [
       }
 
       const url = action.cmd.substr(cmdchar.length + 'play '.length)
+      const anchor = document.createElement('a')
+      anchor.href = url
+      const filenameExtension = anchor.pathname.includes('.')
+        ? anchor.pathname.split('.').pop()
+        : 'html'
```
oskarhane, line 497: This feel like a hack. Can we do something for formal, like using https://developer.mozilla.org/en-US/docs/Web/API/URL or something?

src/browser/modules/Docs/MdxSlide.jsx
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+import MDX from '@mdx-js/runtime'
+import React from 'react'
+import Slide from '../Carousel/Slide'
+import ManualLink from 'browser-components/ManualLink'
+
+export const splitMdxSlides = (mdx = '') => mdx.split(/\r?\n---\r?\n/)
+
+class MdxErrorBoundary extends React.Component {
```
oskarhane, line 28: Great!

src/shared/services/commandInterpreterHelper.js
HerrEmil: Should I create the URL object with `new URL(url)` instead of `document.createElement('a')`? I agree that it is easier to read, I only used the createElement function because it works in IE.

oskarhane: I see. 

The `URL` should be polyfilled so it works in IE as well since we declare that we need support for it in https://github.com/neo4j/neo4j-browser/blob/master/babel.config.js#L19

Good call though, let's change and try it out in IE before next release.

src/browser/modules/Stream/PlayFrame.jsx
```diff
@@ -152,6 +151,22 @@ function generateContent(stackFrame, bus, onSlide, shouldUseSlidePointer) {
 
   // Found a remote guide
   if (stackFrame.result) {
+    if (['md', 'mdx'].includes(stackFrame.filenameExtension)) {
+      return {
+        guide: (
+          <Docs
+            originFrameId={stackFrame.id}
```
oskarhane, line 158: You need to add `lastUpdate={stackFrame.ts}` here for the reload to work

HerrEmil: Thank you :)

PR #1119, opened 2020-06-11
Show graph items properties in table view, not the metadata (id, labels, types etc)

huboneo:
This PR makes showing raw results in tables an opt-in setting, restoring previous display behaviour.

### Screenshots
disabled
Screenshot 2020-06-11 at 11 33 55

enabled
Screenshot 2020-06-11 at 11 33 48

Paths
Screenshot 2020-06-11 at 12 55 05


oskarhane:
We're actually going the other route and will start displaying all metadata info in the text table instead.
Sorry!

PR #1124, opened 2020-06-23
Enable sending metrics when in a Desktop env

oskarhane:


src/shared/modules/udc/udcDuck.js
```diff
@@ -50,13 +50,34 @@ export const CLEAR_EVENTS = `${NAME}/CLEAR_EVENTS`
...
 const typeToEventName = {
   [CYPHER]: 'cypher_attempts',
   [CYPHER_SUCCEEDED]: 'cypher_wins',
-  [CYPHER_FAILED]: 'cypher_fails'
+  [CYPHER_FAILED]: 'cypher_fails',
+  [EVENT_CLIENT_START]: 'client_starts'
+}
+
+export const typeToMetricsObject = {
+  [CYPHER]: { category: 'cypher', label: 'executed' },
```
akollegger, line 71: perhaps `{... label: 'sent'}` since this event is the user action which results in either 'succeeded' or 'failed'

src/shared/modules/udc/udcDuck.js
```diff
@@ -50,13 +50,34 @@ export const CLEAR_EVENTS = `${NAME}/CLEAR_EVENTS`
...
+  [CYPHER_SUCCEEDED]: { category: 'cypher', label: 'succeeded' },
+  [CYPHER_FAILED]: { category: 'cypher', label: 'failed' },
+  [EVENT_CLIENT_START]: { category: 'client', label: 'started' },
```
akollegger, line 74: Could we use `{ category: 'driver' ...}` to be more specific? "client" could be the client application, or a library, or whatever

PR #1129, opened 2020-07-01
Fix Edge support

HerrEmil:
The MDX Guide PR #1105 included non-es5 dependencies which broke the application in Internet Explorer and pre-chromium Edge. This PR passes two dependencies to babel-loader to be transpiled which fixes the app in pre-chromium Edge.

IE11 support was planned to be dropped and is not fixed, but formalized by updating the browserslist target and removing a little bit of IE specific code.


build_scripts/webpack-rules.js
```diff
@@ -25,9 +25,10 @@ module.exports = [
     test: /\.(js|jsx)$/,
     include: [
       path.resolve('src'),
-      path.resolve('node_modules/@neo4j/browser-lambda-parser')
+      path.resolve('node_modules/@literal-jsx/parser'),
+      path.resolve('node_modules/@neo4j/browser-lambda-parser'),
+      path.resolve('node_modules/remark-mdx')
     ],
-    // exclude: /(node_modules(?!:\/@neo4j))|(cypher-codemirror)|(test_utils)|(dist)/,
```
oskarhane, line 30: I take it as this was redundant?

babel.config.js
```diff
@@ -16,7 +16,7 @@ const toExport = {
         corejs: 3,
         modules: isTest ? 'commonjs' : false,
         targets: {
-          browsers: ['last 1 version', 'ie >= 11']
+          browsers: ['last 1 version', 'not ie > 0']
```
oskarhane, line 19: 👍

src/browser/modules/Stream/errorMessageFormater.js
```diff
@@ -18,20 +18,14 @@
...
+const error = title => ({
   title,
-  message: message || title
+  message: title
 })
 
-const knownErrors = {
-  18: IESecurityError18
-}
-
```
oskarhane, line 31: Good catch

build_scripts/webpack-rules.js
HerrEmil: It was commented out, as far as I can see, it didn't do anything

oskarhane: 😆 sorry, I missed the comment. Good cleanup

PR #1133, opened 2020-07-07
Connect labels to form inputs

OskarDamkjaer:
While doing the walk through with Liza I noticed some labels were not clickable. I connected the labels to the inputs and made some smaller readability changes

image

### DIFF
--- a/src/browser/components/Form.jsx
+++ b/src/browser/components/Form.jsx
@@ -110,13 +110,9 @@ export const TextInput = props => {
   return <StyledSettingTextInput {...rest}>{children}</StyledSettingTextInput>
 }
 
-export const CheckboxSelector = props => {
-  return props.checked ? (
-    <StyledCheckbox type="checkbox" {...props} />
-  ) : (
-    <StyledCheckbox type="checkbox" {...props} />
-  )
-}
+export const CheckboxSelector = props => (
+  <StyledCheckbox type="checkbox" {...props} />
+)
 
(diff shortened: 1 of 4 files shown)


src/browser/components/Form.jsx
```diff
@@ -110,13 +110,9 @@ export const TextInput = props => {
...
-export const CheckboxSelector = props => {
-  return props.checked ? (
-    <StyledCheckbox type="checkbox" {...props} />
-  ) : (
-    <StyledCheckbox type="checkbox" {...props} />
-  )
-}
+export const CheckboxSelector = props => (
+  <StyledCheckbox type="checkbox" {...props} />
+)
```
OskarDamkjaer, line 115: Is there some way where these are not equal? Esp. since styled checkbox is a simple input field and doesn't have any side effects?

src/browser/modules/Sidebar/Settings.jsx
```diff
@@ -228,12 +234,13 @@ export const Settings = ({
                 checked={settings[setting]}
                 data-testid={setting}
               />
-              <StyledSettingLabel title={tooltip}>{visual}</StyledSettingLabel>
-            </StyledSetting>
-          )
-        }
-      })
-      .filter(setting => setting !== false)
```
OskarDamkjaer, line 236: I removed this fail-safe check since I figure it's better to have the sidebar crash on malformed form items. Thoughts?

PR #1135, opened 2020-07-09
Generate both fields

OskarDamkjaer:
* Removed the classname dependency:
[Here's a gif to show the focus class is still applied like before](http://g.recordit.co/XIr0foLfRe.gif)

* Fixed some misalignment on the pop-up banner. It used to look like [this](http://g.recordit.co/mpbCgJ2euC.gif), now it looks like [this](http://g.recordit.co/O9G7MdpXQp.gif). 

I'm out of ideas on how to make this look better. The popup now instead misaligns when the scrollbar is present instead. 
image

Edit: This is the current best idea i have. It does make the banner scroll as well.
image

* The password is now generated for both fields, and is show by default. You'll need to show it to know your password anyway. [Like this](http://g.recordit.co/CjvGvJ6OKr.gif)

Also, I think it would be clearer if the generate button instead said "suggest password", thoughts?

### DIFF
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -5637,32 +5637,6 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 -----
 
-The following software may be included in this product: mockdate. A copy of the source code may be downloaded from https://github.com/boblauer/MockDate.git. This software contains the following license and notice below:
-
-The MIT License (MIT)
-
-Copyright (c) 2014 Bob Lauer
-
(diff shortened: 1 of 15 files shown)


HerrEmil:
Check with design if banner should scroll or be locked

OskarDamkjaer:
Made the banner still stick to top, without the layout breaking when the scroll-bar appears. Do you think the approach is reasonale? 
Looks like so now:
http://g.recordit.co/OlB8uG6Eou.gif

PR #1140, opened 2020-07-14
Fix comment filtering on command

OskarDamkjaer:
Replaced the comment regex with a filter that easier to understand and doesn't have this issue:
image

However. This is an edge-case but could happen:
image
It seems to me from https://neo4j.com/docs/2.2.2/cypher-comments.html that cypher supports comments, any reason why we need to "clean the command"? Could it not simply be removed (perhaps some version stuff with cypher?).

From some initial looks this
image
Doesn't work if we let cypher handle the comments because then the second statement is evaluated as cypher instead of a command. So perhaps leaving the cleaning as it is is preferable? The current implementation has that awkward edgecase but at least the syntax highlighting indicates the something might be wrong.

### DIFF
--- a/src/browser/components/TabNavigation/Navigation.jsx
+++ b/src/browser/components/TabNavigation/Navigation.jsx
@@ -51,7 +51,7 @@ class Navigation extends Component {
 
   componentDidUpdate(prevProps, prevState) {
     if (prevProps.openDrawer !== this.props.openDrawer) {
-      var newState = {}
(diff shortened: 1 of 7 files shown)


OskarDamkjaer:
@pe4cey  Code mirror does strip comments but it's seems it just doesn't do it correctly all the time. We're moving away from codemirror so I thought it'd take to long to fix it in that code base instead

PR #1141, opened 2020-07-14
Update UX on :style usage

OskarDamkjaer:
Fixes this problem where semicolons break style when using multiline.
image

There's some behaviour changes, they are outlined as comments.

### DIFF
--- a/src/shared/modules/commands/commandsDuck.js
+++ b/src/shared/modules/commands/commandsDuck.js
@@ -161,7 +161,13 @@ export const handleCommandEpic = (action$, store) =>
       store.dispatch(clearErrorMessage())
       const maxHistory = getMaxHistory(store.getState())
       store.dispatch(addHistory(action.cmd, maxHistory))
-      const statements = shouldEnableMultiStatementMode(store.getState())
+
+      // Semicolons in :style grass break parsing of multiline statements from codemirror.
+      const useMultiLine =
+        !action.cmd.startsWith(':style') &&
+        shouldEnableMultiStatementMode(store.getState())
+
+      const statements = useMultiLine
         ? extractStatementsFromString(action.cmd)
         : [action.cmd]
 
--- a/src/shared/modules/commands/commandsDuck.js
+++ b/src/shared/modules/commands/commandsDuck.js
@@ -163,11 +163,11 @@ export const handleCommandEpic = (action$, store) =>
       store.dispatch(addHistory(action.cmd, maxHistory))
 
       // Semicolons in :style grass break parsing of multiline statements from codemirror.
-      const useMultiLine =
+      const useMultiStatement =
         !action.cmd.startsWith(':style') &&
         shouldEnableMultiStatementMode(store.getState())
 
-      const statements = useMultiLine
+      const statements = useMultiStatement
         ? extractStatementsFromString(action.cmd)
         : [action.cmd]
 
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -82,7 +82,7 @@ import {
   isValidURL
 } from 'shared/modules/commands/helpers/http'
 import { fetchRemoteGrass } from 'shared/modules/commands/helpers/grass'
-import { parseGrass } from 'shared/services/grassUtils'
+import { parseGrass, objToCss } from 'shared/services/grassUtils'
 import {
   shouldUseCypherThread,
   getCmdChar,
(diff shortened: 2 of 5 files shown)


src/shared/modules/commands/commandsDuck.js
```diff
@@ -161,7 +161,13 @@ export const handleCommandEpic = (action$, store) =>
       store.dispatch(clearErrorMessage())
       const maxHistory = getMaxHistory(store.getState())
       store.dispatch(addHistory(action.cmd, maxHistory))
-      const statements = shouldEnableMultiStatementMode(store.getState())
+
+      // Semicolons in :style grass break parsing of multiline statements from codemirror.
+      const useMultiStatement =
```
OskarDamkjaer, line 166: The best way I've found to solve the issue of `extractStatementsFromString` from cypher-codemirror can't parse grass is to avoid using it for style commands. We can't run style and some other command so it makes sense not to use multi-statement anyway. I'm not 100% happy with the magic string of ":style" and it's special handling to other commands though :/

src/shared/services/commandInterpreterHelper.js
```diff
@@ -709,10 +709,12 @@ const availableCommands = [
     name: 'style',
     match: cmd => /^style(\s|$)/.test(cmd),
     exec(action, cmdchar, put, store) {
-      const match = action.cmd.match(/:style\s*(\S.*)$/)
```
OskarDamkjaer, line 712: This regex didn't allow for style with newlines

src/shared/services/commandInterpreterHelper.js
```diff
@@ -722,41 +724,59 @@ const availableCommands = [
...
+        )
+      }
+
+      if (param === '') {
+        showGrass()
+        return
+      }
+
+      if (param === 'reset') {
+        updateGrassAndShow(null)
```
OskarDamkjaer, line 751: I've changed it to show a new pane on `:style reset` and `:style node {color: red}` to make it more obvious that something actually happens. If a user wants to see how their graph is changed they'd need to pin it for it not to scroll away though. It's not perfect this way but we want to rework how grass-editing works so it good for now

src/shared/services/commandInterpreterHelper.js
```diff
@@ -722,41 +725,66 @@ const availableCommands = [
...
+        put(updateGraphStyleData(newGrass))
+        showGrass()
+      }
+
+      function putGrassErrorFrame(message) {
+        put(
+          frames.add({
+            useDb: getUseDb(store.getState()),
+            ...action,
+            error: createErrorObject(InvalidGrassError, message),
```
OskarDamkjaer, line 740: Add a custom error screen for grass instead of sending "unknown error"

src/shared/services/commandInterpreterHelper.js
```diff
@@ -722,41 +725,66 @@ const availableCommands = [
...
+      }
 
+      if (param === 'reset') {
+        updateAndShowGrass(null)
+        return
+      }
+
+      if (
+        isValidURL(param) &&
+        param.includes('.') /* isValid url considers words like rest an url*/
```
jk05, line 758: is this comment suggesting `isValid url considers words like "rest" a url`?

src/shared/modules/commands/commandsDuck.js
jk05: Im not sure how 'special' situations are handled in Browser or if at all. The only comment I would have is whether a `@todo` comment here makes sense too or adding `:style` to some variable that holds non multi statement commands? May not be relevant, Im not sure

src/shared/services/commandInterpreterHelper.js
OskarDamkjaer: yes, It would be clearer if i added quotes perhaps? Even better would be to update the isValidURL to use a better regex though. It was used in quite a lot of places so I wasn't comfortable doing the change when I wrote this. I could look into it now though

src/shared/modules/commands/commandsDuck.js
OskarDamkjaer: When the allow multi statement setting is turned on, some commands (including :style) can't be part of a multistatement but should still work on their own. The problem is that :style is parsed in such a way that browser thinks it's multiple statments, since cypher-code mirror util we have can't parse grass to split the commands properly. So i don't think the variable would be helping and I'm not sure what to even write in a todo comment.. I'll write it down and discuss with hane/emil when they get back

PR #1143, opened 2020-07-21
Make eslint lint tsx files and resolve you aliases

OskarDamkjaer:


nglgzz:
Just to add some more context, `jsconfig.json` seems to be used by VS Code to determine the context of JS files. According to the [VS Code docs](https://code.visualstudio.com/docs/languages/jsconfig) it is equivalent to a `tsconfig.json` file with the `allowJs` attribute set to true.

PR #1144, opened 2020-07-22
Make sidebar more consistent and add ts

OskarDamkjaer:
Previously the favorites tab didn't have any header:
bild
And the some tabs had different names on hover compared to their headers.


.gitignore
```diff
@@ -15,7 +15,6 @@ dist/
 NOTICE.xml
 *.tgz
 *.tar.gz
-release-notes.md
```
OskarDamkjaer, line 18: this file was removed a while ago

package.json
```diff
@@ -135,7 +137,7 @@
   "dependencies": {
     "@literal-jsx/parser": "^0.1.7",
     "@mdx-js/runtime": "^1.6.1",
-    "@neo4j/browser-lambda-parser": "1.0.4",
+    "@neo4j/browser-lambda-parser": "1.0.5",
```
OskarDamkjaer, line 140: update needed for unrelated bug fix

src/browser/modules/Sidebar/About.tsx
```diff
@@ -142,19 +158,7 @@ const About = ({ serverVersion, serverEdition }) => (
...
-  }
-  const tokenisedServerVersion = serverVersion && serverVersion.split('.')
-  const releaseTag = tokenisedServerVersion.join('')
-  const urlServerVersion =
-    serverVersion && tokenisedServerVersion.splice(0, 2).join('.')
-  return `https://github.com/neo4j/neo4j/wiki/Neo4j-${urlServerVersion}-changelog#${releaseTag}`
-}
-
-const mapStateToProps = state => {
+const mapStateToProps = (state: any) => {
```
OskarDamkjaer, line 161: Getting a better solution to state types didn't seem trivial. It'll be something to look into in the future

.gitignore
oskarhane: IMO: We want to keep ignoring this so CI doesn't have to think about it when it's generated

OskarDamkjaer: Oh! I didn't know it was generated! I just saw the one that was years old and assumed it's something we no longer did. I'll put it back in

package.json
OskarDamkjaer: Fixes #1112

PR #1147, opened 2020-07-24
Fix readme e2e test

sadeghpro:
Hi,
I use e2e to test my work with `yarn e2e --env server=4.0` but it's failed at first then I check readme and see it's said fresh installation of neo4j, I create new docker and first test pass but some other test failed again I checked and found that I use docker latest image and it's version 4.1.1 and when use `yarn e2e --env server=4.1` it's worked good.
I think maybe this helps others

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ with modern and easy-to-use tools.
 
 #### E2E Suite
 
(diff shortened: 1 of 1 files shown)


OskarDamkjaer:
Thanks!

PR #1148, opened 2020-07-27
Cypress improvements

OskarDamkjaer:
Adds better integration in vscode for writing tests by using a tsconfig for cypress.
Adds webpack to cypress so we can directly import code from our project.
Gives access to the store so we can dispatch actions programmatically instead of typing all commands
Adds tests to see that typing manually still works as intended
Makes cypress stop on console errors, show them in the video and leave a log file in the artifact for easier debugging in teamcity


e2e_tests/integration/loadcsv.spec.js
```diff
@@ -37,24 +37,22 @@ describe('LOAD CSV', () => {
     }
     cy.executeCommand(':clear')
     cy.executeCommand('MATCH (n) DETACH DELETE n')
-    cy.executeCommand(`LOAD CSV WITH HEADERS FROM 'file:///import.csv' AS row{shift}{enter}
```
OskarDamkjaer, line 40: Cypress uses the {} syntax for special chars and {{} for typing {. Since we send the strings right to redux now and don't type them, they are no longer needed.

e2e_tests/plugins/index.js
```diff
@@ -1,17 +1,9 @@
...
-// ***********************************************************
+const webpack = require('@cypress/webpack-preprocessor')
 
-// This function is called when a project is opened or re-opened (e.g. due to
-// the project's config changing)
-
-module.exports = (on, config) => {
-  // `on` is used to hook into various events Cypress emits
-  // `config` is the resolved Cypress config
+module.exports = on => {
```
OskarDamkjaer, line 3: To be able to import our redux actions we need to use webpack (handles typescript/aliased imports/etc)

e2e_tests/support/commands.js
```diff
@@ -1,10 +1,18 @@
-import { isAura } from './utils'
```
OskarDamkjaer, line 1: unused

e2e_tests/integration/loadcsv.spec.js
oskarhane: I'll review properly next week, I just wanted to point out that I don't like the idea if bypassing user interfaces in e2e tests.

I don't know the motivation behind it in this case, but generally speaking e2e tests should mimic user interactions.

OskarDamkjaer: Then let's wait until you're back! It's from my interpretation of cypress best practices https://docs.cypress.io/guides/references/best-practices.html

OskarDamkjaer:
Yes, this is my view as well. Will rebase and update pr!

e2e_tests/plugins/index.js
HerrEmil: This comment is useful when reading code, I think it should be added in source

PR #1155, opened 2020-08-03
Card view

OskarDamkjaer:
PR moves buttons inside editor like so: 
bild

Multiline now looks:
bild

Full screen is really fullscreen:
bild

Current multiline for reference:
bild

Colors, transitions, positioning, icons and other cosmetics still need more work though.

The gifs' framerate is to low to give it justice but there are some transitions: http://g.recordit.co/akBSL40Dhs.gif

### DIFF
--- a/src/browser/custom.d.ts
+++ b/src/browser/custom.d.ts
@@ -1,0 +1,4 @@
+declare module '*.svg' {
+  const content: string
+  export default content
+}
--- a/src/browser/modules/Editor/ActionButtons.tsx
+++ b/src/browser/modules/Editor/ActionButtons.tsx
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2002-2020 "Neo4j,"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react'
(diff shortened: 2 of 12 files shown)


src/browser/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
HerrEmil, line 1: Check if svgs as component dependencies have @typed that we could use

src/browser/modules/Editor/ActionButtons.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react'
+import { ActionButtonSection } from './styled'
+import { EditorButton } from 'browser-components/buttons'
+export interface ActionButtonProps {
```
HerrEmil, line 24: Remove export on things that are not imported so we can see if they are unused in future

src/browser/modules/Editor/styled.jsx
```diff
@@ -24,51 +24,80 @@ import { dim } from 'browser-styles/constants'
...
+      // 230 is 10 lines + 2*12px padding
+      return '254px'
+    }
+    return '0'
+  }};
+
+  ${props => (props.expanded ? '' : 'transition-duration: 0.1s;')}
+
   .CodeMirror {
-    background-color: ${props => props.theme.editorBackground} !important;
```
HerrEmil, line 71: Neo4j Browser - Google Chrome 8_4_2020 11_25_35 AM

Looks off in dark mode

OskarDamkjaer: Should be fixed now:
image

I do think the dark mode editor is too black, it could be the same color as the frames? Also, the header color doesn't fit as well in dark mode either

src/browser/custom.d.ts
OskarDamkjaer: Looking into types for react-svg-inline, didn't seem to solve the problem. I think this way works well enough. React-svg-inline typed it's icon-prop to be a string.

src/browser/modules/Editor/ActionButtons.tsx
OskarDamkjaer: done!

OskarDamkjaer:
@HerrEmil  Mergeable now? :)

src/browser/modules/Editor/styled.jsx
HerrEmil: Both light and dark mode look correct to me on Windows :)

PR #1157, opened 2020-08-07
Add multi statement query warning when multi statement is disabled

HerrEmil:
When the editor checks the current query with `explain`, the potential mistake of sending multiple statements in one query is not caught. Frontend has the info to flag this issue so this PR adds an extra check on editor update, and displays the warning as a warning triangle in the editor gutter.

multi statement warning

When clicking other warnings in the editor, an explain query is sent and the result displayed in a warning frame. Since this warning is caught in frontend, that query is skipped and a simple error frame with a message is displayed.

### DIFF
--- a/src/browser/modules/Editor/Editor.jsx
+++ b/src/browser/modules/Editor/Editor.jsx
@@ -274,10 +274,13 @@ export class Editor extends Component {
   checkForMultiStatementWarnings(statements) {
     if (statements.length > 1 && !this.props.enableMultiStatementMode) {
       const { offset, line, column } = statements[1].start
+      const message =
+        'To use multi statement queries, please enable multi statement in the settings panel.'
       this.setState({
         notifications: [
           {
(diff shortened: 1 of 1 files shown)


OskarDamkjaer:
Nice little video!  I get this though.. 
image
It sticks even when I remove the query and clear the local storage

OskarDamkjaer:
I think it'd be really nice if the warning went away automatically when the box got ticked, not sure if out of scope though :)

HerrEmil:
> Nice little video! I get this though..
> image
> It sticks even when I remove the query and clear the local storage.
> 
> I'll try to see if i can find what went wrong

Thanks for flagging these, both were my bad and I've found how to fix them :)

HerrEmil:
> I think it'd be really nice if the warning went away automatically when the box got ticked, not sure if out of scope though :)

I agree, I will try adding that

PR #1161, opened 2020-08-17
Shortcuts

OskarDamkjaer:
Adds keyboard shortcuts to resize the editor. Currently uses cmd+alt+c and cmd+alt+f as shortcuts. Updates the docs and hovers as well.
gif

@HerrEmil make sure to try the shortcuts and see that the docs/hover shows the correct symbols/words on windows.

### DIFF
--- a/src/browser/components/icons/IconContainer.jsx
+++ b/src/browser/components/icons/IconContainer.jsx
@@ -62,7 +62,12 @@ export class IconContainer extends Component {
 
     const currentIcon = icon ? (
       <StyledIconWrapper {...rest}>
-        <SVGInline svg={icon} accessibilityLabel={title} width={width + 'px'} />
+        <SVGInline
+          cleanup={['title']}
+          svg={icon}
+          accessibilityLabel={title}
+          width={width + 'px'}
+        />
       </StyledIconWrapper>
     ) : (
       <StyledIconWrapper {...rest} style={regulateSizeStyle} />
--- a/src/browser/documentation/help/keys.jsx
+++ b/src/browser/documentation/help/keys.jsx
@@ -19,6 +19,13 @@
  */
 
 import React from 'react'
+import {
+  FULLSCREEN_SHORTCUT,
+  FOCUS_SHORTCUT,
+  CARDSIZE_SHORTCUT,
+  printShortcut,
+  isMac
+} from 'browser/modules/App/keyboardShortcuts'
 const title = 'Keys'
 const subtitle = 'Keyboard shortcuts'
 const category = 'browserUiCommands'
@@ -36,7 +43,9 @@ const content = (
         <tr>
           <td>Execute current command</td>
           <td>
-            <div className="key code">{'<Ctrl-Return>'}</div>
+            <div className="key code">
+              {isMac ? '<Cmd-Return>' : '<Ctrl-Return>'}
+            </div>
           </td>
           <td>
             <div className="key code">{'<Return>'}</div>
@@ -45,7 +54,9 @@ const content = (
         <tr>
           <td>Previous command in history</td>
           <td>
-            <div className="key code">{'<Ctrl-Up-Arrow>'}</div>
+            <div className="key code">
+              {isMac ? '<Cmd-Up-Arrow>' : '<Ctrl-Up-Arrow>'}
+            </div>
           </td>
           <td>
             <div className="key code">{'<Up-Arrow>'}</div>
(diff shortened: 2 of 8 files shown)


src/browser/components/icons/IconContainer.jsx
```diff
@@ -62,7 +62,12 @@ export class IconContainer extends Component {
 
     const currentIcon = icon ? (
       <StyledIconWrapper {...rest}>
-        <SVGInline svg={icon} accessibilityLabel={title} width={width + 'px'} />
+        <SVGInline
+          cleanup={['title']}
```
OskarDamkjaer, line 66: Previously if you hover exactly the svg, they show the name of the svg.

src/browser/documentation/help/keys.jsx
```diff
@@ -78,30 +91,24 @@ const content = (
...
         <tr>
-          <td />
-        </tr>
-        <tr>
-          <th>Platform specific</th>
-          <th />
-          <th />
-        </tr>
-        <tr>
-          <td>Mac users</td>
```
OskarDamkjaer, line 99: Changed to show the relevant shortcuts directly instead of noting the differences.

src/shared/modules/editor/editorDuck.js
```diff
@@ -32,6 +32,7 @@ export const SET_CONTENT = `${NAME}/SET_CONTENT`
 export const EDIT_CONTENT = `${NAME}/EDIT_CONTENT`
 export const FOCUS = `${NAME}/FOCUS`
 export const EXPAND = `${NAME}/EXPAND`
+export const CARDSIZE = `${NAME}/CARDSIZE`
```
OskarDamkjaer, line 35: I would want the expand/collapse state to live in redux in the future.

src/browser/modules/App/keyboardShortcuts.ts
```diff
@@ -0,0 +1,93 @@
...
+    keyboardShortcuts.forEach(shortcut =>
+      document.addEventListener('keydown', shortcut)
+    )
+
+    return (): void =>
+      keyboardShortcuts.forEach(shortcut =>
+        document.removeEventListener('keydown', shortcut)
+      )
+  }, [])
+}
```
OskarDamkjaer, line 93: will probably move these to be a single function instead of multiple as we add more shortcuts

src/browser/modules/App/keyboardShortcuts.ts
```diff
@@ -0,0 +1,94 @@
...
+const modKey = isMac ? 'metaKey' : 'ctrlKey'
+
+type ModifyerKey = 'metaKey' | 'altKey' | 'ctrlKey'
+const printModifyers: Record<ModifyerKey, string> = {
+  altKey: isMac ? '⌥' : 'alt',
+  metaKey: isMac ? '⌘' : 'win',
+  ctrlKey: isMac ? '⌃' : 'ctrl'
+}
+
+const encodeKey: Record<string, number> = { c: 67, f: 70, '/': 55 }
```
HerrEmil, line 15: I don't understand these values, they are the ascii values for C, F and 7? I think there's a mistake, I can focus the editor by pressing `7` now, but not the `/` key on this branch.

src/browser/modules/App/keyboardShortcuts.ts
```diff
@@ -0,0 +1,94 @@
+import { useEffect } from 'react'
+import { FOCUS, EXPAND, CARDSIZE } from 'shared/modules/editor/editorDuck'
+import { Bus } from 'suber'
+
+export const isMac = /Mac|iPad/.test(navigator.platform)
+const modKey = isMac ? 'metaKey' : 'ctrlKey'
+
+type ModifyerKey = 'metaKey' | 'altKey' | 'ctrlKey'
```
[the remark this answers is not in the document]
OskarDamkjaer: Ahh, I knew something felt wrong, good catch! I googled and found hits for "The Modifyers" and didn't think any more of it 😅

src/browser/modules/App/keyboardShortcuts.ts
OskarDamkjaer: They are the event keycode values, which I also thought were ascii values but turns out they aren't. Pressing the cmd+alt+c here shows keycode 67 instead of 99 which from "c".charCodeAt(0)..

HerrEmil: I think they are ascii values, but for uppercase C and F

PR #1162, opened 2020-08-17
Move to ts-loader, hot reloading, fast type check

OskarDamkjaer:
Looked into ways to speed up the dev build and making hot reloading to work in [this part of the webpack docs](https://webpack.js.org/guides/hot-module-replacement/). For hot reloading of react components they link [here](https://github.com/gaearon/react-hot-loader), that project in recommends we use [this](https://github.com/pmmmwh/react-refresh-webpack-plugin) instead. Inspired by one of their example [setups](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/examples/typescript-without-babel/webpack.config.js) i added hot module reloading.

I found awesome-ts-loader was unmaintained so I moved us over to ts-loader instead. Using their quick [type check example](https://github.com/TypeStrong/ts-loader/tree/master/examples/fork-ts-checker-webpack-plugin
) as inspiration. 

From their examples we also got better eslint checking, so I fixed some new warnings and errors. These kind of config changes are quite fiddly, so I expect at least something to be sub optimal


build_scripts/webpack.config.js
```diff
@@ -57,42 +57,48 @@ module.exports = {
...
       }
-    }
-  },
-  devtool: helpers.isProduction ? false : 'inline-source-map',
+    : {
+        removeAvailableModules: false,
+        removeEmptyChunks: false,
+        splitChunks: false
+      },
+  devtool: helpers.isProduction ? false : 'eval-cheap-module-source-map',
```
OskarDamkjaer, line 101: webpack recommended these settings for increased dev env speed

src/browser/components/EditionView.jsx
```diff
@@ -30,7 +30,7 @@ export const EnterpriseOnlyFrame = ({ command }) => {
       </p>
       <p>
         Find out more over at{' '}
-        <a href="https://neo4j.com/editions/" target="_blank">
+        <a href="https://neo4j.com/editions/" target="_blank" rel="noreferrer">
```
OskarDamkjaer, line 33: Eslint now complains about "_blank" links without norefferrer because of [this vurn](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

src/browser/modules/Editor/ActionButtons.tsx
```diff
@@ -34,7 +34,10 @@ interface ActionButton {
   iconColor?: string
 }
 
-const ActionButtons = ({ buttons, width = 24 }: ActionButtonProps) => (
```
OskarDamkjaer, line 37: This component is hit by the rule of that all exported things need to be fully typed, so I added a return type back

.eslintrc.json
```diff
@@ -36,19 +36,28 @@
...
+      "parser": "@typescript-eslint/parser",
       "files": ["*.ts", "**/*.ts", "**/*.tsx"],
       "extends": [
-        "plugin:@typescript-eslint/recommended",
-        "prettier/@typescript-eslint"
+        "plugin:@typescript-eslint/recommended", // Uses the recommended rules from the @typescript-eslint/eslint-plugin
+        "prettier/@typescript-eslint", // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
+        "plugin:react/recommended", // Uses the recommended rules from @eslint-plugin-react
+        "prettier/react" // disables react-specific linting rules that conflict with prettier
+        // 'plugin:prettier/recommended' // Enables eslint-plugin-prettier and displays prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
```
HerrEmil, line 51: Leftover commented out row

src/shared/modules/commands/useDb.test.js
```diff
@@ -1,76 +0,0 @@
-/*
```
HerrEmil, line 1: Why was this test removed?

OskarDamkjaer: Removed it since it's skipped and has been skipped for a long time. Should i put it back in?

HerrEmil: Looking at history, it was skipped with the comment "Ignoring one test for now. Need to mock `driver.supportsMultiDb()`". I agree that there no reason to keep skipping it, either remove or add that mock is better than skipping

PR #1163, opened 2020-08-18
Add animation to editor

OskarDamkjaer:
Adds an animation to the editor to give meaning to the close action.
CleanShot-2020-08-18-at-135933


HerrEmil:
When the new editor is animated in, the initially reserved space is still there and makes notification and cards below jump:
ezgif-2-d073738c625c

Doesn't happen when there's scrolling, only if stream fits on screen

PR #1164, opened 2020-08-19
Tag bug

OskarDamkjaer:
Current behaviour when a prop has <tag> in it:
bild
now:
bild

We were setting innerHTML and sanitizing with a library. It seemed something was wrong with the library, but I didn't investigate as I found a better solution on how to do it without any external code instead.


package.json
```diff
@@ -175,7 +175,6 @@
     "redux-observable": "^0.16.0",
     "regenerator-runtime": "^0.13.2",
     "rxjs": "^5.4.2",
-    "sanitize-html": "^1.21.1",
```
OskarDamkjaer, line 178: no longer needed 🎉

src/browser/components/ClickableUrls.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import React from 'react'
```
OskarDamkjaer, line 18: I've updated these tests to typescript and made them snapshot tests of the render clickable links components.

src/browser/components/ClickableUrls.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export default function ClickableUrls({
+  text = '',
+  WrappingTag = 'span'
+}: ClickableUrlsProps): JSX.Element {
+  const urls = text.match(URL_REGEX) || []
+  return (
+    <WrappingTag>
+      {text.split(URL_REGEX).map((text, index) => {
+        /* since we never move these components this key should be fine */
```
OskarDamkjaer, line 36: I can't think of a proper key for this fully dynamic content? Using a UUID is possible 🤔

PR #1165, opened 2020-08-19
Ensure frames.allIds is always an array

HerrEmil:
A user reported getting a TypeError:
Screen Shot 2020-08-03 at 6 14 27 PM

We were unable to reproduce the issue, but we can at least prevent the type error by strictly typing the variable and updating the code that could assign a different type value. The variable is the list of frames in the stream and the updated code is called when pinning or unpinning a frame in stream. If the issue is reproduced after this change, the failure could potentially result in an out-of-order list or a cleared list, but never a TypeError.

### DIFF
--- a/src/shared/modules/stream/streamDuck.ts
+++ b/src/shared/modules/stream/streamDuck.ts
@@ -24,6 +24,9 @@ import 'rxjs/add/operator/mapTo'
 import { moveInArray } from 'services/utils'
 import { APP_START } from 'shared/modules/app/appDuck'
 import { UPDATE as SETTINGS_UPDATE } from '../settings/settingsDuck'
+import { Epic } from 'redux-observable'
+import { Action } from 'redux'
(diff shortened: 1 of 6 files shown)


src/shared/modules/stream/streamDuck.ts
```diff
@@ -185,43 +189,99 @@ function ensureFrameLimit(state) {
...
+export default function reducer(
+  state = initialState,
+  action: StreamActions
+): FramesState {
   switch (action.type) {
     case APP_START:
       return { ...initialState, ...state }
     case ADD:
-      return addFrame(state, action.state)
+      return addFrame(state, (action as AddFrameAction).state)
```
OskarDamkjaer, line 257: Use string literals instead of template strings for types, so you don't need to cast!

OskarDamkjaer (approved):

PR #1167, opened 2020-08-21
Fix fullscreen issue in safari

OskarDamkjaer:
On safari fullscreen looked like this:
bild

Now it looks like it should:
bild

In my testing, it seems the z-index: 1 causing this was unneeded, but please do take a look and see if anything looks off.


HerrEmil (approved):
Works well in windows browsers as well

PR #1168, opened 2020-08-21
add aura enterprise edition check

HerrEmil:
Enables `:server user list` and `:server user add` frames to be displayed when connected to an aura instance

changelog: Recognize new Neo4j Aura editions

### DIFF
--- a/src/shared/modules/dbMeta/dbMeta.utils.js
+++ b/src/shared/modules/dbMeta/dbMeta.utils.js
@@ -43,11 +43,11 @@ export function extractServerInfo(res) {
   }
 
   // Temporarily hardcoded solutions for Aura
-  if (serverInfo.version === '4.0-aura') {
+  if (['4.0-aura', '4.0-AuraProfessional'].includes(serverInfo.version)) {
     serverInfo.version = '4.0.0'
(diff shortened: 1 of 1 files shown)


OskarDamkjaer:
LGMT! Let's just wait for the tests :)

e2e_tests/support/utils.js
```diff
@@ -3,7 +3,7 @@
 export const isEnterpriseEdition = () =>
   Cypress.config('serverEdition') === 'enterprise'
 
-export const isAura = () => Cypress.config('serverEdition') === 'aura'
+export const isAura = () => Cypress.config('serverEdition').startsWith('aura')
```
unterstein, line 6: I guess this should do a conversion to lower case before the startsWith is called, since we are currently migration away from the lower case `aura` to the uppercase `AuraProfessional` and `AuraEnterprise`

HerrEmil: This is comparing against an internal browser value, which is `aura` or `auraenterprise`

unterstein: ah ok sorry, then forget my comment 🙏 

Thank you 🎉

PR #1169, opened 2020-08-24
Show comment in frame header

OskarDamkjaer:
Leaves the comments in place when there's only one comment.

Properly resizes editor when moving through history and setting from clicking a frame header.


HerrEmil (approved):
LGTM 🌞

PR #1174, opened 2020-08-27
Update animation and minor ui details

HerrEmil:
Makes the update favorite icon look more clickable.
Show a shortcut on run.
Naming update on the discard icon.
Animation changes is to prevent static height and elements jumping


oskarhane (approved):
LGTM 👍 
Merge on green tests (I think the E2E tests went red on the origin PR, not sure why)

PR #1180, opened 2020-09-02
Fix flaky animation

OskarDamkjaer:
Move to position absolute solution for the double editor problem, and make animation smaller


HerrEmil (approved):
Animation runs smoothly on 6x CPU throttling and does not seem to get stuck

PR #1185, opened 2020-09-04
reimplement isValidURL to fix performance bug

HerrEmil:
Current implementation of isValidURL is very slow for some values, which was noticed when a user set `EnableMultiStatement:true` which takes around 2 minutes(!) on my machine. I don't know what part of the regex is slow, but with all web browser we support now supporting the URL constructor we can remove the regex to fix the issue.
Sep-04-2020 13-59-39


oskarhane (approved):
Nice one 🎧

OskarDamkjaer (approved):
Nice solution!

PR #1190, opened 2020-09-07
Ts migrate e2e tests

HerrEmil:
Tried out using the tool `ts-migrate` on the e2e_tests folder, as a trial-run before using the tool to convert src folder.


e2e_tests/integration/bolt.spec.ts
```diff
@@ -20,7 +20,7 @@
 
 import { isEnterpriseEdition } from '../support/utils'
 
-/* global Cypress, cy, test, expect, before */
+/* global Cypress, cy, before */
```
OskarDamkjaer, line 23: Do we need these comments? Can't it be set in the tsconfig.json for cypress?

e2e_tests/integration/multi-db.spec.ts
```diff
@@ -42,14 +42,13 @@ describe('Multi database', () => {
...
   })
-  after(() => {})
   it('can connect', () => {
     const password = Cypress.config('password')
     cy.connect('neo4j', password)
   })
   if (Cypress.config('serverVersion') >= 4.0) {
     if (isEnterpriseEdition()) {
-      it('shows a message indicating whether system updates have occured', () => {
+      it('shows a message indicating whether system updates have occurred', () => {
```
OskarDamkjaer, line 51: nice

e2e_tests/integration/play-command.spec.ts
```diff
@@ -18,11 +18,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* global Cypress, cy, test, expect, before */
+/* global Cypress, cy, before */
 
+import { ClearEditorButton } from './0.index.spec'
```
OskarDamkjaer, line 23: Looks a bit weird importing these from a .spec file. You mentioned you were looking into best practises, did you find a better way?

e2e_tests/support/global.d.ts
```diff
@@ -0,0 +1,108 @@
+/// <reference types="cypress" />
+
+export {}
```
OskarDamkjaer, line 3: Where did this file come from? Is there no @types/cypress?

OskarDamkjaer (approved):
Neat. Let's just wait for the tests to pass

e2e_tests/integration/multi-db.spec.ts
HerrEmil: Courtesy of  the VS Code plugin `Code Spell Checker` :)

e2e_tests/integration/play-command.spec.ts
HerrEmil: The recommended way I found was to just put the selectors in the spec file they're used, as we're doing, but since typescript thinks they are in the same scope I thought either import or declare in several places and silence typescript with `ts-expect-error`. But come to think of it, it's cleaner to remove them and add a custom clear command instead.

e2e_tests/support/global.d.ts
HerrEmil: Cypress ships with type definitions, these type definitions I wrote here are merged with the standard ones and are just to have definitions for the custom commands and config variables we add

e2e_tests/integration/bolt.spec.ts
HerrEmil: We could clean up making type definitions available, these are from before we had a tsconfig, the edits I made here are just to fix tslint warnings which caught that some imported types were unused

e2e_tests/integration/play-command.spec.ts
HerrEmil: Apparently importing anything from another spec causes all tests in the file with the export to be included in the suite that is importing, so they run several times, so that's a no-go. Each selector was only used once in each of the three suites so I decided to inline the selector.

e2e_tests/support/global.d.ts
OskarDamkjaer: Ah, nice!

PR #1194, opened 2020-09-15
Update :server connect form to pick database

OskarDamkjaer:
Introduces a `db` url param for selecting a database and `dbms` as an alternative param to connectURL.

Prefills form from the url like so:
bild


src/browser/components/buttons/index.jsx
```diff
@@ -217,27 +217,27 @@ export const FormButton = props => {
 
   if (icon && label) {
     return (
-      <ButtonType {...rest} type="button">
+      <ButtonType type="button" {...rest}>
```
OskarDamkjaer, line 220: The other order overrides type=submit

src/shared/modules/connections/connectionsDuck.js
```diff
@@ -37,7 +39,6 @@ import {
 import { inWebEnv, USER_CLEAR, APP_START } from 'shared/modules/app/appDuck'
 
 export const NAME = 'connections'
-export const ADD = 'connections/ADD'
```
OskarDamkjaer, line 40: Unused

src/shared/modules/connections/connectionsDuck.js
```diff
@@ -165,25 +163,17 @@ const removeConnectionHelper = (state, connectionId) => {
...
-    allConnectionIds,
-    connectionsById
+    connectionsById: {
+      ...connectionsById,
+      [id]: { ...connectionsById[id], ...connection }
+    },
+    allConnectionIds: allConnectionIds.includes(id)
+      ? allConnectionIds
+      : [...allConnectionIds, id]
   }
```
OskarDamkjaer, line 177: These method cleanup is think are cleaner, and in passing they also fixed a small bug that left a connection ID "undefined" in the connectionsById state

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -79,17 +79,21 @@ export const getBoltHost = state => {
...
+    'connectTo',
+    'restApi',
+    'showDbField'
+  ]
+  const updateObj = keysToCopy.reduce(
+    (accObj, key) => (action[key] ? { ...accObj, [key]: action[key] } : accObj),
+    { host: action.forceURL }
+  )
+
   if (typeof action.encrypted !== 'undefined') {
```
OskarDamkjaer, line 94: Is there a reason this was checked to be undefined rather than truthyness? @oskarhane

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -112,9 +116,15 @@ export const discoveryOnStartupEpic = (some$, store) => {
...
-      const passedURL = getUrlParamValue('connectURL', action.url)
+      const passedURL =
+        getUrlParamValue('dbms', action.url) ||
+        getUrlParamValue('connectURL', action.url)
+
+      const passedDb = getUrlParamValue('db', action.url)
+
       if (!passedURL || !passedURL.length) return action
       action.forceURL = decodeURIComponent(passedURL[0])
+      action.connectTo = passedDb && passedDb[0]
```
OskarDamkjaer, line 127: Right now db param is only respected if there's also a dbms/connectURL param. Just a note on behaviour

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -124,14 +134,12 @@ export const discoveryOnStartupEpic = (some$, store) => {
...
       if (action.forceURL) {
-        const { username, password, protocol, host } = getUrlInfo(
-          action.forceURL
-        )
+        const { username, protocol, host } = getUrlInfo(action.forceURL)
         updateDiscoveryState(
           {
             ...action,
             username,
-            password,
```
OskarDamkjaer, line 134: No longer keep password from url param

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -124,14 +134,12 @@ export const discoveryOnStartupEpic = (some$, store) => {
...
+            showDbField: !!action.connectTo,
```
OskarDamkjaer, line 142: Not sure if this boolean is the best way to signal if the field should show up or not, was thinking of just checking for the presence of connectTo, but that is a bit awkward when using the version check bellow since there's no value to prefill.

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -156,7 +167,11 @@ export const discoveryOnStartupEpic = (some$, store) => {
               getAllowedBoltSchemes(store.getState()),
               host
             )
-            store.dispatch(updateDiscoveryConnection({ host })) // Update discovery host in redux
+            const supportsMultiDb =
+              parseInt((result.neo4j_version || '0').charAt(0)) >= 4
```
OskarDamkjaer, line 171: I'm thinking there's a better way to get version number? Or is this fine?

src/browser/modules/Stream/Auth/ConnectForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  const onConnectClick = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+    setConnecting(true)
+    props.onConnectClick(() => setConnecting(false))
+  }
+
+  return (
+    <StyledConnectionForm onSubmit={onConnectClick}>
+      <FlexSpaceBetween>
```
OskarDamkjaer, line 110: Not a 100% happy with this styling, it feels more convoluted than it needs to be. I opted to make theses simple Flex components rather than inline style even if they are very simple, for consistency's sake

src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -431,11 +433,26 @@ export const dbMetaEpic = (some$, store) =>
...
+                  if (connectTo) {
+                    const wantedDb = databases.find(
+                      ({ name }) =>
+                        name.toLowerCase() === connectTo.toLowerCase()
+                    )
+                    if (wantedDb) {
+                      store.dispatch(useDb(wantedDb.name))
+                    } else {
+                      // this will show the db not found frame
+                      store.dispatch(executeSingleCommand(`:use ${connectTo}`))
```
OskarDamkjaer, line 450: I thought this was pretty nice error handling of databases that weren't found. Thoughts?

HerrEmil (approved):

src/browser/modules/Stream/Auth/ConnectForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+        </StyledConnectionFormEntry>
+        {props.showDbField && (
+          <StyledConnectionFormEntry>
+            <StyledConnectionLabel>
+              Database
+              <StyledConnectionTextInput
+                data-testid="database"
+                onChange={props.onDatabaseChange}
+                value={props.database}
+                placeholder="default database"
```
oskarhane, line 164: Maybe change this to "Database to use" or something

oskarhane (commented):
Cool, well done!

I have a few comments on naming and UI.

### Connect frame UI
I don't think this looks particularly good to be honest. The form looks very unusual.
Screen Shot 2020-09-18 at 15 05 12

### Auto-connect on refresh
It looks like the auto-connect doesn't work anymore. Successfully connect. Refresh page. You should auto-connect.

### Continuously trying to `:use`
In the scenario of connecting successfully but using a db that doesn't exist for some reason, it looks like it's running `:use` on every dbmeta-tick.
Screen Shot 2020-09-18 at 15 14 24

src/browser/modules/Stream/Auth/ConnectForm.tsx
OskarDamkjaer: You mean the label? I think the current placeholder makes sense

src/shared/modules/discovery/discoveryDuck.js
OskarDamkjaer: Yeah that's a better name!

src/shared/modules/discovery/discoveryDuck.js
OskarDamkjaer: The reason I didn't use useDb is because it think the same state already has a useDb. But connectTo is a bad name, so perhaps requestedUseDb is a better name?

src/shared/modules/discovery/discoveryDuck.js
OskarDamkjaer: So do you suggest we check that the version starts with 4 and is not running on aura?

OskarDamkjaer: Also, good catch!

oskarhane: Some aura editions do support multi-db, so it's tricky. I suggest you talk to Aura and ask which editions that has multi-db support and what version string those editions expose in the discovery endpoint.

src/browser/modules/Stream/Auth/ConnectForm.tsx
oskarhane: I don't :)
It's not a default db the users enters, it's the target db that the user wants to use when a connection is made.

src/shared/modules/discovery/discoveryDuck.js
OskarDamkjaer: Asked just now!

OskarDamkjaer:
Good catches, didn't see the screen shots you sent! Thanks a lot! :))

src/browser/modules/Stream/Auth/ConnectForm.tsx
OskarDamkjaer: Yeah and that's why the place holder is there to show what happens if you leave the field empty! Right now it think it makes sense to have a field titled "Database to use" and the default/placeholder being "default database". Having both be "database to use" I think would lead to me thinking the field was required

src/shared/modules/discovery/discoveryDuck.js
OskarDamkjaer: It seems aura only supports the system + 1 other database at this moment. I think it'd make sense to disable the field for aura, WDYT?

OskarDamkjaer:
Updated the form styling when I was thinking we'd always show the db field. Then it got really tall, but now when I know that's not the case I reverted to the older styling because it was simpler code wise and familiar.

OskarDamkjaer:
Not 100% sure the aura check works. What would be the best way to try it out? Not sure where to find the discovery response object

e2e_tests/integration/connect-form.spec.ts
```diff
@@ -76,10 +76,15 @@ describe('Connect form', () => {
     getBoltSchemeSelect().should('have.value', aliasScheme)
   })
 
-  it('can connect with bolt:// protocol', () => {
+  it.only('can connect with bolt:// protocol', () => {
```
oskarhane, line 79: Was this intentional?

OskarDamkjaer: No, I missed it after debugging the test. Should have read over the files one more time, good catch!

src/browser/modules/Stream/Auth/ConnectForm.tsx
oskarhane: I see your reasoning, but placeholders are hints on what the users should enter, not to indicate value if left blank.
The label could suggest "leave empty for default)" or something to let the users know what happens if it's left empty.

src/shared/modules/discovery/discoveryDuck.js
```diff
@@ -156,7 +168,11 @@ export const discoveryOnStartupEpic = (some$, store) => {
               getAllowedBoltSchemes(store.getState()),
               host
             )
-            store.dispatch(updateDiscoveryConnection({ host })) // Update discovery host in redux
+
+            const isAura = getEnv(store.getState()) === CLOUD
+            const supportsMultiDb =
+              !isAura && parseInt((result.neo4j_version || '0').charAt(0)) >= 4
```
oskarhane, line 174: Aren't there Aura editions that support multi-db?

e2e_tests/integration/connect-form.spec.ts
```diff
@@ -80,6 +80,11 @@ describe('Connect form', () => {
     cy.executeCommand(':clear')
     const boltUrl = 'bolt://' + stripScheme(Cypress.config('boltUrl'))
     cy.connect('neo4j', Cypress.config('password'), boltUrl)
+    cy.executeCommand('RETURN "Hello World";')
+    cy.contains('Hello World')
+    cy.reload()
+    cy.executeCommand('RETURN "Hello again";')
+    cy.contains('Hello again')
```
oskarhane, line 87: 👍

oskarhane (commented):
Good changes!

Still has a few more issues though:

1. When choosing something other than the default db, there's a delay until next dbMeta to update the meta items in the sidebar. See gif.
CleanShot 2020-10-01 at 08 46 30


1. Same thing when entering a db that doesn't exist for the user. It takes until next dbmeta tick to update the UI with the correct info. 
CleanShot 2020-10-01 at 08 51 25

Also, would it be possible to add e2e test for checking so browser actually switches to use the entered db?
And maybe the error case as well?

src/shared/modules/discovery/discoveryDuck.js
OskarDamkjaer: From asking there's only editions with proper support for multi-db. At most they support the system-db in addition to the default one. So I think if it's important there, you'll have to use the url param

e2e_tests/integration/connect-form.spec.ts
```diff
@@ -105,5 +111,68 @@ describe('Connect form', () => {
       cy.connect('neo4j', Cypress.config('password'), boltUrl)
       cy.executeCommand(':server disconnect')
     })
+
+    if (isEnterpriseEdition()) {
+      it.only('shows correct metadata when using db field', () => {
```
oskarhane, line 116: No `.only` please

src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -339,132 +483,20 @@ export const dbMetaEpic = (some$, store) =>
...
-                      store.dispatch(useDb(defaultDb[0].name))
-                    }
-                  }
-                  return Rx.Observable.of(null)
-                })
+          .mergeMap(() => {
+            return Rx.Observable.forkJoin([
+              getLabelsAndTypes(store),
+              clusterRole(store),
+              databaseList(store)
```
oskarhane, line 490: Even if one of these errors we'll need to keep the stream going, so I'f like to see some error catching here

oskarhane: And if I remember correctly, if one fail the other once are cancelled as well, so I think you need to catch on each one of the inner ones.

oskarhane (commented):
Seems to behave as expected now, well done!
You just need to be extra cautious to make sure one failure in the `forkJoin` doesn't cancel the other ones + cancels the whole listener.

But other than that, 👍 

Ping me to review when you're catching any errors and I'll approve it

e2e_tests/integration/connect-form.spec.ts
OskarDamkjaer: I'll set up an on commit hook for this so I won't forget it again

oskarhane (approved):
🌮 Merge on green tests 🌮

PR #1196, opened 2020-09-17
Add support for browser.remote_content_hostname_allowlist

OskarDamkjaer:

### DIFF
--- a/src/shared/modules/commands/helpers/config.test.js
+++ b/src/shared/modules/commands/helpers/config.test.js
@@ -21,6 +21,7 @@
 import nock from 'nock'
 import * as config from './config'
 import { update, replace } from 'shared/modules/settings/settingsDuck'
+import dbMetaReducer, { updateSettings } from 'shared/modules/dbMeta/dbMetaDuck'
 
 function FetchError(message) {
   this.name = 'FetchError'
@@ -35,7 +36,7 @@ describe('commandsDuck config helper', () => {
       return {
         meta: {
           settings: {
-            'browser.remote_content_hostname_whitelist': 'okurl.com'
+            'browser.remote_content_hostname_allowlist': 'okurl.com'
           }
         }
       }
@@ -137,7 +138,7 @@ describe('commandsDuck config helper', () => {
       expect(put).toHaveBeenCalledWith(replace({ x: 1, y: 2 }))
     })
   })
-  test('rejects hostnames not in whitelist', () => {
+  test('rejects hostnames not in allowlist', () => {
     // Given
     const action = { cmd: ':config https://bad.com/cnf.json' }
     const cmdchar = ':'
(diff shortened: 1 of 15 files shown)


src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -104,6 +104,7 @@ export const allowOutgoingConnections = state =>
 export const credentialsTimeout = state =>
   getAvailableSettings(state)['browser.credential_timeout'] || 0
 export const getRemoteContentHostnameWhitelist = state =>
+  state[NAME]['browser.remote_content_hostname_allowlist'] ||
```
OskarDamkjaer, line 107: Done this way because getAvailableSettings seems to be using initial state when not found already.

src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -205,6 +206,7 @@ const initialState = {
   databases: [],
   settings: {
     'browser.allow_outgoing_connections': false,
+    'browser.remote_content_hostname_allowlist': null,
```
OskarDamkjaer, line 209: Done this way so we get a falsy value when it's not set. Otherwise it will never read the whitelist

HerrEmil (approved):

src/shared/modules/dbMeta/dbMetaDuck.js
```diff
@@ -104,6 +104,7 @@ export const allowOutgoingConnections = state =>
 export const credentialsTimeout = state =>
   getAvailableSettings(state)['browser.credential_timeout'] || 0
 export const getRemoteContentHostnameWhitelist = state =>
```
oskarhane, line 106: Rename this too please

src/shared/modules/dbMeta/dbMetaDuck.js
oskarhane: wdyt about just having `browser.remote_content_hostname_allowlist` locally cached?
Browser front-end reads from single source (via selector `getRemoteContentHostnameAllowlist`), but both `browser.remote_content_hostname_allowlist` and `browser.remote_content_hostname_allowlist` from server responses write to that local cache?

src/shared/modules/commands/helpers/config.test.js
```diff
@@ -149,7 +149,33 @@ describe('commandsDuck config helper', () => {
...
+            }
+          }
+        }
+      }
+    })
+
+    // Then
+    return expect(p)
+      .rejects.toEqual(
+        new Error('Hostname is not allowed according to server allowlist')
```
oskarhane, line 178: Good to see tests 👍

src/shared/modules/dbMeta/dbMetaDuck.js
OskarDamkjaer: @oskarhane We say whitelist in tons of places, so it felt inconsistent to only rename it in one method. Should I go ahead and change all whitelist->allowlist or just this on?

src/shared/modules/dbMeta/dbMetaDuck.js
OskarDamkjaer: I've added this and a new test to check that it works now :)

src/shared/modules/dbMeta/dbMetaDuck.js
oskarhane: @OskarDamkjaer Yes, please

PR #1197, opened 2020-09-17
Colorize node labels in sidebar

OskarDamkjaer:
The full screen was a few pixels off:
bild

Fix:
bild

Since we have the colours of node labels readily available I thought it'd be neat if we used them when showing labels in the sidebar!
bild


HerrEmil (approved):
🧑‍🎨

PR #1206, opened 2020-10-02
Improve user statistics

OskarDamkjaer:
Adds better usage stats. Keeps track of where commands are run from (but not what commands). Most common non-sensitive settings etc.

### DIFF
--- a/src/browser/modules/ClickToCode/index.jsx
+++ b/src/browser/modules/ClickToCode/index.jsx
@@ -21,14 +21,17 @@ import React from 'react'
 import { withBus } from 'react-suber'
 import { SET_CONTENT, setContent } from 'shared/modules/editor/editorDuck'
 import { StyledCodeBlock } from './styled'
-import { executeCommand } from 'shared/modules/commands/commandsDuck'
+import {
+  executeCommand,
+  commandSources
+} from 'shared/modules/commands/commandsDuck'
 
 const setOnClick = (bus, code) => {
   code = Array.isArray(code) ? code.join('') : code
   bus.send(SET_CONTENT, setContent(code))
 }
 const execOnClick = (bus, code) => {
-  const cmd = executeCommand(code, { source: 'button-press' })
+  const cmd = executeCommand(code, { source: commandSources.button })
   bus.send(cmd.type, cmd)
 }
 
--- a/src/browser/modules/ClickToCode/index.jsx
+++ b/src/browser/modules/ClickToCode/index.jsx
@@ -28,7 +28,7 @@ const setOnClick = (bus, code) => {
   bus.send(SET_CONTENT, setContent(code))
 }
 const execOnClick = (bus, code) => {
-  const cmd = executeCommand(code)
+  const cmd = executeCommand(code, { source: 'button-press' })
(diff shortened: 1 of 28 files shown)


src/browser/modules/ClickToCode/index.jsx
```diff
@@ -28,7 +28,7 @@ const setOnClick = (bus, code) => {
   bus.send(SET_CONTENT, setContent(code))
 }
 const execOnClick = (bus, code) => {
-  const cmd = executeCommand(code)
+  const cmd = executeCommand(code, { source: 'button-press' })
```
OskarDamkjaer, line 31: Open to moving these to constants

src/shared/modules/udc/udcDuck.js
```diff
@@ -51,6 +67,7 @@ export const UPDATE_SETTINGS = `${NAME}/UPDATE_SETTINGS`
 export const UPDATE_DATA = `${NAME}/UPDATE_DATA`
 export const BOOTED = `${NAME}/BOOTED`
 export const METRICS_EVENT = `${NAME}/METRICS_EVENT`
+export const UPC_STARTUP = `${NAME}/STARTUP`
```
OskarDamkjaer: APP_START seems to be sent before the app.jsx starts listening for metric events. From what I can see in mixpanel and my testing we don't get any APP_START events.. Now app.jsx sends a UDC startup event when it's ready to listen

src/shared/modules/udc/udcDuck.js
```diff
@@ -247,6 +327,25 @@ export const bootEpic = (action$, store) => {
...
+  PIN,
+  UNPIN,
+  REMOVE,
+  EXPAND,
+  ADD_FAVORITE,
+  LOAD_FAVORITES,
+  UPDATE_FAVORITE,
+  REMOVE_FAVORITE,
+  UPDATE
+]
```
OskarDamkjaer, line 340: Moving to having more stats done like this would reduce maintenance burden but might generate a lot of events and not be as useful, since no data is sent.

OskarDamkjaer: We might not care about all of these either

src/shared/modules/udc/udcDuck.js
```diff
@@ -247,6 +331,25 @@ export const bootEpic = (action$, store) => {
...
+export const trackReduxActionsEpic = (action$, store) =>
+  action$
+    .filter(action => actionsOfInterest.includes(action.type))
```
HerrEmil, line 347: Might be a little easier to read as `.ofType(...actionsOfInterest)`

src/shared/modules/udc/udcDuck.js
HerrEmil: I like this way of tracking as action types are defined by us and could never contain user data and all setup is done in one place without cluttering any other code

HerrEmil (approved):

src/shared/modules/udc/udcDuck.js
OskarDamkjaer: Oh didn't know it took multiple arguments, neat!

PR #1213, opened 2020-10-12
Initial Monaco Editor implementation

HerrEmil:
This PR adds a new cypher editor, made with the Monaco editor. It also adds an `:edit` command that instantiates the new editor inside a frame in the stream so the new editor can be tested before it replaces the current editor.
edit frame and frame small

The new editor displays warnings from explain queries or from trying to use multi-statement queries as squiggly underlines with more info on hover.
explain query notifications
multi statement warning

The new editor tokenizes cypher as the current one and has syntax highlighting in dark and light mode.
Screenshot 2020-10-12 at 15 58 58
Screenshot 2020-10-12 at 15 59 13

Autocompleting cypher is not yet implemented in the moncao editor.


HerrEmil:
I will be adding tests while the PR is in review

oskarhane (approved):
This is a very good step, really cool to see it in action!
The only thing I don't really like is the additions to `FrameTitlebar`, but that whole architecture should be reconsidered at some point. 🐳

PR #1214, opened 2020-10-12
execute commands from sidebar and editor logic

jk05:
- While a project file is selected and is in the EditorFrame, executing a command via the sidebar will not clear the editor context - project file remains as it is. However then executing the command via the editor will clear the context. This brings it more in line with how Browser operates now and what makes more sense as an interim step.
- Small fix to logic previously whereby clicking a local script would enable edit pencil icon which would not clear when clicking onto a project file.

[Video showing changes](https://drive.google.com/file/d/1zuj84H9qFXFttPqb7VJZ2T-d2mr_dvAY/view)


OskarDamkjaer (approved):
LGTM!

PR #1218, opened 2020-10-15
add eslint-plugin-react-hooks

jk05:
- add [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks)
- had to change `useBrowserSync` to `utilizeBrowserSync` because of [this](https://github.com/facebook/react/issues/17636)

### DIFF
--- a/src/browser/modules/App/keyboardShortcuts.ts
+++ b/src/browser/modules/App/keyboardShortcuts.ts
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useCallback, useEffect, useMemo } from 'react'
 import { FOCUS, EXPAND } from 'shared/modules/editor/editorDuck'
 import { Bus } from 'suber'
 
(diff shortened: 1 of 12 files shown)


src/browser/modules/App/keyboardShortcuts.ts
```diff
@@ -59,26 +59,38 @@ function isOutsideTextArea(e: KeyboardEvent): boolean {
 }
 
 export function useKeyboardShortcuts(bus: Bus): void {
```
jk05, line 61: had to make several adjustments here, likely could be done in a better way.

src/browser/modules/App/keyboardShortcuts.ts
```diff
@@ -59,26 +59,38 @@ function isOutsideTextArea(e: KeyboardEvent): boolean {
...
+        e.preventDefault()
+        trigger(EXPAND)
+      }
+    },
+    [trigger]
+  )
 
-  const keyboardShortcuts = [focusEditorOnSlash, fullscreenEditor]
+  const keyboardShortcuts = useMemo(
+    () => [focusEditorOnSlash, fullscreenEditor],
```
OskarDamkjaer, line 91: I don't really understand what's going on here? Why do we return a list of useCallback hooks?

src/browser/modules/Editor/EditorFrame.tsx
```diff
@@ -105,11 +111,15 @@ export function EditorFrame({
   )
   const editorRef = useRef<CodeEditor>(null)
 
-  function toggleFullscreen() {
-    setFullscreen(!isFullscreen)
-  }
+  const toggleFullscreen = useCallback(() => setFullscreen(!isFullscreen), [
```
OskarDamkjaer, line 114: Went into the code to understand why we need this useCallback. But it seems we no longer need the isFullScreen for the useEffect() any more.

src/browser/modules/Editor/EditorFrame.tsx
```diff
@@ -105,11 +111,15 @@ export function EditorFrame({
...
+    isFullscreen
+  ])
 
-  useEffect(() => bus && bus.take(EXPAND, toggleFullscreen), [isFullscreen])
+  useEffect(() => bus && bus.take(EXPAND, toggleFullscreen), [
+    bus,
+    isFullscreen,
+    toggleFullscreen
```
OskarDamkjaer, line 121: So this no longer need isFullscreen

src/browser/modules/Editor/EditorFrame.tsx
jk05: To be honest I was just following the hints from this plugin to get some of these working, so do need looking at.

src/browser/modules/Editor/Monaco.tsx
```diff
@@ -64,6 +64,29 @@ const Monaco = ({
   const editorRef = useRef<editor.IStandaloneCodeEditor | null>(null)
   const monacoId = `monaco-${id}`
 
+  // Share current text with parent and add warnings
```
jk05, line 67: @HerrEmil wanted to get you to just check over this and see if it would still work ok?

src/browser/modules/Editor/EditorFrame.tsx
jk05: so remove `isFullScreen` in this file or all files?

src/browser/modules/Editor/EditorFrame.tsx
jk05: shall I remove it ?

src/browser/modules/Editor/EditorFrame.tsx
OskarDamkjaer: see the comment bellow

src/browser/modules/Editor/EditorFrame.tsx
OskarDamkjaer: Yeah line 120 is not needed in this dependency array since it has the toggleFullscreen there instead.

jk05: oh I see what you mean

src/browser/modules/Editor/Monaco.tsx
HerrEmil: I checked out and tried and no, these changes have caused an issue. The empty array was to only run the effect once, it creates the editor instance and a listener for text changes. Running that effect on each value change will destroy and recreate the editor instance on each value update. It is noticeable to the user as it will reset the theme and make you lose focus when typing.

HerrEmil (changes requested):

jk05:
@HerrEmil Thanks Ill take a look. The plugin was complaining in that file until those changes were made.

HerrEmil:
> @HerrEmil Thanks Ill take a look. The plugin was complaining in that file until those changes were made.

The effect that creates the Monaco instance uses the `value` prop to set the initial text content of the editor. I think it would work to remove `value` from that effect and create a separate effect for that prop and sets the value in monaco, but would need to check that there is not a loop. The other dependencies added will never change so listing them as dependencies should not be a problem.

HerrEmil (approved):

PR #1219, opened 2020-10-15
Do not store connection passwords in local storage before receiving neo4j config

HerrEmil:
Passwords are temporarily stored in local storage between connecting and receiving the config even if the `browser.retain_connection_credentials` neo4j config flag is false. To avoid storing passwords in localstorage unless the config flag is true, this PR changes three things:
 - The default value for the flag in redux store is now false
 - The store selector returns false if the flag is null or undefined
 - The local storage middleware checks the flag before saving to storage and replaces all passwords with an empty string unless the flag is true.

### DIFF
--- a/src/shared/services/localstorage.ts
+++ b/src/shared/services/localstorage.ts
@@ -18,13 +18,15 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { dehydrate } from 'services/duckUtils'
+import { Middleware } from 'redux'
+import { shouldRetainConnectionCredentials } from '../modules/dbMeta/dbMetaDuck'
+import { GlobalState } from '../modules/stream/streamDuck'
 
-export let keyPrefix = 'neo4j.'
+export const keyPrefix = 'neo4j.'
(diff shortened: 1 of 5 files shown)


src/shared/services/localstorage.ts
```diff
@@ -18,13 +18,15 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { dehydrate } from 'services/duckUtils'
```
jk05, line 21: is this not needed here now?

HerrEmil: The matching hydrate code that adds a flag on redux states is never called in current version, so it should be safe to never call dehydrate to remove the flag. As far as I can see, it is safe to remove the hydrate/dehydrate functions completely, they seem to only be used in their test cases. But I left them in this PR to not have a large diff to review.

jk05: 👍 I wasnt sure what it does really.

oskarhane (approved):
Well done, seems to work as expected 🏑

PR #1223, opened 2020-10-19
Only remove :auto if it is a prefix

### LINKED ISSUE #1149: Query not returning expected data after upgrade from 3.5.16 community to 3.5.19 enterprise
Hi there, I observed some unexpected query behaviors. Please see this [forum thread](https://community.neo4j.com/t/really-weird-query-not-returning-expected-data-after-upgrade-from-3-5-16-community-to-3-5-19-enterprise/21672) for full context.

> Neo4j version

3.5.19 Enterprise

> Operating system: (for example Windows 95/Ubuntu 16.04)

Ubuntu 14.04.6 LTS

> API/Driver: (for example Cypher/Java API/Python driver vX.Y.Z)

The script that fills my DB uses 1.7.6 of the Python driver

## Description

I recently upgraded my DB from 3.5.16 community to 3.5.19 with the following procedure:

1. `service stop neo4j`
1. Uninstall Neo4j Community 3.5.16, keeping the old data
1. `apt-get install neo4j-enterprise=3.5.19`
1. `service start neo4j`

This query works as expected: 
```
match (n:AWSTag) 
where n.key contains "aws:auto" 
return n.key, n.value 
order by n.key limit 300
``` 
as it returns a list of nodes with n.key = aws:autoscaling:groupName:

image

However when I just add an "s" to the key filter by running 
```
match (n:AWSTag)
where n.key contains "aws:autos"
return n.key, n.value
order by n.key limit 300
```
I get 0 results:

image

I have a second neo4j server that was also upgraded using the same steps, running similar data and it shows this exact problem as well.

Things I've tried:

- I have tried deleting all indexes on this :AWSTag node but that didn't work.
- I then tried turning off the server, deleting /var/lib/neo4j/data/databases/graph.db/schema/*, and turning it on again.
All other queries except for those related to `:AWSTags` work as expected (I think).

- PROFILE of the successful query:
    image

- PROFILE of the failing query: 
    image

Here are some other observations.

### Expected behavior

```
match (n:AWSTag) 
where n.key contains "aws:autos" 
return n.key, n.value 
order by n.key limit 300
```
should return a table that looks like

|n.key|n.value|
|-|-|
| aws:autoscaling:groupName | some group name | 
| aws:autoscaling:groupName | some other group name | 

### Actual behavior
The above query returns zero results.

### Expected

```
match (a:AWSTag)
where a.key starts with "aws:auto"
return distinct a.key
```
should return only keys that actually start with "aws:auto".

### Actual

The above query has these results:

image

### Expected
If I do an exact field search like 
```
match (a:AWSTag{key:"aws:autoscaling:groupName"}) return a
```

I should get back nodes with `key='aws:autoscaling:groupName`, as shown in the "Description" above.

### Actual 

I get no results.
image

## Additional information

- Partial output of `CALL db.stats.retrieve("GRAPH COUNTS")`

    ```json
    {
      "relationships": [
        {
          "count": 8882403
        },

        {
          "relationshipType": "TAGGED",
          "count": 97277,
          "endLabel": "AWSTag"
        },
        {
          "relationshipType": "TAGGED",
          "count": 97277,
          "endLabel": "Tag"
        },

      ],
      "nodes": [
        {
          "count": 2855520
        },

        {
          "count": 66181,
          "label": "AWSTag"
        },
        {
          "count": 66181,
          "label": "Tag"
        },

      ],
      "indexes": [

        {
          "updatesSinceEstimation": 2134,
          "totalSize": 64047,
          "properties": [
            "id"
          ],
          "labels": [
            "AWSTag"
          ],
          "estimatedUniqueSize": 64047
        },

      ],
      "constraints": []
    }
    ```

OskarDamkjaer:
Previously ":auto" would be removed from the middle of queries such as "match (n) where n.key = "test:auto". 

Resolves #1149

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
+++ b/src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
@@ -93,6 +93,7 @@ export class Visualization extends Component {
 
       this.getInternalRelationships(existingNodeIds, newNodeIds)
         .then(graph => {
+          console.log(graph.relationships)
           this.autoCompleteCallback &&
             this.autoCompleteCallback(graph.relationships)
         })
--- a/src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
+++ b/src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
@@ -93,7 +93,6 @@ export class Visualization extends Component {
 
       this.getInternalRelationships(existingNodeIds, newNodeIds)
         .then(graph => {
-          console.log(graph.relationships)
           this.autoCompleteCallback &&
             this.autoCompleteCallback(graph.relationships)
         })
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -385,7 +385,7 @@ const availableCommands = [
       const isAutocommit = query.startsWith(`:${autoCommitTxCommand}`)
 
       action.autoCommit = isAutocommit
-      action.query = isAutocommit ? query.slice(1) : query
+      action.query = isAutocommit ? query.slice(1).trim() : query
 
       const [id, request] = handleCypherCommand(
         action,
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -378,15 +378,14 @@ const availableCommands = [
     exec: (action, put, store) => {
       const state = store.getState()
 
-      // Since we now also handle queries with the :auto prefix, we need to strip that
-      // and attach to the actions object
-      const query = action.cmd.replace(`:${autoCommitTxCommand}`, '')
-      action.query = query.trim()
+      // Since we now also handle queries with the :auto prefix,
+      // we need to strip that and attach to the actions object
+      const query = action.cmd.trim()
 
-      // We need to find out if this is an auto-committing tx or not
-      // i.e. Did we strip something off above here?
-      const autoCommit = action.query.length < action.cmd.trim().length
-      action.autoCommit = autoCommit
+      const isAutocommit = query.startsWith(`:${autoCommitTxCommand}`)
+
+      action.autoCommit = isAutocommit
+      action.query = isAutocommit ? query.slice(1) : query
 
       const [id, request] = handleCypherCommand(
         action,
@@ -400,7 +399,7 @@ const availableCommands = [
           : getBackgroundTxMetadata({
               hasServerSupport: canSendTxMetadata(store.getState())
             }),
-        autoCommit
+        isAutocommit
       )
       put(cypher(action.cmd))
       put(
--- a/src/shared/services/commandInterpreterHelper.js
+++ b/src/shared/services/commandInterpreterHelper.js
@@ -380,20 +380,22 @@ const availableCommands = [
 
       // Since we now also handle queries with the :auto prefix,
       // we need to strip that and attach to the actions object
-      // note that comment removal is crude and will remove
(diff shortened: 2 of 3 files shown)


HerrEmil (approved):

src/shared/services/commandInterpreterHelper.js
```diff
@@ -384,18 +384,17 @@ const availableCommands = [
...
-      // Since we now also handle queries with the :auto prefix, we need to strip that
-      // and attach to the actions object
-      const query = action.cmd.replace(
-        getCmdChar(state) + autoCommitTxCommand,
-        ''
-      )
-      action.query = query.trim()
+      // Since we now also handle queries with the :auto prefix,
+      // we need to strip that and attach to the actions object
+      const query = action.cmd.trim()
```
oskarhane, line 389: This should strip any comments as well.

```
// Comment
:auto CREATE (n:Test {val: 'yo:auto'}) RETURN *
```
will now fail

OskarDamkjaer: Good catch!

OskarDamkjaer: I updated the implementation and added some tests

src/shared/services/commandInterpreterHelper.js
```diff
@@ -384,18 +384,22 @@ const availableCommands = [
...
+        .replace(/\/\/[^\n]*\n/g, '') // singleline comment
+        .trim()
+        .startsWith(autoPrefix)
 
-      // We need to find out if this is an auto-committing tx or not
-      // i.e. Did we strip something off above here?
-      const autoCommit = action.query.length < action.cmd.trim().length
-      action.autoCommit = autoCommit
+      action.autoCommit = isAutocommit
+      action.query = isAutocommit ? query.replace(autoPrefix, '').trim() : query
```
HerrEmil, line 402: I think replacing first instance of a text will fail if a comment has the text being replaced, e.g.
```
// browser should remove :auto in the query below this comment
:auto CREATE (n:Test {val: 'yo:auto'}) RETURN *
```

src/browser/modules/Stream/CypherFrame/VisualizationView.jsx
```diff
@@ -93,6 +93,7 @@ export class Visualization extends Component {
 
       this.getInternalRelationships(existingNodeIds, newNodeIds)
         .then(graph => {
+          console.log(graph.relationships)
```
HerrEmil, line 96: Leftover console.log

OskarDamkjaer: Thanks! It was for an unrelated change

src/shared/services/commandInterpreterHelper.js
OskarDamkjaer: Right. Thinking of how this is best solved without a proper parse.. I'll send an update, do tell what you think

OskarDamkjaer: Hmm... none of these handle the case of `:auto return 1; :auto return 1;`. Is that statement legal with multiline enabled?

OskarDamkjaer: Not sure what the best way forward (barring a proper parse) is. This query for example bypasses the easy-ish solutions I can think of.. `/* :auto */ :auto RETURN 1;`

HerrEmil: The issue I noted is already an issue in production, this fix does not cause any extra issue

HerrEmil (approved):

PR #1224, opened 2020-10-20
Couldn't leave fullscreen

OskarDamkjaer:
Toggle fullscreen was running with out of date state when not provided with the relevant state in the dep array.


oskarhane (approved):
Good find 👍

PR #1225, opened 2020-10-21
Re-add static scripts

OskarDamkjaer:
Add static scripts back in the favorite drawer


oskarhane (commented):
Thanks for taking this on!
What should happen when you edit a static script and save it?
Should saving be disabled?
Should it create a new local favorite?

Right now it looks like you're saving into the static script but the static script isn't changed.

oskarhane (approved):
LGTM 🌮

PR #1230, opened 2020-10-23
Setup sentry

OskarDamkjaer:
I've tested that no events get sent if the production env is not set and that it respects the allowOutGoingConnections setting. Would be good if someone else tested as well, just to be sure. To get the prod flag set correctly run `NODE_ENV=production yarn cross-env webpack-dev-server --colors --config ./build_scripts/webpack.config.js`. 20% of events get sent to the sentry dashboard. If you're a neo4j employee and don't have access ping me

### DIFF
--- a/src/browser/AppInit.jsx
+++ b/src/browser/AppInit.jsx
@@ -38,6 +38,10 @@ import { APP_START } from 'shared/modules/app/appDuck'
 import { GlobalStyle } from './styles/global-styles.js'
 import { detectRuntimeEnv } from 'services/utils.js'
 import { NEO4J_CLOUD_DOMAINS } from 'shared/modules/settings/settingsDuck.js'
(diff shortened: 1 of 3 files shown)


src/browser/AppInit.jsx
```diff
@@ -82,6 +86,26 @@ bus.applyMiddleware((_, origin) => (channel, message, source) => {
   store.dispatch({ ...message, type: channel, ...origin })
 })
 
+async function setupSentry() {
+  let isCanary = false
```
HerrEmil, line 90: The canary check can be skipped if we're not in production (i.e. move it inside the `if` block below)

OskarDamkjaer: True, I'll do so!

HerrEmil (approved):

PR #1233, opened 2020-10-26
add project file name input error check util

jk05:
- added a project file name input error util
- [Video](https://drive.google.com/file/d/1ezdkSOZyDTIVmweewlCRLyLLEq5x06y6/view)
- cleans it up for now and can add to the chars list as they show up. Or we could be better off with existing filename module 🤷

### DIFF
--- a/src/browser/modules/Sidebar/ProjectFiles.tsx
+++ b/src/browser/modules/Sidebar/ProjectFiles.tsx
@@ -28,7 +28,8 @@ import ProjectFilesScripts, { ProjectFilesError } from './ProjectsFilesScripts'
 import NewSavedScript from './NewSavedScript'
 import {
   setProjectFileDefaultFileName,
-  updateCacheAddProjectFile
+  updateCacheAddProjectFile,
+  checkFileNameInput
 } from './project-files.utils'
 import { CYPHER_FILE_EXTENSION } from 'shared/services/export-favorites'
 import { ADD_PROJECT_FILE } from './project-files.constants'
@@ -45,23 +46,12 @@ const ProjectFiles = ({ projectId, scriptDraft, resetDraft }: ProjectFiles) => {
   const [error, setError] = useState('')
 
   function save(inputedFileName: string) {
-    const fileName = inputedFileName.replace(/\.cypher$/, '')
+    const cypherFileExt = new RegExp(`${CYPHER_FILE_EXTENSION}$`)
+    const fileName = inputedFileName.replace(cypherFileExt, '')
     setError('')
 
-    if (!fileName.length) {
-      setError('File name cannot be empty')
-      return
-    }
-
-    // @todo: this needs more thought and extracting to a util
-    if (
-      fileName.includes('/') ||
-      fileName.includes('\\') ||
-      fileName.includes('..') ||
-      fileName.startsWith('.') ||
-      fileName.includes(':') // Windows
-    ) {
-      setError("File name cannot include /, \\, .., : or start with '.'")
+    if (checkFileNameInput(fileName)) {
+      setError(checkFileNameInput(fileName))
       return
     }
 
--- a/src/browser/modules/Sidebar/project-files.constants.ts
+++ b/src/browser/modules/Sidebar/project-files.constants.ts
@@ -33,6 +33,36 @@ export enum EXECUTE_COMMAND_ORIGINS {
   NONE = ''
 }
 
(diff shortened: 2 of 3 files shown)


src/browser/modules/Sidebar/ProjectFiles.tsx
```diff
@@ -45,23 +46,12 @@ const ProjectFiles = ({ projectId, scriptDraft, resetDraft }: ProjectFiles) => {
   const [error, setError] = useState('')
 
   function save(inputedFileName: string) {
```
jk05, line 48: I avoided making any more changes to this save func as I think additional work will be occurring here in the next phase potentially, so best to leave it then.

src/browser/modules/Sidebar/project-files.utils.ts
```diff
@@ -119,6 +122,55 @@ export const createFilePath = (paths: string[]): string => {
...
+        }
+      }
+    }
+  })
+
+  return errors.chars.length || errors.tests.length
+    ? `File name cannot contain ${
+        errors.chars.length ? `${errors.chars.join(', ')}` : ''
+      }${errors.chars.length && errors.tests.length ? ' or ' : ''}${
+        errors.tests.length ? `${errors.tests.join(', ')}` : ''
```
jk05, line 169: was going to present regex tests in a different way but couldnt think of anything for now, so left it like this

src/browser/modules/Sidebar/project-files.constants.ts
```diff
@@ -33,6 +33,36 @@ export enum EXECUTE_COMMAND_ORIGINS {
...
+
+export const PROHIBITED_FILENAME_CHAR_TESTS: (
+  | ProhibitedFilenameChar
+  | ProhibitedFilenameTest
+)[] = [
+  { char: '/' },
+  { char: '\\' },
+  { char: '..' },
+  { test: /^[.|?]/ },
+  { char: ':', platform: 'Win32' }
```
OskarDamkjaer, line 63: does it not make send to remove them for all platforms just to avoid confusion?

src/browser/modules/Sidebar/project-files.utils.ts
```diff
@@ -119,6 +122,55 @@ export const createFilePath = (paths: string[]): string => {
...
 
+export const checkFileNameInput = (fileName: string): string => {
+  if (!fileName.length) {
+    return 'File name cannot be empty'
+  }
+
+  const errors: ProhibitedFilenameErrors = { chars: [], tests: [] }
+
+  PROHIBITED_FILENAME_CHAR_TESTS.forEach(charObj => {
+    if (charObj.test) {
```
OskarDamkjaer, line 133: Not strictly needed but would it not be easier if they were all regex tests?

src/browser/modules/Sidebar/project-files.constants.ts
jk05: yeah probably

src/browser/modules/Sidebar/project-files.utils.ts
jk05: yeah that makes sense actually

OskarDamkjaer:
Sorry bout the late-ish review!

Perhaps we could simplify this by using the built in [pattern in input](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern)? We could add the forbidden char patterns in the pattern attribute, and then set the title (which shows when you try to submit if you don't match the pattern) dynamically to be the error message you created on line 167 in the project utils file!

jk05:
> Sorry bout the late-ish review!
> 
> Perhaps we could simplify this by using the built in [pattern in input](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern)? We could add the forbidden char patterns in the pattern attribute, and then set the title (which shows when you try to submit if you don't match the pattern) dynamically to be the error message you created on line 167 in the project utils file!

So one whole regex in the pattern attribute for all platforms, and then use the util to create a message for the `title` attribute? There seems to be [additional info](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefpattern) in terms of accessibility and not using `title` attr alone, so might be more useful to display the error message as is currently too.

Another way is to do what you suggested and do away with the util entirely and just display a string of the regex exclusions - simpler but less user friendly.

Open to suggestions

OskarDamkjaer (approved):

PR #1235, opened 2020-10-28
Update connection form to better reflect aura usage

OskarDamkjaer:
To test add "localhost" to the NEO4J_CLOUD_DOMAINS in settings duck. I'm unsure how to properly mock the discovery though. I've also tested connecting with the form against a proper aura db.

Form is simplified to show the following:
bild

Unsure if we want to keep the dropdown when we only have one option. I've also considered adding a pattern to the form to require the connection url match the format of the required scheme. Thoughts? 

Also. Aura calls it "Connection URI" and we call it a "Connect URL". Should either change to be more consistent?

### DIFF
--- a/src/browser/modules/Stream/Auth/ConnectionForm.jsx
+++ b/src/browser/modules/Stream/Auth/ConnectionForm.jsx
@@ -46,7 +46,11 @@ import { NATIVE, NO_AUTH } from 'services/bolt/boltHelpers'
 import ConnectForm from './ConnectForm'
 import ConnectedView from './ConnectedView'
 import ChangePasswordForm from './ChangePasswordForm'
-import { getAllowedBoltSchemes, inCloudEnv } from 'shared/modules/app/appDuck'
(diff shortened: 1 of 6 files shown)


src/browser/modules/Stream/Auth/ConnectionForm.jsx
```diff
@@ -315,6 +315,9 @@ export class ConnectionForm extends Component {
           username={this.state.username}
           password={this.state.password}
           database={this.state.requestedUseDb}
+          allowedAuthMethods={
+            this.props.inCloudEnv ? [NATIVE] : [NATIVE, NO_AUTH]
+          }
```
oskarhane, line 320: To be consistent, would you consider moving this to `mapStateToProps`, next to `allowedSchemes`?

OskarDamkjaer: Sure!

PR #1236, opened 2020-10-28
Disable test for feature not in aura

OskarDamkjaer:
Aura doesn't support an arbitrary amount of databases, so we've disabled the choose database in connect form in aura but forgot to add a check in the tests.


oskarhane (approved):
LGTM
Some edition of future Aura might support multi-db though, but we'll tackle that when that happens.

PR #1238, opened 2020-10-30
Fix issue where users couldn't be added when on a cluster

OskarDamkjaer:
Also included a fix for logging into dbs without auth with blank username/pw


src/shared/modules/cypher/cypherDuck.js
```diff
@@ -42,6 +42,7 @@ import {
 
 const NAME = 'cypher'
 export const CYPHER_REQUEST = `${NAME}/REQUEST`
+export const ROUTED_CYPHER_REQUEST = `${NAME}/ROUTED_REQUEST`
```
oskarhane, line 45: Can you rename this to reflect that it's write transactions that are being performed @OskarDamkjaer ?

OskarDamkjaer: Done!

oskarhane (approved):
LGTM 🕺

PR #1239, opened 2020-11-02
Update "create index" examples to use new syntax

OskarDamkjaer:
Example in :help create index-on and static script now includes modern syntax.


oskarhane (approved):
LGTM 👍

PR #1243, opened 2020-11-24
Basic performance overlay

OskarDamkjaer:
First step with performance instrument, more metrics and refinement to come. To turn the overlay on run `:config "showPerformanceOverlay": true`

Keep in mind memory usage is only available in chrome sadly


oskarhane (commented):
Should we run measurements when the overlay isn't activated?
That seems a bit unnecessary to me.

oskarhane:
btw, the preview is really useful for PR:s like these 👍

OskarDamkjaer:
Yeah! I'm pretty happy with it! It could perhaps re-run on updated commits though

oskarhane (approved):
💈

PR #1244, opened 2020-11-25
Use Relate ID instead of Desktop ID for projects

nglgzz:
Everything mentioned here only affects the saved files feature.

Due to a change in how Desktop is retrieving projects, using the Desktop ID to get a project through Relate will no longer work. To avoid breaking this feature, and to be more semantically correct, Desktop will be passing the Relate ID instead when launching graph apps. 

This PR makes Browser aware of the new URL parameter if it exists and falls back to the old one if it doesn't, making the change backward compatible.

More information about the Desktop changes can be found here: https://github.com/neo-technology/neo4j-desktop/pull/927

### DIFF
--- a/src/browser/AppInit.tsx
+++ b/src/browser/AppInit.tsx
@@ -121,7 +121,9 @@ const searchParams = new URL(url).searchParams
 // Desktop/Relate params
 const relateUrl = searchParams.get('relateUrl')
 const relateApiToken = searchParams.get('relateApiToken')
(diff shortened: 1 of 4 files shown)


nglgzz:
[[Kick-off] Link DBMSs to projects with Relate in Desktop](https://trello.com/c/fIXpxVW1/1498-kick-off-link-dbmss-to-projects-with-relate-in-desktop)

jk05 (approved):

PR #1245, opened 2020-11-27
Removed information about creating a new db in :play movies

shkirando:
Because multi-db is not supported in all Browser use cases (Aura, 3.5.X)


OskarDamkjaer:
LGTM!

Issue #447, opened 2017-04-29
Missing zoom buttons in fullscreen graph view 


naisanza:
In the packaged Neo4j database + browser, there is a zoom feature when the graph is fullscreened.

Is this feature removed in this branch?

naisanza:
Closing. The zoom buttons appears only when required for the view

PR #1246, opened 2020-11-27
Allow zoom buttons on small frames (with settings config)

ivnnv:
It seems that by historical reasons (#447) it is intentional to hide the zoom button on non-fullscreen frames., but I think nowadays we have big enough screen to have them visible (in the case the user wants, hence adding a settings option defaulting to false, just as the current behaviour).

So this PR is adding a config checkbox option to have the zoom buttons available and adapts those buttons to make them smaller accordingly.

zoom small frame

PS: CLA sign sent

### DIFF
--- a/src/browser/components/icons/Icons.tsx
+++ b/src/browser/components/icons/Icons.tsx
@@ -244,17 +244,19 @@ export const ErrorIcon = () => (
   <IconContainer className="fa fa-file-text-o" text="Error" />
 )
 
-export const ZoomInIcon = () => (
+export const ZoomInIcon = (props: any) => (
   <IconContainer
     activeStyle={inactive}
     inactiveStyle={inactive}
+    {...props}
     className="sl-zoom-in"
   />
 )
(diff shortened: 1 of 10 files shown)


OskarDamkjaer:
Thanks for your PR and for signing the CLA! 

I agree we should show the zoom controls more often and I don't even think we'd need to hide it behind a setting. It'd be reasonable to always show them.

Could you keep only the icon size change and remove the setting and the `if (this.props.fullscreen || this.props.zoomOnSmallFrames) {` check?

OskarDamkjaer:
@ivnnv Also, I added you to the allow-list so the CI should run your tests properly now.

ivnnv:
@OskarDamkjaer friendly ping, I reworked the PR to remove the config setting and now all tests have passed and everything seems ready. Excited to have my first contribution merged!

OskarDamkjaer:
@ivnnv Looks great, thanks!

PR #1254, opened 2020-12-14
Add more detail in cypher stats

OskarDamkjaer:
Now keeps track of words, lines and statement count in cypher queries.


HerrEmil (approved):
🍰

PR #1255, opened 2020-12-14
Sentry updates

OskarDamkjaer:
Changes:
- Reuse mixpanel id to track how many users unique users run into issues
- Sends console.errors to sentry, not just uncaught errors
- Use sentry's errorboundry so see errors caught by errorboundries as well
- Seperates prod/canary env properly
- Sets up a sentry "transaction", to track how slow cypher queries generally are

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,6 +6,8 @@
     "target": "ES6",
     "strict": true,
     "sourceMap": true,
+    "inlineSources": true,
+    "sourceRoot": "/",
     "noImplicitAny": true,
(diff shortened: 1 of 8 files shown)


tsconfig.json
```diff
@@ -6,6 +6,8 @@
     "target": "ES6",
     "strict": true,
     "sourceMap": true,
+    "inlineSources": true,
```
HerrEmil, line 9: remove

HerrEmil (approved):

PR #1257, opened 2020-12-17
Remove unneeded mixpanel events

OskarDamkjaer:
When switching plans from monthly active users to number of events. Since we have cypher succeeded and failed, we don't need a separate event for totals.


OskarDamkjaer:
Will need to give the tests another look!

HerrEmil (approved):

PR #1264, opened 2021-01-18
Saved script refatoring

OskarDamkjaer:
These changes aim to make nested folders, custom ordering with drag and drop, bulk renaming/exporting/deleting easier to implement as well as make the code easier to understand for new developers and decrease the surface area for bugs. Should preserve all behaviour, except this allows for empty folders to exist. Export is also missing since it's getting reworked in the next PR anyway. 

### Changelist:
- Add proper types, remove ts-ignores and disabled nullchecks
- Move to hook based drag n drop API
- Make duplicated buttons into separate component
- Update file naming to better match our naming standard
- Updated the data structures, details bellow
- Pulled out project files component to allow for lazy loading of the script contents and simpler apis 

In redux and firebase the favorite state is shaped like this (somewhat simplified):
`Favorite: {content: string, id: string, folderId: string}`
`Folder: {name: string, id: string}` 

and is then transformed to only
`Favorite: {content: string, id: string, path: string}` 
Where the folders are all encoded in the path of the favorite. This adds a lot of complexity as:
- Empty folders need special handling in the UI (as they can only exist in the first representation)
- It's harder to add nested folders in the future (instead of just adding a parentFolderId to folders we need to parse and manipulate all nested folders/favorites paths)
- a lot of util functions exist only to handle the transformations 
- it's harder to understand (at least I needed to work with favorites with quite a while until I understood the data flows properly)

Preview available at http://saved_script_refatoring.surge.sh.


src/browser/modules/Sidebar/ProjectsFilesScripts.tsx
```diff
@@ -206,14 +206,12 @@ function ProjectFilesScripts(props: ProjectFilesScripts): JSX.Element {
...
-    },
-    onExportScripts: Function.prototype,
-    onUpdateFolder: Function.prototype,
-    onRemoveFolder: Function.prototype
+    }
   }
 
   return (
     <>
+      {/* @ts-ignore */}
```
HerrEmil, line 214: what is being ignored?

src/shared/services/export-favorites/export-favorites.utils.ts
```diff
@@ -46,24 +43,25 @@ import { SLASH, CYPHER_FILE_EXTENSION } from './export-favorites.constants'
...
 export function mapOldFavoritesAndFolders(
-  favorites: any,
-  folders: any,
-  isAllowed = isNonStatic
+  favorites: Favorite[],
+  folders: Folder[],
+  isAllowed: (fav: Favorite) => boolean = isNonStatic
 ) {
   const oldFoldersMap = new Map(
+    // @ts-ignore
```
HerrEmil, line 51: what is being ignored?

HerrEmil (approved):
I'm loving these changes :)

The space between the Local Scripts and Sample Scripts headline has decreased when there are no Local Scripts, is this an intended design change?

src/browser/modules/Sidebar/ProjectsFilesScripts.tsx
OskarDamkjaer: I've not yet fixed the project files to work with the new changes, so the ignore is there since they're broken! Will ping you for re-review when it's fixed O:)

src/shared/services/export-favorites/export-favorites.utils.ts
OskarDamkjaer: Since we'll be changing how export works I didn't spend any time fixing it for now!

src/browser/modules/Sidebar/ProjectsFilesScripts.tsx
OskarDamkjaer: I've restructured how they work now, do try them out! (to do so add the preview-url as a graph app) I've speed up the loading of them since before we preloaded all the scripts just to list the names. And I moved them out to separate components, since they are different from favorites.

HerrEmil (approved):

build_scripts/webpack-plugins.js
```diff
@@ -84,15 +84,15 @@ module.exports = () => {
       openAnalyzer: false,
       reportFilename: './../bundle-report.html'
     }),
-    new ForkTsCheckerWebpackPlugin({
```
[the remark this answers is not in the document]
OskarDamkjaer: Thanks! I'll make sure to remove it! Found a better way to solve it in another branch O:)

PR #1266, opened 2021-01-19
Aura version fixes

OskarDamkjaer:
No preview link as they need to run on aura to notice the difference. To test, hard-code the getEnv in appDuck to CLOUD.

Changes:
- Disables test that uses bolt as it's not available on aura.
- Guesses the server version when on aura.
- Disables user management frames on aura until we have a way to differentiate Professional and Enterprise on aura. See pictures for fallback.
CleanShot 2021-01-19 at 12 33 31@2x


e2e_tests/integration/connect-form.spec.ts
```diff
@@ -44,16 +44,19 @@ describe('Connect form', () => {
...
-    const scheme = schemeWithEncryptionFlag('bolt')
-    const host = 'localhost:1212'
-    getBoltUrlField()
-      .clear()
-      .type(scheme + host)
-
-    getBoltUrlField().should('have.value', host)
-    getBoltSchemeSelect().should('have.value', scheme)
-  })
+  if (!isAura) {
```
oskarhane, line 47: Should be function call

src/browser/modules/User/UserManagementOnAura.tsx
```diff
@@ -0,0 +1,22 @@
...
+          href="https://neo4j.com/docs/cypher-manual/current/administration/security/users-and-roles"
+          target="_blank"
+          rel="noreferrer"
+        >
+          the Neo4j Cypher docs.
+        </a>
+      </p>
+    </div>
+  )
+}
```
oskarhane, line 22: Looks like you're not using this new component?

oskarhane (commented):
Looks good, but looks like you refactored but didn't use the new component :)

e2e_tests/integration/connect-form.spec.ts
OskarDamkjaer: Oh right, thanks!

src/browser/modules/User/UserManagementOnAura.tsx
OskarDamkjaer: Yeah I made that one first but decided not to use it

oskarhane (approved):
LGTM 🦩

PR #1268, opened 2021-01-21
Make commands editable and re-usable

OskarDamkjaer:
This pull request aims to make it possible to edit and re-run frames. It's added as an experimental setting for now, but defaults to on for ease of testing.

I've annotated the source code to explain what's changed and why. Apologies for the size of the PR, I recommend setting the "side by side" diff ignoring whitespace to get a better view of the bigger changes. 

Demo: https://user-images.githubusercontent.com/10564538/105364088-098cad80-5bfd-11eb-9869-d606b59699a6.mp4

Try it yourself at: http://reusable-frame.surge.sh/


src/browser/modules/Frame/FrameTemplate.tsx
```diff
@@ -30,137 +31,135 @@ import {
...
-    )
-  }
+  const {
+    isFullscreen,
+    isCollapsed,
+    isPinned,
+    toggleFullScreen,
+    toggleCollapse,
+    togglePin
+  } = useSizeToggles()
```
OskarDamkjaer, line 73: extracted a hook to make the component more readable

src/browser/modules/Frame/FrameTitlebar.tsx
```diff
@@ -77,13 +78,78 @@ import {
...
+  To regain editor focus we run an effect dependant on the isRerun prop.
+  However, when the frame prop changes in some way the effect is retriggered
+  although the "isRun" is still true. Use effect does not check for equality
+  but instead re-runs the effect to take focus again. To prevent this
+  we use the useCallback hook as well. As a best effort we set the cursor position
+  to be at the end of the query.
+
+  A better solution is to change the frame titlebar to reside outside of the 
+  frame contents.
+  */
```
OskarDamkjaer, line 132: This is something I hope to be able to address in a futures PR

src/browser/modules/Frame/styled.tsx
```diff
@@ -18,47 +18,26 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
OskarDamkjaer, line 18: Mostly added proper typings to these components so the ignore-comments could be removed.

src/browser/modules/Stream/CypherFrame/index.tsx
```diff
@@ -392,7 +409,7 @@ export class CypherFrame extends Component<any, CypherFrameState> {
 
     return (
       <FrameTemplate
-        sidebar={requestStatus !== 'error' ? this.sidebar : null}
+        sidebar={requestStatus !== 'error' ? this.sidebar : undefined}
```
OskarDamkjaer, line 412: This fixed a type error

src/browser/modules/Stream/PlayFrame.tsx
```diff
@@ -76,11 +77,18 @@ export function PlayFrame({ stack, bus }: any) {
         onSlide,
         shouldUseSlidePointer
       )
-      setInitialPlay(false)
-      setGuideObj({ guide, aside, hasCarousel, isRemote })
+      if (stillMounted) {
+        setInitialPlay(false)
+        setGuideObj({ guide, aside, hasCarousel, isRemote })
```
OskarDamkjaer, line 82: Prevents setState on unmounted component if frame changes.

src/shared/modules/commands/commandsDuck.ts
```diff
@@ -200,12 +200,13 @@ export const handleCommandEpic = (action$: any, store: any) =>
         // Single command
         return store.dispatch(executeSingleCommand(action.cmd, action))
       }
-      const parentId = action.parentId || v4()
+      const parentId = (action.isRerun ? action.id : action.parentId) || v4()
```
OskarDamkjaer, line 203: Keep the id for re-run actions. New Id gives new frame

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -452,14 +452,14 @@ const switchToRequestedDb = (store: any) => {
   const activeConnection = getActiveConnectionData(store.getState())
   const requestedUseDb = activeConnection?.requestedUseDb
 
-  const useDefaultDb = () => {
+  const switchToDefaultDb = () => {
```
OskarDamkjaer, line 455: renamed because of react hooks lint rule

src/shared/modules/requests/requestsDuck.ts
```diff
@@ -23,25 +23,49 @@ import bolt from 'services/bolt/bolt'
...
 
 export const NAME = 'requests'
-export const REQUEST_SENT = `${NAME}/SENT`
-export const CANCEL_REQUEST = `${NAME}/CANCEL`
-export const REQUEST_CANCELED = `${NAME}/CANCELED`
-export const REQUEST_UPDATED = `${NAME}/UPDATED`
+export const REQUEST_SENT = 'requests/SENT'
+export const CANCEL_REQUEST = 'requests/CANCEL'
+export const REQUEST_CANCELED = 'requests/CANCELED'
+export const REQUEST_UPDATED = 'requests/UPDATED'
```
OskarDamkjaer, line 29: Typescript struggles to correctly type the actions if they´re not simple strings

src/shared/modules/udc/udcDuck.ts
```diff
@@ -45,9 +45,7 @@ import {
 } from 'shared/modules/favorites/favoritesDuck'
 import {
   shouldReportUdc,
-  getSettings,
-  REPLACE,
-  UPDATE
+  getSettings
```
OskarDamkjaer, line 48: A few unused imports

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -218,23 +218,28 @@ const availableCommands = [
           (db: any) => db.name.toLowerCase() === cleanDbName
         )
 
-        function UseDbError({ code, message }: any) {
-          // @ts-expect-error ts-migrate(2683) FIXME: 'this' implicitly has type 'any' because it does n... Remove this comment to see the full error message
-          this.code = code
-          // @ts-expect-error ts-migrate(2683) FIXME: 'this' implicitly has type 'any' because it does n... Remove this comment to see the full error message
-          this.message = message
+        class UseDbError extends Error {
```
OskarDamkjaer, line 221: Threw together an a better Error to avoid ts-errors

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -463,28 +468,29 @@ const availableCommands = [
...
-      } else if (response) {
-        put(
-          frames.add({
-            useDb: getUseDb(store.getState()),
-            ...action,
-            ...response
+      if (response) {
+        if (response.then) {
+          response.then((res: any) => {
+            if (res) {
```
OskarDamkjaer, line 474: Moved the conditions in a way I found easier to read

e2e_tests/integration/topics.spec.ts
```diff
@@ -31,9 +31,7 @@ describe('Help topics', () => {
     cy.executeCommand(':clear')
     const query = ':help commands'
     cy.executeCommand(query)
-    cy.get('[data-testid="frameCommand"]', { timeout: 10000 })
-      .first()
-      .should('contain', query)
+    cy.get('[data-testid="frameCommand"]', { timeout: 10000 }).contains(query)
```
oskarhane, line 34: Not sure I like these `.should('contain', x)` -> `contains(x)` changes. 
The former is an explicit assertion where the second returns the element so we can chain other operations on it.
Both will fail if the element isn't there, but the explicit assertion is a more explicit failure if I remember correctly.

src/browser/modules/Editor/Monaco.tsx
```diff
@@ -73,13 +73,14 @@ interface MonacoProps {
...
-  history?: HistoryState
+  history?: string[]
   id: string
   value?: string
   onChange?: (value: string) => void
   onDisplayHelpKeys?: () => void
-  onExecute?: () => void
+  onExecute?: (value: string) => void
   useDb?: null | string
+  toggleFullscreen?: () => void
```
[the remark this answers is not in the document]
oskarhane: While I really like the feature of fullscrrening every editor by itself, the fullscreen doesn't look that good, see attachment.
Screen Shot 2021-01-22 at 09 16 39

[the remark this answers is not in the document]
oskarhane: I also ran into a situation where I had two editors open at the same time. 
TRP
1. Click in editor in result frame
2. Press Esc
3. When in full screen, click outside of the edit, in the result itself
4. Press Esc
5. Close fullscreen on the frame (not all frame seems to have this button!)
6. You're now in fullscreen of the main editor

I think needs a proper state machine to be confident what's going on.

oskarhane (commented):
Wow, I like this feature more than I thought!
The user experience is actually really good, it's no big deal to just click in a query in the stream and edit it.
Well done!

I've left some comments but the main issue is around fullscreen mode.

src/browser/modules/Editor/Monaco.tsx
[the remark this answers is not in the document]
OskarDamkjaer: I've considered only full screening the main editor while it has input focus as well. That behaviour I think would be consistent and and simple to implement (we'd only have to remove code really)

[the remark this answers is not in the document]
OskarDamkjaer: Thoughts? @oskarhane

e2e_tests/integration/topics.spec.ts
OskarDamkjaer: I agree the should contain is clearer as an assertion. The reason I changed them as because the dom structure was changed and .contains() looks through all grandchildren rather than just its own. So I use .contains() as it accomplishes the same thing but was easier to write. I could change them back if you'd like

oskarhane: I didn't know that `.contains` checks deeper. No need to change then.

src/browser/modules/Editor/Monaco.tsx
```diff
@@ -446,7 +455,7 @@ const Monaco = forwardRef<MonacoHandles, MonacoProps>(
       })
     }
 
-    return <MonacoStyleWrapper id={monacoId} />
+    return <MonacoStyleWrapper data-testid="monaco-editors" id={monacoId} />
```
HerrEmil, line 458: Since the components already have IDs that start with the same string, the mutual test id is not needed

src/browser/modules/Stream/PlayFrame.tsx
```diff
@@ -76,11 +77,18 @@ export function PlayFrame({ stack, bus }: any) {
...
+      }
     }
     generate()
-  }, [currentFrame])
+
+    return () => {
+      stillMounted = false
+    }
+    // The full dependency array causes a re-run witch switches to slide 1
```
HerrEmil, line 90: witch -> which

HerrEmil (commented):
Really cool feature :)

The gutter width does not update immediately when the useDb changes, so it's often cut off when starting to edit in a frame title bar for me
Screenshot 2021-01-28 at 17 41 58

src/browser/modules/Editor/Monaco.tsx
OskarDamkjaer: Cypress doesn't seem to have an easy way to do get with a regex, so I'm thinking this'd be worth keeping for that reason.

e2e_tests/integration/params.spec.ts
```diff
@@ -64,7 +64,7 @@ function runTests() {
   // it(':param x => {prop: 1} multi line', () => {
   // Set param
   cy.executeCommand(':clear')
-  setParamQ = `:param [x] => {{}{shift}{enter}RETURN {{}prop: 1} AS x`
+  setParamQ = `:param [x] => {{}{shift}{enter}RETURN {{}prop: 1} AS x{enter}`
```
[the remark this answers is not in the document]
OskarDamkjaer: Removed!

e2e_tests/support/global.d.ts
```diff
@@ -24,6 +24,11 @@ declare global {
         boltUrl?: string,
         force?: boolean
       ): Cypress.Chainable<void>
+      /**
+       * Custom command to type a command in a frame
+       */
+      (): Cypress.Chainable<void>
```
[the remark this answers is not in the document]
OskarDamkjaer: Fixed!

e2e_tests/support/commands.ts
```diff
@@ -102,6 +100,17 @@ Cypress.Commands.add('disconnect', () => {
   cy.executeCommand(query)
 })
 
+Cypress.Commands.add('typeInFrame', (cmd: string, frameIndex = 0) => {
+  cy.get('[data-testid=monaco-editors]')
+    .eq(frameIndex + 1) // the first monaco editor is the main one
+    .type(
+      Cypress.platform === 'darwin'
+        ? '{cmd}a {backspace}{backspace}'
```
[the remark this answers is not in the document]
OskarDamkjaer: Removed!

e2e_tests/support/commands.ts
```diff
@@ -102,6 +100,17 @@ Cypress.Commands.add('disconnect', () => {
   cy.executeCommand(query)
 })
 
+Cypress.Commands.add('typeInFrame', (cmd: string, frameIndex = 0) => {
+  cy.get('[data-testid=monaco-editors]')
```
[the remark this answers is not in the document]
OskarDamkjaer: Right! Fixed

e2e_tests/integration/0.index.spec.ts
```diff
@@ -106,9 +98,8 @@ describe('Neo4j Browser', () => {
       .click()
     cy.get(SubmitQueryButton).click()
     cy.waitForCommandResult()
-    cy.get('[data-testid="frameCommand"]', { timeout: 10000 })
-      .first()
-      .should('contain', 'Emil Eifrem')
+    //cy.get('[data-testid="frameCommand"]', { timeout: 10000 }).contains( 'Tom Hanks')
```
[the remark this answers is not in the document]
OskarDamkjaer: Updated!

src/browser/modules/Editor/Monaco.tsx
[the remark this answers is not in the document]
OskarDamkjaer: Waiting with the looks but fixed the double fullscreen issue.

OskarDamkjaer:
I've gone through your comments, tests pass locally and the preview is updated (http://reusable-frame.surge.sh/). Awaiting more feedback/green pipe :)

HerrEmil (approved):
🎉

oskarhane (commented):
Not all frames seems to have the "exit fullscreen" button, which would be needed now that all frame can go into fullscreen.

Example: 
1. On `:play start` that shows up, click the frame editor.
2. Press Esc to go into fullscreen on that frame
3. Click on an input field on the frame
4. Press Esc. 
5. Nothing happens and there's no obvious way for the user how to get out of the fullscreen state (click editor + press Esc is the only way out)

OskarDamkjaer:
Oh right forgot about that, I've removed the conditional render of it now! @oskarhane

src/browser/modules/Stream/Stream.tsx
```diff
@@ -78,65 +80,70 @@ const getFrame = (type: any) => {
...
-            stack: frameObject.stack
-          }
-          let MyFrame = getFrame(frame.type)
-          if (frame.type === 'error') {
-            try {
-              const cmd = frame.cmd.replace(/^:/, '')
-              const Frame = cmd[0].toUpperCase() + cmd.slice(1) + 'Frame'
-              MyFrame = require('./Extras/index')[Frame]
-              if (!MyFrame) {
-                MyFrame = getFrame(frame.type)
```
oskarhane, line 121: This was written for us to sneak in fun frames without having to explicitly import them. I think we should keep it :)

src/shared/modules/experimentalFeatures/experimentalFeaturesDuck.ts
```diff
@@ -9,13 +9,20 @@ export const showFeature = (state: any, name: any) =>
...
     name: experimentalFeatureSelfName,
     on: true,
     displayName: 'Show experimental features',
     tooltip: 'Show feature section in settings drawer'
+  },
+  [reusableFrame]: {
+    name: reusableFrame,
+    on: true,
+    displayName: 'Enable reuseable frame',
+    tooltip: 'Edit and rerun right in a frame'
```
oskarhane, line 25: 👍

src/browser/modules/Stream/styled.tsx
```diff
@@ -370,3 +370,17 @@ export const StyledHistoryRow = styled.li`
...
+    max-height: 0;
+  }
+  to {
+    transform: translateY(0);
+    max-height: 500px; /* Greater than a frame can be */
+  }
+`
+export const AnimationContainer = styled.div`
+  animation: ${rollDownAnimation} 0.4s ease-in;
+`
```
oskarhane, line 386: 👍

oskarhane (commented):
Executing:

```
RETURN 1;
RETURN "hej";
```
kills the stream (at least when I'm not connected to a DBMS)

src/browser/modules/Stream/Stream.tsx
OskarDamkjaer: It's always easier to take fun out than put it in, I'll keep it as is then :)

OskarDamkjaer:
Good that you caught the multi statement issue! Should be resolved now.

oskarhane (approved):
Let's do this 🎞️

PR #1269, opened 2021-01-21
Refresh bug

OskarDamkjaer:
In the discoveryOnStartupEpic from discoveryDuck.ts we update the discovery connection to the discovered host regardless of previous connection data. 

The fix for this is to only ping the discovery when the saved connection data doesn't work. We'd still want connectURL params to be respected regardless. The solution I've put in place is for discovery to pass it's findings to the next Epic instead of writing directly to the connection state.

Reproduction steps:
- Login to browser running against neo4j 4.X (with discovery endpoint available) with bolt:// 
- Refresh to see neo4j:// is used instead of bolt

Currently working to fix tests and make sure we get the expected behaviour


.eslintrc.json
```diff
@@ -56,23 +56,22 @@
         "react"
       ],
       "rules": {
-        "@typescript-eslint/no-empty-function": "off",
-        "@typescript-eslint/ban-types": "off",
```
OskarDamkjaer, line 60: With the warnings disabled in the console, turning these on to want is fine.

HerrEmil (approved):

PR #1274, opened 2021-02-04
Improve performance when receiving large query results

HerrEmil:
After reading [Dan Flavin's post on waiting for large query results in browser](https://medium.com/neo4j/wheres-my-neo4j-cypher-query-results-%EF%B8%8F-%EF%B8%8F-9c3b150e6e19), I did some performance testing of browser. I found the loop that builds the graph visualisation to be unoptimised and reworked it which lead to a noticeable performance improvement. The first two commits in the branch are the performance improvements.

While testing large query results in a dev build, redux devtools crashed a few times due to out of memory errors. The third commit is a change that omits query results from redux devtools.

The 4th, 5th and 6th commits are code cleanups. They touch many lines so I recommend reviewing commit by commit to not miss other changes. Note that the typing commit does contain some changes to prevent runtime type errors that typescript found with the added type information.

### DIFF
--- a/src/browser/AppInit.tsx
+++ b/src/browser/AppInit.tsx
@@ -20,7 +20,13 @@
 
 import React from 'react'
 import { createEpicMiddleware } from 'redux-observable'
-import { createStore, combineReducers, applyMiddleware, compose } from 'redux'
+import {
+  createStore,
+  combineReducers,
+  applyMiddleware,
+  compose,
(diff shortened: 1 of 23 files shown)


OskarDamkjaer (approved):
LGTM!

OskarDamkjaer:
Lovely seeing this performance improvement 💯

PR #1275, opened 2021-02-05
Refactor favorites 

OskarDamkjaer:
New attempt att same pr as #1264, opened since rebasing got tricky.

These changes aim to make nested folders, custom ordering with drag and drop, bulk renaming/exporting/deleting easier to implement as well as make the code easier to understand for new developers and decrease the surface area for bugs. Should preserve all behaviour, except this allows for empty folders to exist. Export is also missing since it's getting reworked in the next PR anyway. 

### Changelist:
- Add proper types, remove ts-ignores and disabled nullchecks
- Move to hook based drag n drop API
- Make duplicated buttons into separate component
- Update file naming to better match our naming standard
- Updated the data structures, details bellow
- Pulled out project files component to allow for lazy loading of the script contents and simpler apis

In redux and firebase the favorite state is shaped like this (somewhat simplified):
`Favorite: {content: string, id: string, folderId: string}`
`Folder: {name: string, id: string}` 

and is then transformed to only
`Favorite: {content: string, id: string, path: string}` 
Where the folders are all encoded in the path of the favorite. This adds a lot of complexity as:
- Empty folders need special handling in the UI (as they can only exist in the first representation)
- It's harder to add nested folders in the future (instead of just adding a parentFolderId to folders we need to parse and manipulate all nested folders/favorites paths)
- a lot of util functions exist only to handle the transformations 
- it's harder to understand (at least I needed to work with favorites with quite a while until I understood the data flows properly)

Preview available at http://favorites_refactoring.surge.sh.


src/browser/components/ProjectFiles/ProjectFilesList.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
```
HerrEmil, line 1: New files should have the copyright comment on top

src/shared/services/exporting/favoriteUtils.ts
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2002-2021 "Neo4j,"
```
HerrEmil, line 2: The years were removed from copyright texts in a recent PR

HerrEmil (commented):
Screenshot 2021-02-08 at 10 36 56
Black text on dark grey background is hard to read, in dark mode.

Clicking outside to close the rename/edit dropdown did not work for me, I have to press the X again or close the sidepanel to close the dropdown.

src/browser/components/SavedScripts/SavedScriptsListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+    setNameValue
+  } = useNameUpdate(
+    displayName,
+    () => renameScript && renameScript(script, currentNameValue)
+  )
+  const blurRef = useCustomBlur(doneEditing)
+  const drag = useDrag({
+    item: { id: script.id, type: 'script' }
+  })[1]
+  const [showThing, set] = useState(false) // testa :active ist för denna proppen
```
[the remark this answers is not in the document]
OskarDamkjaer: Oh these changes were meant to go in the next PR, that's why the naming is so bad naming/hard to read text, sorry!

HerrEmil (approved):

PR #1278, opened 2021-02-09
Favorites ux refresh

OskarDamkjaer:
This PR aims to improve the usefulness of favorites by adding a context menu to each favorite and adding bulk operations.

Changelist (in the order they appear in the "Files changed" tab):
- Remove dependency on dateformat
- Update e2e tests to test new functionality
- Remove unneeded wrapper tags and classnames
- Simplify project files errors component (visible when you try to create a project file with a name that already exists)
- Use regex pattern in <input/> to find what filenames are okay rather than code
- Add selection state of favorites and bulk delete/export
- Add context menu on favorite for that allows for deleting/renaming/editing/duplicating/running

CleanShot 2021-02-09 at 12 42 21

Preview available at http://favorites_ux_refresh.surge.sh


src/browser/modules/App/App.tsx
```diff
@@ -136,10 +136,8 @@ export function App(props: any) {
     databases
   } = props
 
-  const wrapperClassNames = []
-  if (!codeFontLigatures) {
-    wrapperClassNames.push('disable-font-ligatures')
-  }
+  const wrapperClassNames = codeFontLigatures ? '' : 'disable-font-ligatures'
```
OskarDamkjaer, line 139: Fixes a type error and maybe issue with disabled font ligatures

src/browser/modules/Editor/EditorFrame.tsx
```diff
@@ -290,7 +290,6 @@ export function EditorFrame({
                     variables: {
                       projectId,
                       fileUpload: new File([editorValue], name),
-                      destination: `./${name}`,
                       overwrite: true
```
OskarDamkjaer, line 293: not needed since we only save to root

src/shared/modules/favorites/favoritesDuck.ts
```diff
@@ -309,11 +309,11 @@ function updateFavoriteFields(
...
   // Name of favorite is the comment on the first line by convention
-  const [nameLine, ...contents] = script.content.split('\n')
+  const [nameLine, ...NonNamecontents] = script.content.split('\n')
   const alreadyHasName = nameLine.startsWith('//')
   return alreadyHasName
-    ? `// ${newName}
-  ${contents.join('\n')}`
-    : `// ${newName}
+    ? `//${newName}
+${NonNamecontents.join('\n')}`
```
OskarDamkjaer, line 316: Solves bug that added whitespace on rename

src/shared/services/exporting/favoriteUtils.ts
```diff
@@ -26,6 +26,22 @@ import { Folder } from 'shared/modules/favorites/foldersDuck'
 
 export const CYPHER_FILE_EXTENSION = '.cypher'
 
+export function exportFavoritesAsBigCypherFile(favorites: Favorite[]): void {
```
OskarDamkjaer, line 29: This feature has been requested, and the implementation works. But there's no UI element to call this function yet

src/shared/services/exporting/favoriteUtils.ts
```diff
@@ -26,6 +26,22 @@ import { Folder } from 'shared/modules/favorites/foldersDuck'
...
+  const fileContent = favorites
```
HerrEmil, line 30: Maybe a matter of taste, but I think a map and join is easier to read than a reduce and template literal
```
const fileContent = favorites
  .map((favorite) => favorite.content)
  .join("\n\n")
  .trim();
```

HerrEmil (approved):

src/shared/services/exporting/favoriteUtils.ts
OskarDamkjaer: Yeah it's clearer, thanks!

src/browser/icons/add-circle.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor">
+	<g>
```
HerrEmil, line 3: I think tab indentation is inconsistent with the other SVGs

HerrEmil (approved):
LGTM

src/browser/icons/add-circle.svg
OskarDamkjaer: Hmm they seem to re-appear on save when I remove them. I'll look into it in another PR

PR #1280, opened 2021-02-10
Frame button interaction changes

HerrEmil:
This PR contains a handful of UI and interaction changes to the frame buttons;

The color of the save as favorite button and the options in the download dropdown have been changed.
Current colors:
Screenshot 2021-02-10 at 16 29 08
New colors:
Screenshot 2021-02-10 at 16 29 36

The save as project file option is displayed as an option in the download dropdown instead of as an icon:
Screenshot 2021-02-10 at 16 35 47

The run button is replaced by a stop button while a query is running. The cancel view inside the frame also has two states to differentiate the query cancelling and query cancelled states:
stop query 2

When closing a frame that has been used more than once, a confirmation modal is displayed asking for confirmation and giving the option to cancel the close frame action:
are you sure 3

A demo of the changes is available here: http://frame-button-interactions.surge.sh/


shkirando (approved):

OskarDamkjaer (approved):
LGTM!

PR #1282, opened 2021-02-19
Canny Support

HerrEmil:
This PR contains 5 changes:
 - Adds a canny notifications icon and widget in the help panel
 - Adds a button with a link to canny feature request page in the help panel
 - Moves the help panel button from the top section to the bottom section of the side panel
 - Displays an unread notifications badge on the help panel button (and on the notifications button)
 - Makes the "Other Resources" section in the help panel collapsed by default with expand/collapse on click

Feb-19-2021 16-52-38

The changes can be demoed in this sandbox: http://canny.surge.sh/


src/browser/modules/Sidebar/Sidebar.tsx
```diff
@@ -152,18 +155,19 @@ const Sidebar = ({
...
       icon: function docsIcon(isOpen: boolean): ReactElement {
         return <DocumentsIcon isOpen={isOpen} title="Help &amp; Resources" />
       },
-      content: DocumentsDrawer
-    }
-  ]
-
-  const bottomNavItemsList = [
+      content: DocumentsDrawer,
+      enableCannyBadge: true
```
OskarDamkjaer, line 169: In what situation would we change this to false? A future setting?

src/browser/services/canny.ts
```diff
@@ -0,0 +1,76 @@
...
+export const cannyOptions = {
+  appID: '601d00fb7a41e3035f75e1e8',
+  position: 'right',
+  align: 'top'
+}
+export const CANNY_FEATURE_REQUEST_URL =
+  'https://neo4j-browser.canny.io/feature-requests'
+
+declare global {
+  interface Window {
```
OskarDamkjaer, line 30: second global window declared in this PR, are they automatically merged? or are they only global for this file

src/browser/services/canny.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      function loadCanny() {
+        if (document.getElementById('canny-jssdk')) {
+          return
+        }
+        const f = document.getElementsByTagName('script')[0]
+        const e = document.createElement('script') as HTMLScriptElement
+        e.type = 'text/javascript'
+        e.async = true
+        e.src = 'https://canny.io/sdk.js'
```
OskarDamkjaer, line 60: Now this feels pretty cursed for some reason, I guess it's not much different to a normal cdn:ed dependency

OskarDamkjaer (commented):
Great to see the increase of proper types :party:

src/browser/modules/Sidebar/Sidebar.tsx
HerrEmil: It's never false, it's true or undefined. We could remove this flag and when building the jsx for the button instead check the name for documents instead of presence of this flag

src/browser/services/canny.ts
HerrEmil: Typescript interfaces are merge when declare several yes, we could move this declaration to the custom file, not sure what is more structured

src/browser/modules/Sidebar/Sidebar.tsx
OskarDamkjaer: Ah had another look and saw why it's needed

src/browser/services/canny.ts
HerrEmil: Any suggestions on how to dispel the curse?

OskarDamkjaer: I think we'll have to live with it if we want canny 🤷

OskarDamkjaer (approved):

PR #1285, opened 2021-02-22
Autocollapse reusable editor when not in use

OskarDamkjaer:
This is a pure css change that collapses the editor when it doesn't have focus to preserve space in the stream.

Preview @ http://autocollapse.surge.sh
CleanShot 2021-02-22 at 14 50 54


HerrEmil (approved):
LGTM 👍

PR #1299, opened 2021-02-25
strip backticks around procedure completion items, correct insertion …

HerrEmil:
also corrects the insertion range of procedures

Before change:
old

After change:
new

Demo: http://strip-backticks.surge.sh/

### DIFF
--- a/src/shared/modules/editor/editorDuck.ts
+++ b/src/shared/modules/editor/editorDuck.ts
@@ -24,7 +24,7 @@ import {
 } from 'cypher-editor-support'
 import { editor, languages } from 'monaco-editor'
 import { Action } from 'redux'
-import { ActionsObservable, Epic } from 'redux-observable'
+import { Epic } from 'redux-observable'
 import Rx from 'rxjs/Rx'
 
 import {
@@ -35,6 +35,7 @@ import { CypherTokensProvider } from 'browser/modules/Editor/CypherTokensProvide
 import cypherFunctions from 'browser/modules/Editor/cypher/functions'
 import consoleCommands from 'browser/modules/Editor/language/consoleCommands'
 import { getUrlParamValue } from 'services/utils'
+import { GlobalState } from 'shared/globalState'
 import { APP_START, URL_ARGUMENTS_CHANGE } from 'shared/modules/app/appDuck'
 import {
   commandSources,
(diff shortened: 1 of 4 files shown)


src/shared/modules/editor/editorDuck.ts
```diff
@@ -139,11 +161,19 @@ export const populateEditorFromUrlEpic = (some$: any, _store: any) => {
     })
 }
 
+// CypherEditorSupport returns the content attributes of procedures with dots wrapped in backticks, e.g. "`apoc.coll.avg`"
+// This function strips any surrounding backticks before we use the .content value in the completion item provider
```
OskarDamkjaer, line 165: 👍

src/shared/modules/editor/editorDuck.ts
```diff
@@ -261,16 +295,14 @@ export const updateEditorSupportSchemaEpic: Epic<Action, any> = (
...
+        labels: state.meta.labels.map(toLabel),
+        relationshipTypes: state.meta.relationshipTypes.map(toRelationshipType),
+        propertyKeys: state.meta.properties.map(toPropertyKey),
         functions: [
           ...cypherFunctions,
           ...state.meta.functions.map(toFunction)
-        ] as string[],
-        procedures: state.meta.procedures.map(toProcedure) as string[]
+        ],
+        procedures: state.meta.procedures.map(toProcedure)
```
OskarDamkjaer, line 305: Nice to see these ts improvements!

OskarDamkjaer (approved):

oskarhane (commented):
Backticks enables tokens (not sure what to call them, but labels, prop-types, procedure names etc) to contain characters like `-`.

This would fail: 
```
CREATE (:Kor-v)
```

This would pass: 
```
CREATE (:`Kor-v`)
```

The problem with this change is that nothing will be escaped anymore.
If you create the above label and type a `MATCH` clause to autocomplete, it will autocomplete with `MATCH (n:Kor-v) RETURN n` which would fail. 

So we do want to keep the back-ticks. Is there a way to work around this?
@HerrEmil @OskarDamkjaer

OskarDamkjaer:
@oskarhane We do have the completion-type available, is there an issue with only stripping backticks for procedures?

HerrEmil:
@oskarhane @OskarDamkjaer I added a check for completion type and only strip surrounding backticks on functions and procedures. I've updated the surge link which works with \`Kor-v\` example now

oskarhane (approved):
Wohoo, that work better!

Can we write some tests to show that this is intended and that we don't break it in the future?

HerrEmil:
Added a few unit tests and plan on merging on green test builds :)

OskarDamkjaer (approved):
:+1:

PR #1300, opened 2021-02-26
Fix: AsciiView ligatures disable

konsalex:
Fixed the ligatures of the `Fira` font family in Text View, by replacing with `Courier New`.

Also replaced `\n` with the escape version `\\n` in table so there are no new lines that will break the ascii table.

Before: before

After:
image

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/AsciiView.tsx
+++ b/src/browser/modules/Stream/CypherFrame/AsciiView.tsx
@@ -30,7 +30,8 @@ import {
   StyledRightPartial,
   StyledWidthSliderContainer,
   StyledWidthSlider,
-  StyledTruncatedMessage
+  StyledTruncatedMessage,
+  StyledAsciiPre
 } from '../styled'
 import {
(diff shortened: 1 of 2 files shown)


src/browser/modules/Stream/CypherFrame/AsciiView.tsx
```diff
@@ -122,6 +126,16 @@ export class AsciiViewComponent extends Component<
     this.props.setAsciiMaxColWidth(maxColWidth)
   }
 
+  /**
+   * Replaces newline characters, with a double \\ to escape newline in render
+   */
+  removeNewlines(serializedRows: SerializedRows): SerializedRows {
+    const newSerializedRows: SerializedRows = serializedRows.map(row => {
+      return row.map(value => value.replace('\n', '\\n'))
```
OskarDamkjaer, line 134: In the rest of the codebase, we've not aliased simple types, so I'd personally prefer if we left it as string[][]. You also only need to write the type on the input argument, so it's not needed to specify it on row 133 as it's inferred.

PR #1301, opened 2021-03-01
Hide Peek Problem status bar, disable escape key toggling fullscreen when autocompleting

HerrEmil:
Two small fixes, demo: http://peek-problem.surge.sh/


OskarDamkjaer (approved):
👍

PR #1302, opened 2021-03-01
Fix issue where fullscreen couldn't be closed

OskarDamkjaer:
Previously fullscreen couldn't be closed after doing the following:
Steps to reproduce

1. Write a query (eg RETURN 1) and execute
2. Update the query in reusable frame (eg make it RETURN 2) and execute
3. Go fullscreen in reusable frame
4. Try to delete the reusable frame with X

This was due to the built in modal going underneath the fullscreen. I decreased all our z-indexes so the modal is now on top.
Preview at http://fullscreen_close.surge.sh


HerrEmil (approved):
🥇

PR #1303, opened 2021-03-01
remove click handler on editor container

HerrEmil:
The click handler was causing the search and replace input fields to immediately lose focus when clicking them. The click handler was not necessary since the monaco component fills almost the whole container in any case.

demo: http://search-input-click.surge.sh/

GIFs recorded of opening search and clicking the input fields, before:
old

After:
new


OskarDamkjaer (approved):
💯

OskarDamkjaer:
Good to see global state spreading!

PR #1304, opened 2021-03-01
Update dependency to fix bug with missing line numbers

OskarDamkjaer:
Steps to reproduce issue fixed:
1. Write query that spans multiple lines
2. Click to edit the query in the result frame and see the following:
image

This PR fixes that by updating the monaco dependency. However later versions of monaco also highlight the current line number, I've changed the colors so it keeps the normal colors.


HerrEmil (approved):
👍

PR #1305, opened 2021-03-01
Stay in viz when loading new results

OskarDamkjaer:
Sometimes reusable frame switches away from the graph visualisation. Steps to reproduce, with the movie db:
Run:
MATCH (n) RETURN n;
Edit and run:
MATCH (n :Movie) RETURN n; 

should stay as on the viz with these changes.

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/index.tsx
+++ b/src/browser/modules/Stream/CypherFrame/index.tsx
@@ -27,6 +27,7 @@ import Centered from 'browser-components/Centered'
 import {
   getRequest,
   REQUEST_STATUS_PENDING,
+  REQUEST_STATUS_SUCCESS,
   isCancelStatus,
   BrowserRequest,
   BrowserRequestResult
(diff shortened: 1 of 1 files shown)


src/browser/modules/Stream/CypherFrame/index.tsx
```diff
@@ -171,7 +171,9 @@ export class CypherFrame extends Component<CypherFrameProps, CypherFrameState> {
     }
 
     // When frame re-use leads to result without visualization
-    if (!this.canShowViz() && this.state.openView === viewTypes.VISUALIZATION) {
+    const doneLoading = this.props.request.status === REQUEST_STATUS_PENDING
```
HerrEmil, line 174: I don't see how status pending equals loading being done, is that true?

OskarDamkjaer: Oh, you're right. This fix was actually not working.. I just thought it was.

HerrEmil (approved):

PR #1306, opened 2021-03-02
Fix ligatures text view

konsalex:
Updated Jest snapshots to pass the Jest tests, and disconnected user from browser on cypress Contact form test before running.


OskarDamkjaer (changes requested):
Please do an interactive rebase/ squash to make sure it's only the 
konsalex user in the commits :+1:

OskarDamkjaer (approved):

PR #1308, opened 2021-03-03
Hide empty duplicate folders

OskarDamkjaer:
Old localstorage state contains duplicated folders, we will want to remove the duplicated folders from there, but for now let's just hide them.

demo: http://fix_duplicate_folders.surge.sh


OskarDamkjaer:
The failing e2e test is due to not having rebased on master to get the skipped :play-command test, it's a non-issue

HerrEmil (approved):
LGTM

PR #1309, opened 2021-03-03
Show favorites with malformed folderId

OskarDamkjaer:
Bug is caused by drag and drop putting incorrect folderIds to favorites in a previous version of browser

demo @ http://fix_drag_outside_bug.surge.sh

### DIFF
--- a/src/browser/components/SavedScripts/SavedScripts.tsx
+++ b/src/browser/components/SavedScripts/SavedScripts.tsx
@@ -84,8 +84,11 @@ export default function SavedScripts({
   exportScripts,
   createNewFolder
 }: SavedScriptsProps): JSX.Element {
+  const folderExists = (folderId: string) =>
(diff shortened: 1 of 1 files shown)


src/browser/components/SavedScripts/SavedScripts.tsx
```diff
@@ -84,8 +84,11 @@ export default function SavedScripts({
   exportScripts,
   createNewFolder
 }: SavedScriptsProps): JSX.Element {
+  const folderExists = (folderId: string) =>
+    folders.find(folder => folder.id === folderId)
```
jharris4, line 88: Thanks for explaining the reason for the change, it makes sense to me. Minor nit would be that this find could be made more efficient using something like a map, but it's not worth blocking this PR for, especially since we plan to refactor this part of the code anyway :-)

jharris4 (approved):
LGTM

PR #1312, opened 2021-03-04
add workaround to squashed command palette in small cypher editors

HerrEmil:
Fix for squashed command palette. Demo: http://command-palette-height.surge.sh/

Before:
old

After:
new


jharris4 (approved):
Tested this out in multiple frames etc and resizing the browser and works well!

PR #1313, opened 2021-03-05
Display Cypher warnings on correct positions

HerrEmil:
Before:
Screenshot 2021-03-05 at 14 50 37
After:
Screenshot 2021-03-05 at 14 50 51

Demo: http://warning-column-position.surge.sh/

### DIFF
--- a/src/browser/modules/Editor/Monaco.tsx
+++ b/src/browser/modules/Editor/Monaco.tsx
@@ -19,7 +19,7 @@
  */
 
 import { QuickInputList } from 'monaco-editor/esm/vs/base/parts/quickinput/browser/quickInputList'
-import { parse } from 'cypher-editor-support'
+import { parse, QueryOrCommand } from 'cypher-editor-support'
 import { debounce } from 'lodash-es'
 import {
   editor,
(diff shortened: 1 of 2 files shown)


src/browser/modules/Editor/Monaco.tsx
```diff
@@ -446,33 +447,32 @@ const Monaco = forwardRef<MonacoHandles, MonacoProps>(
...
-                  }) => ({
-                    startLineNumber: statementLineNumber + line,
-                    startColumn: 1,
-                    endLineNumber: statementLineNumber + line,
-                    endColumn: 1000,
-                    message: title + '\n\n' + description,
-                    severity: MarkerSeverity.Warning
-                  })
+                  ({ description, position, title }) => {
+                    const { line, column } = position as NotificationPosition
```
OskarDamkjaer, line 462: Seems from the types that position could be `{}` and that's why we cast to NotificationPosition is that not a case we need to handle?

HerrEmil: I have not seen it happen, the type info might be incorrect, but I've refactored this line so accessing line and column is type safe now.

OskarDamkjaer (approved):

PR #1320, opened 2021-03-11
add type information to schema converters

HerrEmil:
adds typescript types, caught one potential type error where a regex match was assumed to not be null


OskarDamkjaer:
Closing this in favour of #1396 to rebase without pushing to/depending on emils fork :)

PR #1321, opened 2021-03-12
Promote Aura in community edition

OskarDamkjaer:
Keep in mind we've only added this for community edition, so you'll need a community server to connect to. Quickest way to do so: `docker run --env NEO4J_AUTH=neo4j/newpassword -p7687:7687 neo4j:4.2.2`

Preview at: http://aura_promotion.surge.sh image

The plan is to revert this pull request after one or two releases when the promotion ends.


jharris4 (approved):
This looks great, especially now that the static components have been moved to the top level!

As mentioned in a comment we might want to clean up the svg content to serve it from files later on, but that's not worth holding things up for now.

OskarDamkjaer:
Some of the changes in this commit have already been ported to master, some are supposed to be reverted in the next release and there's a lot of merge conflicts, so I'm closing this in favor of: #1378

PR #1323, opened 2021-03-16
Cleanup icon naming

OskarDamkjaer:
When adding a new icon for guides, I figured it was time to clean up some of the icons we already have. 

Changelist:
- Sorted icon imports
- standardized naming of icons when imported and as files (open to any other naming convention) 
- Added a few missing types
- Renamed another file also called icons to make search easier
- Ran prettier on canny icons

@jharris4 If you'd like to add dynamic loading of icons as well, we could pair on that


src/browser/components/icons/Icons.tsx
```diff
@@ -22,35 +22,35 @@ import React from 'react'
...
-import cannyNotifications from 'icons/canny-changelog.svg'
+import pin from 'icons/pin.svg'
+import ratingStar from 'icons/rating-star.svg'
+import runIcon from 'icons/run-icon.svg'
+import saveFavorite from 'icons/save-favorite.svg'
+import shrink from 'icons/shrink.svg'
+import skipPrev from 'icons/skip-prev.svg'
+import stopIcon from 'icons/stop-icon.svg'
+import table from 'icons/table.svg'
+import text from 'icons/text.svg'
```
jharris4, line 53: This new sort order makes my inner OCD very happy, thanks for doing this! :-)

src/browser/components/icons/Icons.tsx
```diff
@@ -220,10 +220,10 @@ export const AboutIcon = ({ isOpen, title }: SidebarIconProps): JSX.Element => (
 )
 
 export const SlidePreviousIcon = (): JSX.Element => (
-  <IconContainer icon={arrowLeft1} width={20} />
+  <IconContainer icon={arrowLeft} width={20} />
```
jharris4, line 223: Can we make these widths constants?

```
const LARGE_SIZE = 20
const SMALL_SIZE = 12
```

src/browser/components/icons/Icons.tsx
```diff
@@ -421,12 +421,20 @@ export const FireExtinguisherIcon = ({
   <IconContainer className="fa fa-fire-extinguisher" title={title} />
 )
 
-export const NewFolderIcon = () => <IconContainer icon={newFolder} width={12} />
-export const NavIcon = () => <IconContainer icon={navIcon} width={12} />
-export const AddIcon = () => <IconContainer icon={addCircle} width={12} />
-export const FolderIcon = () => <IconContainer icon={folder} width={12} />
-export const SavedScriptsPlay = () => (
-  <IconContainer icon={run_icon} width={12} />
+export const NewFolderIcon = (): JSX.Element => (
```
jharris4, line 424: There's a lot of duplication in creating all these IconContainer wrappers.

What do you think of the idea of having helpers to clean that up?


Something like:
```
const createIconContainer = props => <IconContainer ...props />
const createSmallIconContainer = props => createIconContainer({...props, width: SMALL_SIZE})
const createLargeIconContainer = props => createIconContainer({...props, width: LARGE_SIZE})
```

If nothing else, this might remove the need for all the TypeScript types?

src/browser/components/icons/Icons.tsx
```diff
@@ -326,7 +326,7 @@ export const PinIcon = (): JSX.Element => (
   <IconContainer icon={pin} width={12} />
 )
 export const SaveFavorite = (): JSX.Element => (
```
jharris4, line 328: Missing the `Icon` suffix here, not sure if it got missed anywhere else, but would be nice to have them all be consistent! :-)

jharris4 (changes requested):
I really like this cleanup! I have some minor comments about cleaning up some duplication, and also that we could also fix some inconsistent naming of the exported icons (some are missing `Icon` suffix).

One other things that we should watch out for is that I notice that some of the svg files have inline style declarations like `<style>.a { ... }</style><svg class="a"/>` which can cause a LOT of problems where the styles leak out and affect other svgs. We definitely want to clean those up, but could be part of another PR...

src/browser/components/icons/Icons.tsx
OskarDamkjaer: Really easy to do in vim as well, with it the lines selected all I had to to was type :sort 💯 kinda want to start doing it every now and then on messy imports

src/browser/components/icons/Icons.tsx
OskarDamkjaer: Thanks! Went through and found a couple more misnamed icons

src/browser/components/icons/Icons.tsx
OskarDamkjaer: Hmm, the typescript types are there since we have a rule on fully specified In/out values for things that are exported. I'm open to letting return values be inferred even on exported functions, but that's another discussion

OskarDamkjaer: I would want to do helpers like these in the future, but I feel it's a premature optimization to do so now as we have such a large spread of ways icons are done in the code base (via several fonts, semantic UI, as svg files and a few more). So I think another layer of indirection isn't helpful until we've standardized our iconhandling

src/browser/components/icons/Icons.tsx
OskarDamkjaer: Sure! There's another one as well, adding that one as well

OskarDamkjaer:
Thanks for the close look, I've made some adjustments/comments now. Agree on the inline styles and in addition to that I think there are some svgs that don't respect their color/width properly, will leave both to another PR.

src/browser/components/icons/Icons.tsx
jharris4: Yeah, makes sense to me! Thanks for clarifying your thoughts :-)

jharris4 (approved):
Changes look good to me, LGTM! (Approving and happy to merge once tests go green)

PR #1336, opened 2021-04-06
Stream refactor

OskarDamkjaer:
This refactor lifts the FrameTitlebar up to enable the fullscreen behaviour we're looking for in guides.

testable here: http://stream_refactor.surge.sh


src/browser/modules/Frame/FrameTitlebar.tsx
```diff
@@ -127,138 +96,19 @@ type FrameTitleBarProps = FrameTitleBarBaseProps & {
 }
 
 function FrameTitlebar(props: FrameTitleBarProps) {
-  const [editorValue, setEditorValue] = useState(props.frame.cmd)
-  const [renderEditor, setRenderEditor] = useState(props.frame.isRerun)
+  const frame = props.frame
```
jharris4, line 99: Any reason not to use destructuring here? ( `const { frame } = props` )

src/browser/modules/Stream/CypherScriptFrame/CypherScriptFrame.tsx
```diff
@@ -102,9 +102,8 @@ function CypherScriptFrame({
       </ContentSizer>
     </WrapperCenter>
   )
-  return (
-    <FrameTemplate className="no-padding" header={frame} contents={contents} />
-  )
+  // This wanted to disable padding on the template. will have to do it in other way than classname.
```
jharris4, line 105: Looks like you've already addressed this in the StyledComponents logic, so this comment can be removed?

src/browser/modules/Stream/Stream.tsx
```diff
@@ -50,9 +50,13 @@ import {
...
 import DbsFrame from './Auth/DbsFrame'
-import EditFrame from './EditFrame'
+import { StyledFrame } from '../Frame/styled'
+import FrameTitlebar from '../Frame/FrameTitlebar'
+import { dim } from 'browser-styles/constants'
+import styled from 'styled-components'
+import ExportButton, { ExportItem } from './ExportButtons'
 
-const trans = {
+const nameToFrame: Record<string, React.ComponentType<any>> = {
```
jharris4, line 59: Love that you've given this a name that's actually descriptive!

OskarDamkjaer: And it also ryhmes

src/browser/modules/Stream/CypherScriptFrame/CypherScriptFrame.tsx
OskarDamkjaer: yes, good catch

src/browser/modules/Frame/FrameTitlebar.tsx
OskarDamkjaer: no

PR #1339, opened 2021-04-07
Update node version

OskarDamkjaer:


jharris4:
I'd leave this the Node version at 14 since 15 is not LTS... And in the package.json, I'd use `^12` since it works fine with version 12 and above...

jharris4 (approved):
Thanks for making this change, LGTM!

PR #1341, opened 2021-04-07
Update cypress to v7

OskarDamkjaer:


package.json
```diff
@@ -101,7 +101,7 @@
     "copy-webpack-plugin": "^6.0.3",
     "cross-env": "^6.0.3",
     "css-loader": "^1.0.0",
-    "cypress": "4.12.1",
+    "cypress": "^7.0.0",
```
jharris4, line 104: Cypress 7 requires node `>= 14.16.0` so we should probably update the `engines` rule in our `package.json` to match that...

https://github.com/cypress-io/cypress/blob/d32076be075c17001249afb712005da32a31d636/package.json#L197

jharris4 (approved):
LGTM

PR #1351, opened 2021-04-09
Update ui bundle to 0.10.5

recrwplay:
Updates the ui bundle version for the docs to the latest: 0.10.5

### DIFF
--- a/docs/preview.yml
+++ b/docs/preview.yml
@@ -15,7 +15,7 @@ content:
     - '!**/README.adoc'
 ui:
   bundle:
-    url: https://s3-eu-west-1.amazonaws.com/static-content.neo4j.com/build/ui-bundle-v0.10.5.zip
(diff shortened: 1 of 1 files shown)


docs/preview.yml
```diff
@@ -37,11 +37,12 @@ asciidoc:
     page-no-canonical: true
     page-origin-private: true
     page-hide-toc: false
+    page-mixpanel: 4bfb2414ab973c741b6f067bf06d5575
```
martin-neotech, line 40: what does it do?

martin-neotech (approved):

docs/preview.yml
recrwplay: Good question!

It is used to track some user interactions in the page, such as switching tabs (eg in drivers), copying or running code from code blocks.

PR #1353, opened 2021-04-11
Name exports

OskarDamkjaer:
An attempt to have the component tree be more readable. However doesn't seem to have any effect from a cursory glance, thoughts?

### DIFF
--- a/src/browser/components/Directives.tsx
+++ b/src/browser/components/Directives.tsx
@@ -142,4 +142,7 @@ const mapDispatchToProps = (_dispatch: any, ownProps: any) => {
   }
 }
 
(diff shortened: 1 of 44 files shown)


jharris4:
Hmm, I'm remembering now that in the past I've had to do some fiddling with the `displayName` for HOC functions, but maybe that's not applicable at least for `react-redux` `connect`.

Some interesting reading [here](https://github.com/reduxjs/react-redux/issues/329) and [here](https://stackoverflow.com/questions/55290633/how-do-i-modify-the-default-name-of-react-redux-connect-function)

Now that I look more closely at the component tree (both with & without this PR) I'm still seeing a bunch of `Anonymous` components, but maybe that's unrelated to the `export default` issue...

PR #1354, opened 2021-04-12
Immersive guides

OskarDamkjaer:


src/browser/documentation/guides/sandboxMovies.tsx
```diff
@@ -0,0 +1,1120 @@
+import React from 'react'
```
jharris4, line 1: Is this file used for anything?

src/browser/modules/Frame/styled.tsx
```diff
@@ -105,9 +106,9 @@ export const StyledFrameContents = styled.div<FullscreenProps>`
   min-height: ${dim.frameBodyHeight / 2}px;
   max-height: ${props =>
     props.fullscreen
-      ? '100vh'
+      ? 'auto'
       : dim.frameBodyHeight - dim.frameStatusbarHeight * 2 + 'px'};
-  ${props => (props.fullscreen ? 'height: 100vh' : null)};
+  ${props => (props.fullscreen ? 'height: calc(100vh - 40px)' : null)};
```
jharris4, line 111: Would it make sense to make the `40` value here a const, especially if it's also used to set the height of the title bar?

src/browser/modules/Sidebar/Sidebar.tsx
```diff
@@ -24,6 +24,7 @@ import DatabaseDrawer from '../DBMSInfo/DBMSInfo'
 import DocumentsDrawer from './Documents'
 import AboutDrawer from './About'
 import SettingsDrawer from './Settings'
+import GuideDrawer from './GuidesDrawer'
```
jharris4, line 27: Naming mismatch here

src/browser/modules/Stream/CypherFrame/VisualizationView.styled.tsx
```diff
@@ -21,13 +21,12 @@
...
-export const StyledVisContainer: any = styled.div`
+export const StyledVisContainer = styled.div<{ fullscreen: boolean }>`
   width: 100%;
   overflow: hidden;
-  ${(props: any) => (props.fullscreen ? 'padding-bottom: 39px' : null)};
-  height: ${(props: any) =>
+  height: ${props =>
     props.fullscreen
-      ? '100vh'
+      ? 'calc(100vh - 40px)'
```
jharris4, line 29: same comment about making this value (40) a const...

src/shared/services/guideResolverHelper.tsx
```diff
@@ -0,0 +1,147 @@
+import React from 'react'
```
jharris4, line 1: As discussed on our call, it's nice that you've moved this to its own file, and while it may be out of scope for this PR, it'd be really nice to also refactor the play command to share this code to get rid of the duplicated logic later on.

src/browser/documentation/guides/sandboxMovies.tsx
OskarDamkjaer: It's not! Should have a done a quick self review and removed it 😓

src/browser/modules/Frame/styled.tsx
OskarDamkjaer: Yes, even better would be to refactor the fullscreen stuff to not need these calc workarounds. After a quick demo /w design/pm we decided to skip the fullscreen until we can do so properly, so I've reverted the commit that adds these fullscreen changes.

src/browser/components/Directives.tsx
```diff
@@ -104,7 +118,10 @@ export const Directives = (props: any) => {
       directives.forEach(directive => {
         const elems = elem.querySelectorAll(directive.selector)
         Array.from(elems).forEach((e: any) => {
-          if (e.firstChild.nodeName !== 'I') {
+          if (
+            e.firstChild.nodeName !== 'I' &&
```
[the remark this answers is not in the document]
OskarDamkjaer: updated.

src/browser/components/Directives.tsx
```diff
@@ -28,6 +28,20 @@ import * as editor from 'shared/modules/editor/editorDuck'
 import { addClass, prependIcon } from 'shared/services/dom-helpers'
 
 const directives = [
+  {
+    selector: '[data-exec]',
```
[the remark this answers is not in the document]
OskarDamkjaer: Kept naming as they omit the : and therefore work differently

src/shared/services/remoteUtils.ts
```diff
@@ -33,7 +33,7 @@ const removeOnHandlersFromHtml = (string: any) =>
     ''
   )
 
-export function cleanHtml(string: any) {
+export function cleanHtml(string: string): string {
   if (typeof string !== 'string') return string
```
jharris4, line 37: is this check still necessary now that you've added the type to the param?

src/shared/services/guideResolverHelper.tsx
```diff
@@ -0,0 +1,173 @@
...
+  StyledHelpFrame,
+  StyledPreformattedArea
+} from 'browser/modules/Stream/styled'
+
+const { chapters } = docs.guide
+const unfound = { slides: [guideUnfound.content], title: guideUnfound.title }
+
+export async function resolveGuide(
+  guideName: string,
+  store: any
```
jharris4, line 52: I think it'd be cleaner to just pass the state here instead of the store, since all of the downstream functions only seem to use `store.getState()`

src/shared/modules/sidebar/sidebarDuck.ts
```diff
@@ -51,6 +53,7 @@ type DrawerId =
   | 'about'
   | 'project files'
   | 'settings'
+  | typeof GUIDE_DRAWER_ID
```
jharris4, line 56: why the use of `typeof` here?

src/browser/modules/Sidebar/GuidesDrawer.tsx
```diff
@@ -0,0 +1,93 @@
...
+}: GuidesDrawerProps): JSX.Element {
+  const scrollRef = useRef<HTMLDivElement>(null)
+  const title = guide.title.startsWith('http')
+    ? guide.title
+    : guide.title.charAt(0).toUpperCase() + guide.title.slice(1)
+
+  return (
+    <StyledGuidesDrawer id="guide-drawer" ref={scrollRef}>
+      <StyledGuidesDrawerHeader onClick={backToAllGuides}>
+        {guide.title !== defaultGuide.title && (
```
jharris4, line 61: I find this logic a little hard to follow, makes me wonder if making it some appropriately named utility function could make it more clear. Ex, either `isDefaultGuide(guide)` or some other name that better describes the logic/reasoning for this condition.

src/browser/modules/Sidebar/GuidesDrawer.tsx
```diff
@@ -0,0 +1,93 @@
...
+  gotoSlide: (slideIndex: number) => void
+}
+
+function GuidesDrawer({
+  guide,
+  backToAllGuides,
+  gotoSlide
+}: GuidesDrawerProps): JSX.Element {
+  const scrollRef = useRef<HTMLDivElement>(null)
+  const title = guide.title.startsWith('http')
```
jharris4, line 54: I'm curious how this works with the idea of using the html title element content for remote guides...

src/browser/modules/Carousel/styled.tsx
```diff
@@ -302,3 +304,120 @@ export const StyledSlide = styled.div`
...
+  }
+  & pre.runnable {
+    background-color: ${dark.preBackground};
+    color: ${dark.preText};
+  }
+  & pre.content {
+    background-color: ${dark.secondaryBackground};
+    color: ${dark.preText};
+  }
+  & a[help-topic],
```
jharris4, line 391: is it intentional that the first one had `&` prefix but the others don't? Also it feels like it could be better to just use a classname to simplify this kind of styling...

src/browser/documentation/index.ts
```diff
@@ -188,25 +338,41 @@ export default {
...
+      movieGraph: guideMovieGraph,
+      'movie-graph': guideMovieGraph,
       northwind: guideNorthwindGraph,
-      iconography: guideIconography,
-      start: guideStart,
-      typography: guideTypography,
-      unfound: guideUnfound,
-      writeCode: guideWritecode,
-      // Commands only
-      'query-template': {
```
jharris4, line 206: was this removed on purpose? I don't see it in the new `play` section you created.

src/shared/services/remoteUtils.ts
OskarDamkjaer: No but we have a lint rule to require explicit return types on exported fns

src/shared/services/guideResolverHelper.tsx
OskarDamkjaer: makes sense!

src/shared/modules/sidebar/sidebarDuck.ts
OskarDamkjaer: typeof GUIDE_DRAWER_ID evaluates to 'guides', so this is just a way to reuse the variable

src/browser/modules/Sidebar/GuidesDrawer.tsx
OskarDamkjaer: yes that's better

src/browser/modules/Sidebar/GuidesDrawer.tsx
OskarDamkjaer: Removed

src/browser/modules/Sidebar/GuidesDrawer.tsx
```diff
@@ -51,21 +51,26 @@ function GuidesDrawer({
...
+        {!isDefaultGuide(guide) && (
           <BackIconContainer>
             <BackIcon width={16} />
           </BackIconContainer>
         )}
         Neo4j Browser Guides{' '}
       </StyledGuidesDrawerHeader>
-      <GuideTitle title={title}>{title}</GuideTitle>
+      <div
+        style={{
```
jharris4, line 66: Can these styles be moved to `styled.tsx` or what not?

OskarDamkjaer: Yeah, I forgot it! Moving it now..

jharris4 (approved):
I tested everything I could think of, including remote guides, cypher-template, missing guides and it all worked really solidly. Great work!

jharris4 (approved):
LGTM

PR #1358, opened 2021-04-13
Fix styled component names in dev tools

OskarDamkjaer:


jharris4:
It's great that you've made this improvement (and made it a separate PR now!). The only question I have is whether this will be enabled in production mode, and if so, whether we should disable it in production mode...?

jharris4 (approved):
LGTM

PR #1359, opened 2021-04-14
add segment tracking

jharris4:

### DIFF
--- a/src/browser/modules/App/App.tsx
+++ b/src/browser/modules/App/App.tsx
@@ -67,6 +67,7 @@ import UserInteraction from '../UserInteraction'
 import DocTitle from '../DocTitle'
 import asTitleString from '../DocTitle/titleStringBuilder'
 import Intercom from '../Intercom'
+import Segment from '../Segment'
 import Render from 'browser-components/Render'
 import BrowserSyncInit from '../Sync/BrowserSyncInit'
 import { getMetadata, getUserAuthStatus } from 'shared/modules/sync/syncDuck'
@@ -85,6 +86,8 @@ import { METRICS_EVENT, udcInit } from 'shared/modules/udc/udcDuck'
 import { useKeyboardShortcuts } from './keyboardShortcuts'
 import PerformanceOverlay from './PerformanceOverlay'
 
+declare let SEGMENT_KEY: string
+
 export function App(props: any) {
   const [derivedTheme, setEnvironmentTheme] = useDerivedTheme(
     props.theme,
@@ -101,6 +104,7 @@ export function App(props: any) {
   useKeyboardShortcuts(props.bus)
 
   const eventMetricsCallback = useRef((_: any) => {})
+  const segmentTrackCallback = useRef((_a: any, _b: any) => {})
 
   useEffect(() => {
     const unsub =
@@ -109,6 +113,13 @@ export function App(props: any) {
         eventMetricsCallback &&
           eventMetricsCallback.current &&
           eventMetricsCallback.current({ category, label, data })
+        segmentTrackCallback &&
+          segmentTrackCallback.current &&
+          segmentTrackCallback.current(category + '-' + label, {
+            category,
+            label,
+            data
+          })
       })
     const initAction = udcInit()
     props.bus && props.bus.send(initAction.type, initAction)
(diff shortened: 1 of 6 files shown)


src/browser/modules/App/App.tsx
```diff
@@ -178,6 +193,10 @@ export function App(props: any) {
               <UserInteraction />
               <Render if={loadExternalScripts}>
                 <Intercom appID="lq70afwx" />
+                <Segment
+                  segmentKey={SEGMENT_KEY}
+                  setTrackCallback={setTrackSegmentCallback}
+                />
```
oskarhane, line 199: 👍 Good to see that you found and use existing patterns!

src/browser/modules/Segment/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+    } = this.props
+    if (!canUseDOM()) return
+    updateData({ ...otherProps, segmentKey: segmentKey })
+  }
+
+  shouldComponentUpdate() {
+    return false
+  }
+
+  // @ts-expect-error ts-migrate(7030) FIXME: Not all code paths return a value.
```
OskarDamkjaer, line 139: could we fix this warning?

OskarDamkjaer (approved):

src/browser/modules/Segment/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+              'page',
+              'once',
+              'off',
+              'on',
+              'addSourceMiddleware',
+              'addIntegrationMiddleware',
+              'setAnonymousId',
+              'addDestinationMiddleware'
+            ]
+            analytics.factory = function(t: any) {
```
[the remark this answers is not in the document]
jharris4: yes, copied and pasted from their setup intructions...

src/browser/modules/App/App.tsx
```diff
@@ -100,16 +103,23 @@ export function App(props: any) {
 
   useKeyboardShortcuts(props.bus)
 
-  const eventMetricsCallback = useRef((_: any) => {})
+  const eventMetricsCallback = useRef((_: MetricsData) => {})
+  const segmentTrackCallback = useRef((_: MetricsData) => {})
```
OskarDamkjaer, line 107: you can probably remove the argument all together here

src/browser/modules/Segment/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+              analytics._loadOptions = e
+            }
+            analytics._writeKey = segmentKey
+            analytics.SNIPPET_VERSION = '4.13.2'
+            analytics.load(segmentKey)
+            const doTrack = (metricsData: MetricsData) => {
+              const { category, label, data } = metricsData
+              window.analytics.track(category + '-' + label, {
+                ...data,
+                desktop: inDesktop
```
OskarDamkjaer, line 113: @oskarhane I was thinking we would set the environment to WEB, CLOUD, or DESKTOP?

OskarDamkjaer (changes requested):

src/browser/modules/App/App.tsx
OskarDamkjaer: Let's leave it as it help useRef infer the right typ

src/browser/modules/Segment/index.tsx
jharris4: yep, but with the new centrally hosted Browser we will never have `CLOUD` as an `env` anymore once/if Aura move over to using the centrally hosted Browser...

OskarDamkjaer: Hmm, seems env will not tell us if we're on aura anymore, so rather add a connected to aura 🤔

OskarDamkjaer: {
hostedInDesktop: boolean,
connectedToAuraHost: boolean
}

oskarhane: @OskarDamkjaer @jharris4 
Maybe not part of this PR, but it's a good point that it might be interesting to know.
But the envs would be something like STANDALONE, EMBEDDED, DESKTOP where EMBEDDED it when it's served up from Neo4j server.
Could be worth a card to think about if it's useful and how to identify the differents envs.

OskarDamkjaer (approved):

PR #1360, opened 2021-04-14
Initial review comments

gregoryking:


renetapopova:
Thanks so much for the review @gregoryking! I'll address your comments in a separate PR(s) because they are based on older versions of these files.

renetapopova:
Thanks a lot @gregoryking! The comments that haven't been addressed in the manual are now here #1383.

renetapopova:
Comments have been addressed.

PR #1367, opened 2021-04-20
ignore enter/escape key from Monaco when find widget visible

jharris4:
http://monaco-search-fixes.surge.sh/


OskarDamkjaer (approved):
Nice and clean solution! 🙏

PR #1370, opened 2021-04-21
Use last db instead of default db where possible

jharris4:
http://remember-last-used-db.surge.sh

### DIFF
--- a/src/shared/modules/connections/connectionsDuck.ts
+++ b/src/shared/modules/connections/connectionsDuck.ts
@@ -73,6 +73,7 @@ export type ConnectionReduxState = {
   connectionState: ConnectionState
   lastUpdate: number
   useDb: string | null
+  lastUseDb: string | null
 }
 type ConnectionState =
   | typeof DISCONNECTED_STATE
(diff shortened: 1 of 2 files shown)


src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -356,12 +363,32 @@ export const setAuthEnabled = (authEnabled: any) => {
 
 export const useDb = (db: any = null) => ({ type: USE_DB, useDb: db })
 
+export const resetUseDb = () => ({ type: USE_DB, useDb: null })
+
+export const useDefaultDb = (db: any, databases: any) => ({
```
jharris4, line 368: TODO rename this to `useDefaultOrLastUsedDb`

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -49,7 +50,7 @@ import {
   getDbClusterRole
 } from '../features/versionedFeatures'
 import { extractServerInfo } from './dbMeta.utils'
-import { assign, reduce } from 'lodash-es'
+import { assign, last, reduce } from 'lodash-es'
```
OskarDamkjaer, line 53: Doesn't seem to be used anywhere

OskarDamkjaer (approved):
Looks good to me! 💯

PR #1371, opened 2021-04-21
Guides touchups

OskarDamkjaer:
This PR contains the last polish of the new Guides section in the sidebar.


src/browser/styles/bootstrap.grid-only.min.css
```diff
@@ -7,8 +7,1097 @@
...
+ */ /*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+html {
```
jharris4, line 15: hmm, usually I'd expect minified css to be well, minified! Hard to see what was changed in this commit as well because of all the formatting changes...

jharris4:
There seem to be partially obscured tooltips on the pagination dots when you hover...

Screen Shot 2021-04-21 at 6 17 31 PM

src/browser/modules/GuideCarousel/GuideCarousel.tsx
```diff
@@ -65,15 +58,6 @@ function GuidesCarousel({
     }
   }
 
-  function onKeyUp(e: React.KeyboardEvent) {
```
jharris4, line 68: Was the reasoning for this that using left/right arrow makes sense in a frame, but not in a sidebar? There's a tiny chance people will notice the inconsistency that you can use keyboard shortcuts in the main play guides, but not the sidebar guides.

jharris4 (dismissed):
Looking good, I only found a couple of small issues with the pagination controls...

src/browser/styles/bootstrap.grid-only.min.css
OskarDamkjaer: I've remade the change without our on commit hook. It's still hard to see though. What changed is that figure{margin: 1em 40px;}. I wasn't 100% happy with the change anyway so I opted to remove the <Figure> tags instead and leave this css as is.

src/browser/modules/GuideCarousel/GuideCarousel.tsx
OskarDamkjaer: I agree, and after not getting the keyEvents to fire in this component I gave up and removed it to cut scope for the release. Should have documented that properly! Hope to get the arrow keys working again before the release but I had more pressing matters 😓

OskarDamkjaer:
image

To handle guides that have more than 25 slides gracefully we now show this, in the future we hope to have more advanced pagination

src/browser/modules/GuideCarousel/GuideCarousel.tsx
OskarDamkjaer: Feel free to give it a try, I might have missed something obvious

oskarhane (approved):
Looks good and works really well!

PR #1372, opened 2021-04-22
Respect outgoing connections canny

OskarDamkjaer:

### DIFF
--- a/src/browser/services/canny.ts
+++ b/src/browser/services/canny.ts
@@ -97,10 +97,6 @@ export class CannyLoader extends Component {
     if (canUseDOM()) {
       delete (window as any).CannyIsLoaded
       delete (window as any).Canny
-      const cannyTag = document.getElementById('canny-jssdk')
-      if (cannyTag?.outerHTML) {
-        cannyTag.outerHTML = ''
-      }
     }
   }
 
--- a/src/browser/services/canny.ts
+++ b/src/browser/services/canny.ts
@@ -18,6 +18,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { Component, ReactNode } from 'react'
+import { canUseDOM } from 'services/utils'
+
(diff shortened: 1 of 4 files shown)


src/browser/services/canny.ts
```diff
@@ -74,3 +77,34 @@ export const CannySDK = {
...
+  shouldComponentUpdate(): boolean {
+    return false
+  }
+
+  componentWillUnmount(): void {
+    if (canUseDOM()) {
+      delete (window as any).CannyIsLoaded
+      delete (window as any).Canny
+      const cannyTag = document.getElementById('canny-jssdk')
+      if (cannyTag?.outerHTML) {
```
jharris4, line 101: hmm, a little unconventional to remove a dom element this way imo, it's more normal to just do something like:

```
elem.parentNode.removeChild(elem)
```

Also, the Intercom & Segment components that work similar to this one don't seem to be removing the script element, so it might be good to make that consistent...

OskarDamkjaer: I'll leave the script tag in.

src/browser/services/canny.ts
```diff
@@ -54,7 +57,7 @@ export const CannySDK = {
           return
         }
         const f = document.getElementsByTagName('script')[0]
-        const e = document.createElement('script') as HTMLScriptElement
+        const e = document.createElement('script')
```
jharris4, line 60: It's not clear to me that the check for an element with id `canny-jssdk` will definitely prevent the script from being added more than once. We should definitely check this before merging this PR... :-)

OskarDamkjaer: I think you're right, and it seems it'd load again even if not needed from testing. However I feel it's nigh impossible to a perfect cleanup and remount of the external script from canny. The worst case of an extra load from canny is still fine I think

oskarhane (approved):
Good to see usage of existing patterns!

PR #1375, opened 2021-04-22
Add call to segment identify

OskarDamkjaer:
This PR adds a temporary ID to our segment calls that's better than nothing.
Link to the docs on how identify works: https://segment.com/docs/connections/spec/identify/


oskarhane (approved):
Nice! I trust you it's being received and used by the service. Looks good on our side.

PR #1384, opened 2021-04-26
Change test failing on 4.3

OskarDamkjaer:
With 4.3 enterprise users will be able to call dbms.cluster.overview, this in turn causes a bug in our codebase where some transaction metadata is not present in the :queries table. We'll want to fix this, but not before the 4.2.6 release.


oskarhane (approved):
Looks good, merge on green

PR #1390, opened 2021-04-26
Filter out warnings when building

OskarDamkjaer:
This will make logs easier to read and faster, the warnings still show up in vscode, so we can act on them when able.


oskarhane (approved):
🙇

PR #1398, opened 2021-04-29
Update playbook to use versionless bundle

recrwplay:
Updates the Antora playbook to use the new versionless UI bundle.
This change means that when future updates are made to the docs UI we won't have to raise a PR to update the playbook, and we will get the latest UI the next time we rebuild and publish.


AlexicaWright (approved):
LGTM and builds fine.

PR #1401, opened 2021-05-03
Fix rgb function used by mistake

OskarDamkjaer:


jharris4 (approved):
Nice catch on these two things

PR #1402, opened 2021-05-03
Bump monaco editor to remove extra scroll effect

OskarDamkjaer:
The monaco editor version we used, added vertical scroll which we removed with a workaround. This pr instead solves the problem by updating monacos version.
http://bump_monaco.surge.sh


jharris4 (approved):
LGTM 🦟 🔨

PR #1403, opened 2021-05-04
Re-add populating main editor

OskarDamkjaer:
When introducing the re-usable frame we removed to ability to populate the main editor easily. This PR adds it back as a CMD/CTRL click. I've opted to have both ctrl and cmd work. Tested to work in a windows VM.

Preview at: http://bring_to_editor.surge.sh
CleanShot 2021-05-04 at 11 39 58

### DIFF
--- a/e2e_tests/integration/commands.spec.ts
+++ b/e2e_tests/integration/commands.spec.ts
@@ -118,4 +118,23 @@ describe('Commands', () => {
       cy.wait(getCommandDelay(cmd))
     })
   })
+
+  it('can ctrl+click to re-populate the main editor', () => {
+    cy.executeCommand(':clear')
+    cy.executeCommand('return 1')
+    cy.get('[data-testid="frameCommand"]')
+      .contains('return 1')
+      .click()
+
+    cy.typeInFrame('Vermilion')
+
+    // click outside frame
+    cy.get('#monaco-main-editor').click()
+
+    cy.get('body').type('{ctrl}', { force: true, release: false })
+    cy.get('[data-testid="frameCommand"]').click()
+    cy.get('body').type('{ctrl}') // release mod
+
+    cy.get('#monaco-main-editor').contains('Vermilion')
+  })
 })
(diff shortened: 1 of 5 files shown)


oskarhane:
Sweet! Would it be possible to add a test for this so we don't introduce regressions in the future?

OskarDamkjaer:
Good call!

e2e_tests/integration/commands.spec.ts
```diff
@@ -118,4 +118,23 @@ describe('Commands', () => {
...
+    cy.get('[data-testid="frameCommand"]')
+      .contains('return 1')
+      .click()
+
+    cy.typeInFrame('Vermilion')
+
+    // click outside frame
+    cy.get('#monaco-main-editor').click()
+
+    cy.get('body').type('{ctrl}', { force: true, release: false })
```
jharris4, line 134: will this still work running tests locally on a Mac? ;-)

src/browser/modules/Frame/FrameTitlebar.tsx
```diff
@@ -256,16 +258,61 @@ function FrameTitlebar(props: FrameTitleBarProps) {
...
+        // a really quick click will lose some of the last edits
+        const editorRefVal = editorRef.current?.getValue()
+        if (editorRefVal && editorRefVal !== editorValue) {
+          setEditorValue(editorRefVal)
+        }
+        setRenderEditor(false)
+      }
+    }
+
+    document.addEventListener('mouseup', handleClickOutside)
```
jharris4, line 297: nice, the change to mouseup instead of mousedown seems to fix the bug clicking another frame when a multi-statement frame was previously being edited! :-)

src/browser/modules/Frame/FrameTitlebar.tsx
```diff
@@ -256,16 +258,61 @@ function FrameTitlebar(props: FrameTitleBarProps) {
...
+  useEffect(() => {
+    // We want clicks outside the frame itself, not just the titlebar.
+    // Because of how the component tree is built (we don't have a
+    // reference to the full frame body) we'd need to pass
+    // a ref from each parent to avoid this dom traversal
+    function handleClickOutside(event: MouseEvent) {
+      if (!(event.target instanceof Element)) {
+        return
+      }
+      const insideFrame = titleBarRef.current
```
jharris4, line 279: this is a good workaround for now, but would be nice to clean up the refs to avoid this later...

e2e_tests/integration/commands.spec.ts
OskarDamkjaer: yes

PR #1410, opened 2021-05-12
Add support for home database

jharris4:
http://home-database.surge.sh/

### DIFF
--- a/src/browser/modules/DBMSInfo/DatabaseSelector.tsx
+++ b/src/browser/modules/DBMSInfo/DatabaseSelector.tsx
@@ -36,6 +36,8 @@ const Select = styled.select`
 
 const EMPTY_OPTION = 'Select db to use'
 
+const HOUSE_EMOJI = '\u{1F3E0}'
+
 export const DatabaseSelector = ({
   databases = [],
   selectedDb = '',
@@ -59,6 +61,9 @@ export const DatabaseSelector = ({
     )
   }
   const uniqDatabases = uniqBy(databasesList, 'name')
+  const homeDb =
+    uniqDatabases.find((db: any) => db.home) ||
+    uniqDatabases.find((db: any) => db.default)
 
   return (
     <DrawerSection>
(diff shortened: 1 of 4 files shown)


src/browser/modules/DBMSInfo/DatabaseSelector.tsx
```diff
@@ -70,12 +70,16 @@ export const DatabaseSelector = ({
           onChange={selectionChange}
         >
           {uniqDatabases.map(db => {
-            const defaultStr = db.default ? ' - default' : ''
+            const defaultOrHomeStr = db.home
+              ? ' - home'
+              : db.default
+              ? ' - default'
+              : ''
```
OskarDamkjaer, line 77: I'm generally not a big fan of nested ternaries, but I think this is readable enough since it's simple

OskarDamkjaer (changes requested):
image
Sys info frame seems to be missing a header. I think it should be added in sysinfo-utils.tsx on line 140.

src/browser/modules/DBMSInfo/DatabaseSelector.tsx
jharris4: yeah, I don't like massive nested ternaries either, but as you said, this case is simple & concise enough to be ok imo.

jharris4:
@OskarDamkjaer Good catch on the missing `:sysinfo` header. I've pushed a fix, and verified that it now renders properly both on `4.3` and `4.0.2`

src/browser/modules/DBMSInfo/DatabaseSelector.tsx
```diff
@@ -59,6 +62,9 @@ export const DatabaseSelector = ({
     )
   }
   const uniqDatabases = uniqBy(databasesList, 'name')
+  const homeDb =
+    uniqDatabases.find((db: any) => db.home) ||
```
OskarDamkjaer, line 66: Assuming we don't have any neat types for this, any is fine for now

OskarDamkjaer (approved):

PR #1412, opened 2021-05-18
Replace gifs with compressed mp4

OskarDamkjaer:
This PR is part of the effort to prevent our tests from crashing from memory issues in CI. Preview @ http://reduce_gif_size.surge.sh


jharris4 (approved):
Nice! much smaller file sizes!

PR #1413, opened 2021-05-19
Attempt to fix flaky test where favorite is not created

OskarDamkjaer:
Can't reproduce the flakiness locally, but looking at the recorded video it seems the favorite is not created properly. If nothing else this change will give more information if the test is still flaky.


jharris4 (approved):
Thanks for explaining why this should help fix the test flakiness, hopefully it works!

PR #1414, opened 2021-05-19
upgrade to driver 4.3.0

jharris4:


jharris4:
@OskarDamkjaer I think we should merge this once tests pass, since version 4.3.0 of the driver has been released.

OskarDamkjaer (approved):
Seems to fail only on the plan output, and that's a different issue. Would prefer you merge with the squash button, to avoid the "wow vscode won't let me commit this" commits :)

PR #1415, opened 2021-05-20
Make sure timeout is cleared properly

OskarDamkjaer:
This is a small PR to prevent the waiting banner from showing even though you're already connected in sandbox. I've added some types and some minor refactoring as well. The problem was not clearing setTimeout properly.

### DIFF
--- a/src/browser/modules/Main/Main.tsx
+++ b/src/browser/modules/Main/Main.tsx
@@ -75,11 +75,7 @@ const Main = React.memo(function Main(props: any) {
           Connection to server lost. Reconnecting...
         </WarningBanner>
       </Render>
-      <Render
-        if={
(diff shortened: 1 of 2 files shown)


jexp:
Works now for me, no yellow bars anymore.

jharris4:
@OskarDamkjaer I think we should re-target this to the 4.3-release branch...

OskarDamkjaer:
I'll rebase!

jharris4:
The changes look good to me! I'd just like to be able to reproduce the issue this fixes if possible, so I can verify the fix...

jharris4 (approved):
This one was impossible to test, but I can see that the change definitely fixes some issues with the initial implementation (missing deps on useEffect hook) so let's merge it!

PR #1419, opened 2021-05-26
add support for browser.retain_editor_history setting

jharris4:
also change local defaults for community edition

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaDuck.ts
+++ b/src/shared/modules/dbMeta/dbMetaDuck.ts
@@ -142,7 +142,7 @@ export const versionHasEditorHistorySetting = (version: string) =>
   semver.gte(version, VERSION_FOR_EDITOR_HISTORY_SETTING)
 
 export const supportsEditorHistorySetting = (state: any) =>
-  versionHasEditorHistorySetting(getVersion(state))
+  isEnterprise(state) && versionHasEditorHistorySetting(getVersion(state))
 
 export const shouldAllowOutgoingConnections = (state: any) =>
   !isEnterprise(state) || getAllowOutgoingConnections(state)
@@ -151,9 +151,7 @@ export const shouldRetainConnectionCredentials = (state: any) =>
   !isEnterprise(state) || getRetainConnectionCredentials(state)
 
 export const shouldRetainEditorHistory = (state: any) =>
-  !isEnterprise(state) ||
-  !supportsEditorHistorySetting(state) ||
-  getRetainEditorHistory(state)
+  !supportsEditorHistorySetting(state) || getRetainEditorHistory(state)
 
 function updateMetaForContext(state: any, meta: any, context: any) {
   if (!meta || !meta.records || !meta.records.length) {
--- a/src/shared/modules/dbMeta/dbMetaDuck.ts
+++ b/src/shared/modules/dbMeta/dbMetaDuck.ts
@@ -20,6 +20,7 @@
 
 import neo4j from 'neo4j-driver'
 import Rx from 'rxjs/Rx'
+import semver from 'semver'
 import bolt from 'services/bolt/bolt'
 import { isConfigValFalsy } from 'services/bolt/boltHelpers'
 import { APP_START } from 'shared/modules/app/appDuck'
@@ -59,6 +60,7 @@ import {
   isACausalCluster,
   setClientConfig
 } from '../features/featuresDuck'
+import { clearHistory } from 'shared/modules/history/historyDuck'
 
 export const NAME = 'meta'
 export const UPDATE = 'meta/UPDATE'
@@ -106,7 +108,7 @@ export const getStoreId = (state: any) =>
 
 export const getAvailableSettings = (state: any) =>
   (state[NAME] || initialState).settings
-export const allowOutgoingConnections = (state: any) =>
+export const getAllowOutgoingConnections = (state: any) =>
   getAvailableSettings(state)['browser.allow_outgoing_connections']
 export const credentialsTimeout = (state: any) =>
   getAvailableSettings(state)['browser.credential_timeout'] || 0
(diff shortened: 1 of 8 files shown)


src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -128,6 +136,25 @@ export const getActiveDbName = (state: any) =>
...
+
+export const shouldAllowOutgoingConnections = (state: any) =>
+  !isEnterprise(state) || getAllowOutgoingConnections(state)
+
+export const shouldRetainConnectionCredentials = (state: any) =>
+  !isEnterprise(state) || getRetainConnectionCredentials(state)
+
+export const shouldRetainEditorHistory = (state: any) =>
+  !isEnterprise(state) ||
+  !supportsEditorHistorySetting(state) ||
```
OskarDamkjaer, line 155: If the setting is enterprise only, then it makes sense to more the enterprise check inside the supports editor history setting imo since community by definition doesn't support the setting

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -216,7 +243,8 @@ export const initialState = {
   settings: {
     'browser.allow_outgoing_connections': false,
     'browser.remote_content_hostname_allowlist': 'guides.neo4j.com, localhost',
-    'browser.retain_connection_credentials': false
+    'browser.retain_connection_credentials': false,
+    'browser.retain_editor_history': false
```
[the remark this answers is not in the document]
OskarDamkjaer: Worth noting that the !isEnterprise stuff defaults to true since edition default/before connect is null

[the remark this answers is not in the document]
OskarDamkjaer: Perhaps it's better to explicitly look for "community" rather than not enterprise?

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -605,6 +633,13 @@ export const serverConfigEpic = (some$: any, store: any) =>
               }
               store.dispatch(setRetainCredentials(retainCredentials))
               value = retainCredentials
+            } else if (name === 'browser.retain_editor_history') {
+              let retainHistory = true
```
OskarDamkjaer, line 637: here we seem to default to true? But not in the intial state?

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -623,6 +658,12 @@ export const serverConfigEpic = (some$: any, store: any) =>
             updateUserCapability(USER_CAPABILITIES.serverConfigReadable, true)
           )
           store.dispatch(updateSettings(settings))
+          if (
+            supportsEditorHistorySetting(store.getState()) &&
+            !settings['browser.retain_editor_history']
+          ) {
+            store.dispatch(clearHistory())
```
OskarDamkjaer, line 665: Would this not always clear history on community version 4.3?

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -128,6 +136,25 @@ export const getActiveDbName = (state: any) =>
...
+  semver.gte(version, VERSION_FOR_EDITOR_HISTORY_SETTING)
+
+export const supportsEditorHistorySetting = (state: any) =>
+  versionHasEditorHistorySetting(getVersion(state))
+
+export const shouldAllowOutgoingConnections = (state: any) =>
+  !isEnterprise(state) || getAllowOutgoingConnections(state)
+
+export const shouldRetainConnectionCredentials = (state: any) =>
+  !isEnterprise(state) || getRetainConnectionCredentials(state)
```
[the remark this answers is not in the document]
jharris4: The only settings provided by community are:

```
browser.post_connect_cmd(pin):""
browser.remote_content_hostname_whitelist(pin):"guides.neo4j.com,localhost"
```

src/shared/modules/dbMeta/dbMetaDuck.ts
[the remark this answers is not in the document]
jharris4: The server does default this setting to true, but for safety it was decided to default to false on the client side.

src/shared/modules/dbMeta/dbMetaDuck.ts
jharris4: Interesting idea about moving the check inside the supports function, I'll have a look at that.

src/shared/modules/dbMeta/dbMetaDuck.ts
jharris4: ah good catch, looks like you're right, I'll look into fixing this.

src/shared/modules/dbMeta/dbMetaDuck.ts
jharris4: I copied the logic from `retain_connection_credentials` and it doesn't default to true per se. It only stays true if the config exists and is not falsy...

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -128,6 +136,23 @@ export const getActiveDbName = (state: any) =>
  * Helpers
  */
 
+export const VERSION_FOR_EDITOR_HISTORY_SETTING = '4.3.0'
```
OskarDamkjaer, line 139: does this need to be exported

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -128,6 +136,23 @@ export const getActiveDbName = (state: any) =>
...
+
+export const versionHasEditorHistorySetting = (version: string) =>
+  semver.gte(version, VERSION_FOR_EDITOR_HISTORY_SETTING)
+
+export const supportsEditorHistorySetting = (state: any) =>
+  isEnterprise(state) && versionHasEditorHistorySetting(getVersion(state))
+
+export const shouldAllowOutgoingConnections = (state: any) =>
+  !isEnterprise(state) || getAllowOutgoingConnections(state)
```
OskarDamkjaer, line 148: I think this would send outgoing connections before we connect to a database? I think that's a no-go in airgapped deployments

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -128,6 +136,23 @@ export const getActiveDbName = (state: any) =>
...
+
+export const shouldRetainConnectionCredentials = (state: any) =>
+  !isEnterprise(state) || getRetainConnectionCredentials(state)
+
+export const shouldRetainEditorHistory = (state: any) =>
```
OskarDamkjaer, line 153: it's a nit but could we move this to by be the other history getters? (so to line 146). Feel free to disregard this comment, just find it slightly more readable

src/shared/modules/dbMeta/dbMetaDuck.ts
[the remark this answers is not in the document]
OskarDamkjaer: Canny and Sentry both check allow shouldAllowOutgoing connections before connecting

src/shared/modules/dbMeta/dbMetaDuck.ts
OskarDamkjaer: Any reason we don't use the "shouldRetainEditorHistory" getter?

OskarDamkjaer (changes requested):
I the current impl would do outgoing connections before connecting

OskarDamkjaer (approved):
There might be more places where we want to check hasEdition before checking isEnterprise in the future, this looks good for the release

PR #1425, opened 2021-06-02
fix clear history happening too often when setting is false

jharris4:
also clear history on disconnect from server when settings is false

also improve disconnect handling

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaDuck.ts
+++ b/src/shared/modules/dbMeta/dbMetaDuck.ts
@@ -27,8 +27,11 @@ import { APP_START } from 'shared/modules/app/appDuck'
 import {
   CONNECTED_STATE,
   CONNECTION_SUCCESS,
+  SWITCH_CONNECTION,
+  SWITCH_CONNECTION_SUCCESS,
   connectionLossFilter,
   DISCONNECTION_SUCCESS,
+  SILENT_DISCONNECT,
   LOST_CONNECTION,
   UPDATE_CONNECTION_STATE,
   setRetainCredentials,
@@ -531,6 +534,7 @@ export const dbMetaEpic = (some$: any, store: any) =>
     .ofType(UPDATE_CONNECTION_STATE)
     .filter((s: any) => s.state === CONNECTED_STATE)
     .merge(some$.ofType(CONNECTION_SUCCESS))
+    .merge(some$.ofType(SWITCH_CONNECTION_SUCCESS))
     .mergeMap(() => {
       return (
         Rx.Observable.timer(1, 20000)
@@ -551,6 +555,7 @@ export const dbMetaEpic = (some$: any, store: any) =>
               .ofType(LOST_CONNECTION)
               .filter(connectionLossFilter)
               .merge(some$.ofType(DISCONNECTION_SUCCESS))
+              .merge(some$.ofType(SILENT_DISCONNECT))
           )
           .mergeMap(() => switchToRequestedDb(store))
           .mapTo({ type: DB_META_DONE })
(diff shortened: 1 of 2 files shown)


src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -531,6 +534,7 @@ export const dbMetaEpic = (some$: any, store: any) =>
     .ofType(UPDATE_CONNECTION_STATE)
     .filter((s: any) => s.state === CONNECTED_STATE)
     .merge(some$.ofType(CONNECTION_SUCCESS))
+    .merge(some$.ofType(SWITCH_CONNECTION_SUCCESS))
```
oskarhane, line 537: Is this really needed?
I _think_ it's already triggered by the lines right above by the connection state going into `CONNECTED_STATE`.

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -551,6 +555,7 @@ export const dbMetaEpic = (some$: any, store: any) =>
               .ofType(LOST_CONNECTION)
               .filter(connectionLossFilter)
               .merge(some$.ofType(DISCONNECTION_SUCCESS))
+              .merge(some$.ofType(SILENT_DISCONNECT))
```
oskarhane, line 558: Good catch, just make sure it works in desktop (which it should do).

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -658,9 +663,15 @@ export const serverConfigEpic = (some$: any, store: any) =>
...
+            store.getState()
+          )
           store.dispatch(updateSettings(settings))
-          // settings must be updated in state for this check to work
-          if (!shouldRetainEditorHistory(store.getState())) {
+          const shouldRetainHistory = shouldRetainEditorHistory(
+            store.getState()
+          )
+          // if setting changed either prev/next was false so clear history
+          if (prevShouldRetainHistory !== shouldRetainHistory) {
```
oskarhane, line 674: It's not clear to me if we should clear the history here? 
Move it in / out of local storage, yes.
Remove it from runtime state, why?

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -699,5 +710,14 @@ export const serverInfoEpic = (some$: any, store: any) =>
     })
     .mapTo({ type: 'NOOP' })
 
-export const clearMetaOnDisconnectEpic = (some$: any, _store: any) =>
-  some$.ofType(DISCONNECTION_SUCCESS).mapTo({ type: CLEAR })
+export const clearMetaOnDisconnectEpic = (some$: any, store: any) =>
+  some$
+    .ofType(DISCONNECTION_SUCCESS)
+    .merge(some$.ofType(SWITCH_CONNECTION))
+    .merge(some$.ofType(SILENT_DISCONNECT))
```
oskarhane, line 717: This already happens in https://github.com/neo4j/neo4j-browser/blob/f76ff9c731f62ecb606d715872d6aed96f273885/src/shared/modules/connections/connectionsDuck.ts#L581-L588 so maybe create a separate epic just to handle this case so we don't get any unexpected side effects.

oskarhane (commented):
A bit too many changes for my taste: pulling in and listening on new action types in dbMetaEpic and listening on new events on the clearMetaOnDisconnect epic.
That's a bit risky in this situation imo.

I've commented a bit and real only real question is if we really should clear the history in other cases than when we're switching connection?

src/shared/modules/dbMeta/dbMetaDuck.ts
jharris4: This is a fail safe for when the connection is switched to a different server without the connection events.

src/shared/modules/dbMeta/dbMetaDuck.ts
jharris4: From our conversation earlier I thought they were separate & distinct events. I guess the best way to clear up that question is by performing the actions in desktop that trigger it.

OskarDamkjaer (approved):
Changes look good to me, thanks for changing to the .do operator and removing unused imports as per my comments.

oskarhane (approved):
LGTM

oskarhane (approved):
👯

PR #1427, opened 2021-06-07
Frame Refresh

OskarDamkjaer:
Preview at http://title_bar.surge.sh
image

Changelist: 
- Adds yarn script for easier updating of snapshots
- Splits FrameTitlebar into three compontents, the editor, the export button and the titlebar  
- Replaces a few icons with svg versions from liza
- Switches to a new color scheme 
- Adds a few settings to monaco to prevent highlighting matched brackets in the gutter 
- Renames EditorFrame to MainEditor avoid confusion


src/browser/components/buttons/index.tsx
```diff
@@ -314,28 +314,86 @@ const StyledSelectedCypherFrameButton = styled(StyledCypherFrameButton)`
...
+  line-height: 20px;
+  display: inline-block;
+  margin-left: 12px;
+  &:hover {
+    background-color: ${props => props.theme.frameButtonHoverBackground};
+    color: ${props => props.theme.secondaryButtonTextHover};
+    fill: ${props => props.theme.secondaryButtonTextHover};
+    text-decoration: none;
+  }
+  ${props =>
```
jharris4, line 374: I have mixed feelings about the readability of this, and whether it was clearer with the original distinction with 2 separate styled components. but so long as this syntax works properly I guess it's fine.

src/browser/modules/D3Visualization/components/styled.tsx
```diff
@@ -24,28 +24,14 @@ export const legendRowHeight = 32
...
-}
-
 export const StyledSvgWrapper = styled.div`
   line-height: 0;
   height: 100%;
   position: relative;
   > svg {
     height: 100%;
     width: 100%;
-    background-color: ${props => getColor(props.theme.name, 'svgBackground')};
```
jharris4, line 48: why are we changing the logic of the colour of the svg background?

src/browser/modules/Editor/Monaco.tsx
```diff
@@ -171,7 +171,9 @@ const Monaco = forwardRef<MonacoHandles, MonacoProps>(
           selectionHighlight: false,
           value,
           wordWrap: 'on',
-          wrappingStrategy: 'advanced'
+          wrappingStrategy: 'advanced',
```
jharris4, line 174: A little too much scope creep to be changing monaco settings for a PR about buttons imo, let's try to do less of this in the future... :-)

src/browser/modules/User/UserList.test.tsx
```diff
@@ -29,8 +29,9 @@ import {
   listUsersQuery
 } from 'shared/modules/cypher/boltUserHelper'
 
-// Stubbing out title bar as it depends on store
+// Stubbing out title bar & template as they depend on store
 jest.mock('browser/modules/Frame/FrameTitlebar', () => () => null)
+jest.mock('browser/modules/Frame/FrameTemplate', () => () => null)
```
jharris4, line 34: Since you seem to have removed the connect() from the FrameTemplate, this isn't needed anymore?

src/browser/components/buttons/index.tsx
OskarDamkjaer: I think it's a lot more readable to write:
  `background-color: ${props => props.pressed ? props.theme.frameButtonHoverBackground} : 'transparent'}`
 
 Than override in the way I've done here.

src/browser/modules/Editor/Monaco.tsx
OskarDamkjaer: Yeah, it changed from just buttons to a full UI-refresh.

src/browser/modules/User/UserList.test.tsx
OskarDamkjaer: Yes but it renders FrameEditor which is connected. So either the template or the Editor needs to be stubbed.

jharris4 (approved):
LGTM

PR #1430, opened 2021-06-14
guide command

martin-neotech:
4.3 now has the guide command.
Also bumped the version.

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,13 +1,13 @@
 name: browser-manual
 title: Neo4j Browser
-version: '4.2'
+version: '4.3'
 start_page: ROOT:index.adoc
 nav:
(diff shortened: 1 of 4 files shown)


renetapopova (commented):
Looks good! Thanks, @martin-neotech. Some minor comments inline.

renetapopova (approved):
LGTM!

gregoryking (commented):
Hey Reneta, do a quick view of all the thing's you've updated to have :guide commands.  Only the ones in the sidebar have been migrated to the new format I believe, not the full list you have moved over.  For example write-code doesn't work with :guide. Check in with @OskarDamkjaer if you're curious for reasons. @OskarDamkjaer Is it feasible for docs team to help us migrate that content to :guide?

PR #1431, opened 2021-06-14
More guide polish

OskarDamkjaer:
Moves padding that was on the Guide Sidebar into our built in guides. This way we don't get misaligned padding on ascidoc guides. Also changes color of next buttons in pagination.
http://more_guide_polish.surge.sh/


jharris4 (approved):
🎨 👍

PR #1432, opened 2021-06-14
Fix label issue and bump driver

OskarDamkjaer:
Preview: http://fix_multitstatement_label.surge.sh
Before: 
bild

After:
bild


src/browser/modules/Stream/CypherScriptFrame/CypherScriptFrame.tsx
```diff
@@ -72,7 +72,7 @@ function CypherScriptFrame({
                       data-testid="multi-statement-list-title"
                       {...titleProps}
                     >
-                      <PointerFrameCommand selectedDb={frames[id].cmd}>
+                      <PointerFrameCommand selectedDb={frames[id].useDb || ''}>
```
jharris4, line 75: Nice find, is the precedent to use empty string vs null/undefined when there is no db?

NOTICE.txt
```diff
@@ -223,7 +223,15 @@ Third-party licenses
 │  ├─ long@4.0.0
 │  │  ├─ URL: https://github.com/dcodeIO/long.js.git
 │  │  └─ VendorName: Daniel Wirtz
-│  ├─ neo4j-driver@4.2.2
+│  ├─ neo4j-driver-bolt-connection@4.3.1
```
jharris4, line 226: Nice, licenses update seems to have been missed with the last driver upgrade

jharris4 (approved):
LGTM

src/browser/modules/Stream/CypherScriptFrame/CypherScriptFrame.tsx
OskarDamkjaer: Fixed this in another commit

jharris4 (approved):

PR #1437, opened 2021-06-16
Focus reusable editor on mounting it

OskarDamkjaer:
Preview: http://focus_editor_on_click.surge.sh

Focuses editor and sets focus to the end of the statement.
https://user-images.githubusercontent.com/10564538/122192901-5a46c080-ce94-11eb-8dd7-556e2c1614bd.mp4


jharris4 (approved):
Works really well. I noticed that the connect frame doesn't auto-focus anymore, but that seems to be a regression from an earlier version, so we can fix that separately.

PR #1440, opened 2021-06-16
Fix bug where button shrunk in safari

OskarDamkjaer:
Min item width is not respected by default in safari.
Preview @ http://safari_button_shrink_bug.surge.sh

Before:
CleanShot 2021-06-16 at 17 28 44

After:
CleanShot 2021-06-16 at 17 29 18


jharris4:
Nice find on the solution for this! https://stackoverflow.com/questions/32689686/overlapping-css-flexbox-items-in-safari

jharris4 (approved):
LGTM

PR #1441, opened 2021-06-17
Remove unused style components and exports

OskarDamkjaer:
I know we have lots of unused exports in our code base, and found [ts-prune](https://github.com/nadeesha/ts-prune) to help us find unused imports. There are still exports that aren't used, but this PR includes all the unused styled components.

No preview as there's no changes to the app.


jharris4 (approved):
Great cleanup, thanks for doing this!

PR #1442, opened 2021-06-17
Minor code cleanups

OskarDamkjaer:
This is a refactoring PR that takes a stab at bringing our warnings down from the thousands. Best reviewed Commit by Commit to only see changes relevant for each step (thought the changes were too small for 4 PRs)

On updating the styled components, we passed some incorrect props that we're discovered when removing the 'any' which I've also fixed (mostly unused props and numbers passed as strings)

### DIFF
--- a/src/shared/modules/cypher/cypherDuck.ts
+++ b/src/shared/modules/cypher/cypherDuck.ts
@@ -87,7 +87,7 @@ const queryAndResolve = async (
         })
   })
 }
-const callClusterMember = async (connection: any, action: any, _store: any) => {
+const callClusterMember = async (connection: any, action: any) => {
   return new Promise(resolve => {
     bolt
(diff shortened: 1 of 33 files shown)


src/shared/modules/cypher/cypherDuck.ts
```diff
@@ -256,7 +256,7 @@ export const handleForcePasswordChangeEpic = (some$: any, store: any) =>
       ) => {
         if (!action.$$responseChannel) return Rx.Observable.of(null)
 
-        return new Promise((resolve, _reject) => {
+        return new Promise(resolve => {
```
jharris4, line 259: a little concerning (maybe?) that we're not handling the error here, but that's out of scope for this PR

jharris4 (approved):
moar nice cleanup 👍

PR #1443, opened 2021-06-17
add new playbook to demo edit this page link

recrwplay:
Creates a new Antora playbook that builds from the github url rather than local content.

Advantages of this are

- VCS checkout no longer required as part of build
- this repo is public, so we can add an 'Edit this page' link by setting `page-origin-private: false` or removing the attribute.

To test, use `$(npm bin)/antora publish.yml --stacktrace --fetch`

### DIFF
--- a/docs/publish.yml
+++ b/docs/publish.yml
@@ -1,0 +1,48 @@
+site:
+  title: Neo4j Docs
+  url: https://neo4j.com/docs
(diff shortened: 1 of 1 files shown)


OskarDamkjaer:
@recrwplay what's the state of this PR?

renetapopova (approved):
LGTM! Thanks, @recrwplay!

PR #1444, opened 2021-06-18
Remove edit frame

OskarDamkjaer:
We put the :edit command in to test the monaco editor in production without doing the full switch with [this PR](https://github.com/neo4j/neo4j-browser/pull/1213/files). Now we've fully moved over to monaco and no longer need it. I've also updated the docs accordingly.

Preview @ http://remove_edit_frame.surge.sh


jharris4 (approved):
Good that you updated the docs too

PR #1445, opened 2021-06-18
Sensible cypher frame

OskarDamkjaer:
- Renames the cypherframe to be searchable (so not called index.ts).
- Adds proptypes to components related to the cypher result frame
- No longer passes unneeded props
- No longer spreads state and binds setstate to children

No visible changes so no preview.


jharris4 (approved):
More nice cleanup 👍

Nice that we'll have the whole dev phase to test these changes too in case some of those props/state that didn't appear to be used were triggering re-renders etc.

PR #1447, opened 2021-06-18
Monaco class comp

OskarDamkjaer:
This PR aims to refactor our Monaco component to be a class rather than function component and preserve the same functionality. This might seem odd at first glance, but for dealing with a mutable, external library, I've found the class component approach works better, since it allows for handling mutable state.

### DIFF
--- a/src/browser/modules/Editor/Monaco.tsx
+++ b/src/browser/modules/Editor/Monaco.tsx
@@ -23,17 +23,11 @@ import { parse, QueryOrCommand } from 'cypher-editor-support'
 import { debounce } from 'lodash-es'
 import {
   editor,
-  IPosition,
   KeyCode,
   KeyMod,
   MarkerSeverity
 } from 'monaco-editor/esm/vs/editor/editor.api'
-import React, {
-  forwardRef,
-  useEffect,
-  useImperativeHandle,
-  useRef
-} from 'react'
+import React from 'react'
 import ResizeObserver from 'resize-observer-polyfill'
 import styled from 'styled-components'
 import { Bus } from 'suber'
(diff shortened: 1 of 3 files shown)


src/browser/modules/Editor/Monaco.tsx
```diff
@@ -76,410 +65,380 @@ const MonacoStyleWrapper = styled.div`
...
-        KeyMod.CtrlCmd | KeyCode.US_DOT,
-        onDisplayHelpKeys
-      )
-      editorRef.current.addCommand(
-        KeyCode.Escape,
-        toggleFullscreen,
-        '!suggestWidgetVisible && !findWidgetVisible'
-      )
+class Monaco extends React.Component<MonacoProps, MonacoState> {
+  state: MonacoState = { currentHistoryIndex: 0, draft: '' }
```
jharris4, line 87: currentHistoryIndex should start with -1

src/browser/modules/Editor/Monaco.tsx
```diff
@@ -76,410 +65,380 @@ const MonacoStyleWrapper = styled.div`
...
+    this.editor.layout({
+      height,
+      width: this.container.offsetWidth
+    })
+  }
 
-      isFullscreenRef.current = fillContainer
+  getValue = (): string => this.editor?.getValue() || ''
+  setValue = (value: string): void => {
+    this.setState({ currentHistoryIndex: -1 })
```
jharris4, line 143: Kind of makes me think we should make -1 a constant, something like `UNSET_HISTORY`

src/browser/modules/Editor/Monaco.tsx
```diff
@@ -76,410 +65,379 @@ const MonacoStyleWrapper = styled.div`
...
+    this.editor.addCommand(
+      KeyCode.Enter,
+      () => {
+        this.isMultiLine() ? this.newLine() : this.execute()
+      },
+      '!suggestWidgetVisible && !findWidgetVisible'
+    )
+    this.editor.addCommand(
+      KeyCode.UpArrow,
+      this.handleUp,
```
jharris4, line 351: It'd be nice to have more consistent naming of the command callbacks, but that's a nit.

jharris4 (approved):
Great that you got rid of all the hook-induced mess :-D

PR #1449, opened 2021-06-22
Improve data on use of built-in guides

OskarDamkjaer:
To know when we can migrate of the :play command, we'll need to know more about it's usage. (Right now we don't know how many of the :play uses are the automatic :play start). And for the future it's good to know the adoption of new guides we add.

### DIFF
--- a/src/shared/modules/udc/udcDuck.ts
+++ b/src/shared/modules/udc/udcDuck.ts
@@ -67,6 +67,7 @@ import { shouldTriggerConnectEvent, getTodayDate } from './udcHelpers'
 import api from 'services/intercom'
 import cmdHelper from 'shared/services/commandInterpreterHelper'
 import { extractStatementsFromString } from 'services/commandUtils'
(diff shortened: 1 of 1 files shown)


src/shared/modules/udc/udcDuck.ts
```diff
@@ -378,6 +379,30 @@ export const trackCommandUsageEpic: Epic<Action, GlobalState> = action$ =>
 
     const type = cmdHelper.interpret(action.cmd.slice(1))?.name
 
+    if (type === 'play') {
+      const guideName = action.cmd.substr(':play'.length).trim()
+
+      const content = isPlayChapter(guideName) ? 'built-in' : 'non-built-in'
+
+      return metricsEvent({
```
jharris4, line 387: Would be nice to remove the extra returns here, and just use an extra `extraDataProps = {}` that you spread into the last return...

jharris4 (approved):
Code is nice a concise now, and easy to read! nice work.

Issue #1418, opened 2021-05-25
Query editor cursor end of line


JohnWE:
### Description
It is hard for me to navigate the query editor window with the arrow keys.  

### Reproduce Steps
Take this example:

![Screenshot 2021-05-25 112024](https://user-images.githubusercontent.com/2379622/119533206-3d164a80-bd4b-11eb-9bd1-17b970803912.jpg)

I want to edit the company field.  My cursor is at the end of line marked 3.

Press Up Arrow
Cursor is now at the middle of line marked 2.  (:led_to)
Press Down Arrow
Cursor is now at the end of line marked 3.

We're skipping over the company line?  I can't get to it.

Press Home
Cursor is at the start of line marked 3.
Press Up Arrow
Cursor is at the start of line marked 2.

Still can't get to the company line.

### Expected Behavior
I would expect, since there are four lines of text here, pressing up would go move to the previous line, and not skip the company line.

### Your Environment

| software name  | version
| ---------------|----------
|                |  
| OS             |  Windows
| Web browser    |  
| node.js        | 
| npm            |
| yarn           |
| Neo4j Browser  | 1.4.5
| Neo4j          | 4.1.1

OskarDamkjaer:
While I agree it'd feel more natural to navigate the extra "fake" line created by the text editor wrapping like you suggested, I've not found a good way to accomplish it in the editor. To avoid the issue you can add more newlines to your query, to avoid the overflow. If I find a good way to solve it in the future, I'll post an update here. 

If you think this is an issue you think we should prioritize, you open a feature request on [canny](https://neo4j-browser.canny.io/feature-requests) and then we'll get a better feeling for how it compares to other feature requests.

PR #1450, opened 2021-06-22
Improve movement in wrapped lines

OskarDamkjaer:
After a lot of digging through the VSCode/monaco source code I finally found what command the DownArrow fires here: 

https://github.com/microsoft/vscode/blob/94c9ea46838a9a619aeafb7e8afd1170c967bb55/src/vs/editor/browser/controller/coreCommands.ts#L749

Before:
https://user-images.githubusercontent.com/10564538/122933152-a8653380-d36e-11eb-89e2-8ac3dab95276.mp4

After: 
https://user-images.githubusercontent.com/10564538/122933106-9edbcb80-d36e-11eb-9363-10d111b8eb48.mp4

Preview @ http://improve_monaco_navigation.surge.sh

Solves #1418

### DIFF
--- a/src/browser/modules/Editor/Monaco.tsx
+++ b/src/browser/modules/Editor/Monaco.tsx
@@ -308,10 +308,7 @@ const Monaco = forwardRef<MonacoHandles, MonacoProps>(
 
     const handleUp = () => {
       if (isMultiLine()) {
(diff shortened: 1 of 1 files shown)


jharris4 (approved):
Really great find! key up&down work much nicer now, especially on wrapped lines!

PR #1451, opened 2021-06-22
Fix plan view not updating

OskarDamkjaer:
When reusing a frame for profiling, the react lifecycle methods were written so that the component didn't update properly when it received new props. Meaning the db-hits numbers in PlanStatusBar didn't refresh as a new query was profiled.

Before:
https://user-images.githubusercontent.com/10564538/122965716-ccce0980-d388-11eb-8032-8a5d54c45505.mp4

After:
https://user-images.githubusercontent.com/10564538/122965400-6fd25380-d388-11eb-8876-7967d8a6b13a.mp4

This PR is dependent on #1445. The change intended for review is only the last commit. Preview at http://fix_plan_view_not_updating.surge.sh


OskarDamkjaer:
Closing in favour of #1455

PR #1453, opened 2021-06-24
Improve table view

OskarDamkjaer:
This PR solves the issues outlines in #584 by setting the max width to 100% and creating a "Copy to Clipboard" component to do the copying.

[Clipboard writeText](https://caniuse.com/?search=Clipboard%20API%3A%20writeText) is widely supported, but only works on HTTPS. So I've used an old workaround. 

Preview @ http://table-max-width.surge.sh


src/browser/components/ClipboardCopier.tsx
```diff
@@ -0,0 +1,76 @@
...
+  bottom: 0;
+  border-radius: 2px;
+  background-color: ${props => props.theme.frameSidebarBackground};
+  box-shadow: ${props => props.theme.standardShadow};
+  color: ${props => props.theme.primaryText}
+  font-family: ${props => props.theme.drawerHeaderFontFamily};
+  padding: 0 5px;
+`
+
+export function copyToClipboard(text: string): Promise<void> {
```
jharris4, line 55: nice util function, as you said it might be good to eventually move this to a separate utils file or something but for now it's fine to keep it here.

jharris4 (approved):
LGTM

Issue #1454, opened 2021-06-28
Profiling stats are wrong after first execution


alexabidri:
This is a regression observed some months ago.

### Reproduce Steps
Write a first query with profiling assuming the property `name` is indexed

`PROFILE MATCH (n:Person { name: "superman"}) RETURN n`

Then modify the query for another field non indexed

`PROFILE MATCH (n:Person { title: "superman"}) RETURN n`

You will see at the bottom that the statistics of the profiling didnt change, it keep displaying the one from the first query.

### Expected Behavior
I should see the profiling statistics of the last query I have executed in the window

### Current Behavior

I see the profiling statistics of the very first time I executed the query.

### Can you propose a solution?
**Manual quick win**: Launch a query in a new window

**Fix**: update the data at the bottom of the window

PR => https://github.com/neo4j/neo4j-browser/pull/1455

### Your Environment

| software name  | version
| ---------------|----------
|                |  
| OS             |  macOS Big 11.4
| Web browser    |  
| node.js        | 
| npm            |
| yarn           |
| Neo4j Browser  | 1.3.11
| Neo4j          | 4.2.4

OskarDamkjaer:
Closing this as we merged your PR

PR #1455, opened 2021-06-28
Refresh plan statistics after re-run the query

alexabidri:
Problem has been explained here -> https://github.com/neo4j/neo4j-browser/issues/1454

The issue is to refresh the state of the react component that display the plan stats at the bottom of the window

Here is a little video of the fix, sorry for the bad quality, you can see the stats are updating at the bottom of the query window
ezgif com-gif-maker (1)

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,6 +181,7 @@
     "jsonic": "^0.3.0",
     "jszip": "^3.2.2",
     "lodash-es": "^4.17.15",
+    "memoize-one": "^5.2.1",
     "monaco-editor": "0.23.0",
     "neo4j-driver": "^4.3.1",
     "react": "^16.9.0",
--- a/src/browser/modules/Stream/CypherFrame/PlanView.tsx
+++ b/src/browser/modules/Stream/CypherFrame/PlanView.tsx
@@ -161,69 +161,56 @@ export class PlanView extends Component<PlanViewProps, PlanViewState> {
   }
 }
 
-type PlanStatusbarState = { extractedPlan: any }
 type PlanStatusbarProps = {
   result: any
   setPlanExpand: (p: PlanExpand) => void
 }
 
-export class PlanStatusbar extends PureComponent<
-  PlanStatusbarProps,
-  PlanStatusbarState
-> {
-  state: PlanStatusbarState = {
-    extractedPlan: null
-  }
+export function PlanStatusbar(props: PlanStatusbarProps) {
+  const extractMemoizedPlan = memoize(
+    result => bolt.extractPlan(result, true),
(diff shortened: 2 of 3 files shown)


OskarDamkjaer (changes requested):
Hi! Thanks for both reporting and fixing this issue! We're already aware of it and I put a fix up in #1451, but I'm happy to include your fix instead. 

The bug snuck in as we didn't want to run bolt.extractPlan unless it was needed, as you noted this was done in a way so that only ever ran once. With your suggested fix, it'll be recalculated every time the component re-renders, which would be less performant but not an issue for such a simple component. However it seems to me that we no longer need the componentDidMount and the componentDidUpdate lifecycle methods when we have derivedStateFromprops?

Also for the tests to run properly and for us to be able to merge your change you'll need to sign our [CLA](https://neo4j.com/developer/cla/).

OskarDamkjaer:
To prevent the performance hit, we could add [memoize-one](https://github.com/alexreardon/memoize-one#custom-equality-function) with a custom equality function using the check from shouldComponentUpdate. What do you think @alexabidri?

alexabidri:
@OskarDamkjaer Thanks for your comment and happy to consider my fix!
Will take into account and re-test the flow 👍

OskarDamkjaer:
There's a manual step to whitelist contributors, didn't see your email but I've added you now!

I had a look at your code, and the memorization is not quite there yet, it'll still run extractPlan when it's not needed. For it to only rerun when we the summary changes we'll need to do something like this (custom equality function): 

```javascript 
const extractMemoizedPlan = memoize(
  result => bolt.extractPlan(result, true),
  (a: any, b: any) => deepEquals(a[0]?.result?.summary, b[0]?.result?.summary)
)
```

Reading your changes I found we also no longer need to keep the plan in state, so we could just do something like this in the render():
```javascript
const { result } = this.props
if (!result || !result.summary) return null

const plan = extractMemoizedPlan(result)
if (!plan) return null

return (<StyledOneRowStatsBar>//... and so on
```

Are you happy to update the memo function?

alexabidri:
You are right, will be even better indeed. I took your feedback into account and also passed the component as a pure component for readability and understandability. Thanks

OskarDamkjaer:
Nice cleanup, thanks a lot! I'll merge when the tests come back green 🎉

OskarDamkjaer (approved):
Thanks for this! 💯

OskarDamkjaer:
If you want to use your changes before they're released (some time in august), master is deployed to https://browser-canary.graphapp.io/ a few hours after merging. So you should be able to use it from there tomorrow (not guaranteed to be 100% stable though 😅 )

PR #1458, opened 2021-07-01
Add hash to filenames of webpack split/chunk bundles

jharris4:


OskarDamkjaer (approved):
Surprised to see we didn't already have this, good to see it fixed. Worked, well in my testing. Let's merge on green! 🎉

OskarDamkjaer:
Feel free to include the boltWorker changes we paired on as well! 💯

PR #1459, opened 2021-07-02
Include browser version in manifest

OskarDamkjaer:
Demo says it all @ http://update_manifest.surge.sh/manifest.json


jharris4 (approved):
Works perfectly, thanks for looking into this!

PR #1463, opened 2021-07-05
Prevent crash on database connection lost

OskarDamkjaer:
If the database was shut off, `bolt.hasMultiDbSupport` will throw which we don't handle properly. This PR catches whenever it could throw. 

Our the metadata check running every 20 seconds is _supposed_ to detect when a database connection is lost but it doesn't work properly (in part because of hasMultiDbSupport throwing). I spent a few hours trying to understand why/how it doesn't work, but didn't get anywhere and I think the whole file requires heavy refactoring. I ended up including a another call to `onConnectionLost` as part of the metadata check instead, so that we don't retry the websocket in eternity. It's somewhat of a scope creep and not a perfect solution, so I'm happy to drop it.

Preview @ http://crash_on_db_switch.surge.sh

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaDuck.ts
+++ b/src/shared/modules/dbMeta/dbMetaDuck.ts
@@ -376,7 +376,7 @@ const databaseList = (store: any) =>
           return resolve(null)
         }
 
-        const res = bolt.directTransaction(
+        const res = await bolt.directTransaction(
           'SHOW DATABASES',
           {},
           {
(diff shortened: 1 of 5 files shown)


src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -386,8 +387,10 @@ const databaseList = (store: any) =>
             useDb: SYSTEM_DB
           }
         )
-        .then(resolve)
```
jharris4, line 389: Looks like a missing catch on this promise...

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -259,54 +259,62 @@ const availableCommands = [
...
-          frames.add({
-            useDb: getUseDb(store.getState()),
-            ...action,
-            type: 'error',
-            error: UnsupportedError('No multi db support detected.')
-          })
-        )
-      }
+    exec(action: any, put: any, store: any) {
+      bolt
```
jharris4, line 263: just in case, we should return the promise, so that the code using these can await them if necessary...

jharris4 (approved):
LGTM

Issue #1461, opened 2021-07-05
Zoom IN way to limited!


RayLuxembourg:
**Description:**
Cannot zoom in to my liking. 

**Expected**
Zoom as much as I like.

OskarDamkjaer:
We'll be increasing the max zooming in/out limits in the next release, thanks for brining it to our attention.

PR #1464, opened 2021-07-06
Increase zoom limits and individual steps

OskarDamkjaer:
Prompted by #1461, I looked into increasing the max/min zoom. Was an easy fix, worked out zoom increments and max values with PM. 

preview @ http://more_zoom.surge.sh/


jharris4:
Changes look fine to me. My only concern is that changing the zoom bounds/steps like this to improve the experience for some users may come at the cost of worsening the experience for some other users.

Would be good to have @gregoryking weigh in to confirm that we're happy making this change.

OskarDamkjaer:
Got confirmation offline from Greg, merging now.

jharris4 (approved):
LGTM

PR #1466, opened 2021-07-06
Add index on productName for northwind guides

OskarDamkjaer:
@jexp requested the northwind guide be updated to have more indexes to increase lookup speed. This pr replaces #1434 as per offline discussion with Hunger. 

Preview @ http://guide_indexes.surge.sh


jharris4 (approved):
LGTM

PR #1468, opened 2021-07-07
Disable sentry and segment in CI

OskarDamkjaer:
window.Cypress is used to detect if we're running e2e tests as per the [docs](https://docs.cypress.io/faq/questions/using-cypress-faq#Is-there-any-way-to-detect-if-my-app-is-running-under-Cypress)

Preview @ http://disable_sentry_segment_in_ci.surge.sh

I think it'd make sense to gather the constants such as isMac or isInCI in some centralized place, but unsure where. Suggestions?

### DIFF
--- a/src/browser/AppInit.tsx
+++ b/src/browser/AppInit.tsx
@@ -166,10 +166,16 @@ export function setupSentry(): void {
           return 0.2
         }
       },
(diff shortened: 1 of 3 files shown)


jharris4 (approved):
LGTM

PR #1469, opened 2021-07-07
Persist UUID

OskarDamkjaer:
Currently UUID is lost when clearing favorites/localstorage through the browser sync tab, which is not the intention of the feature


jharris4 (approved):
Glad we got this nit fixed!

PR #1470, opened 2021-07-07
Update serializer to escape newlines

OskarDamkjaer:
PR fixes the following issue.

`return "a\nb" as a,"b\nc" as b`

creates this csv

```csv
a,b
a
b,b
c
```

while it should have been

```csv
a,b
"a
b","b
c"
```

Preview at http://csv_export_fix.surge.sh

### DIFF
--- a/src/shared/services/serializer.test.ts
+++ b/src/shared/services/serializer.test.ts
@@ -99,5 +99,14 @@ describe('resultTransform', () => {
       const s = CSVSerializer(cols)
       expect(s.output()).toEqual('"{""name"":""John""}"')
     })
+    test('should escape newlines in values properly ', () => {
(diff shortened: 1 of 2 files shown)


src/shared/services/serializer.test.ts
```diff
@@ -99,5 +99,14 @@ describe('resultTransform', () => {
       const s = CSVSerializer(cols)
       expect(s.output()).toEqual('"{""name"":""John""}"')
     })
+    test('should escape newlines in values properly ', () => {
```
jharris4, line 102: Let's add another test case here to make sure it handles the "\r\n" variation too.

jharris4 (approved):
LGTM, merge after tests pass

PR #1471, opened 2021-07-07
Show what command was unknown

OskarDamkjaer:
Solves error not being created properly. 

Before:
image

After: image

Preview at http://name_unknown_command.surge.sh


jharris4 (approved):
Nice simple clean fix, thanks!

PR #1472, opened 2021-07-12
Sysinfo cleanup

OskarDamkjaer:
To understand how the :sysinfo frame works, I've refactored and cleaned it up. This PR doesn't intend change any behaviour. 

Changes:
- Add types
- Renane SysInfoFrame from index.tsx to SysInfoFrame to be searchable and match other frames
- Remove unused state and props
- Extract SysinfoTable to it's own component
- Remove left over cluster handling that should have been removed in https://github.com/neo4j/neo4j-browser/pull/997


src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.test.tsx
```diff
@@ -34,27 +47,12 @@ jest.mock(
 )
 
 describe('sysinfo component', () => {
-  test('should render causal cluster table', () => {
```
jharris4, line 37: what's the reason for removing this test?

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
```diff
@@ -0,0 +1,205 @@
+/*
```
jharris4, line 1: Could we perhaps use git commands to rename this file? As it stands now, it looks like we'll lose the commit history on this file after it was renamed...

src/browser/modules/Stream/SysInfoFrame/legacyHelpers.tsx
```diff
@@ -236,41 +236,3 @@ export const responseHandler = (setState: any) =>
       success: true
     })
   }
-
-export const clusterResponseHandler = (setState: any) =>
```
jharris4, line 240: None of this cluster code was used? We should probably check that things are working properly with a server cluster... :-)

OskarDamkjaer: In the PR description I've linked the PR that motivated this change :)

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
OskarDamkjaer: I was thinking a single commit that only renamed would be enough, I'll try git mv!

OskarDamkjaer: Seems to me from this stack overflow thread it doesn't matter if I use git mv instead? https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history

jharris4 (approved):
LGTM

PR #1473, opened 2021-07-12
Fix sysinfo frame

OskarDamkjaer:
Depends on #1472 

Changelist:
- Fixes :sysinfo running before useDb has been set by rerunning the query when it's set.
- Adds docs on how the :sysinfo frame works
- Respects metrics.prefix
- Respect metrics.namespaces.enabled
- Adds error message on missing metrics to check the settings
- Changes default metrics to use metrics that aren't filtered out by default

This task quickly crept in scope and one of the issues I didn't have time to solve properly was collecting the settings through dbMetaDuck. Got a thumbs on from greg on the newly used metrics.

preview at http://blank_sysinfo_on_boot.surge.sh

### DIFF
--- a/src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
+++ b/src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
@@ -28,7 +28,6 @@ import {
   getUseDb
 } from 'shared/modules/connections/connectionsDuck'
 import FrameTemplate from 'browser/modules/Frame/FrameTemplate'
-import FrameError from 'browser/modules/Frame/FrameError'
 import {
   StyledStatusBar,
   AutoRefreshToggle,
@@ -60,6 +59,8 @@ export type SysInfoFrameState = {
   success: boolean
   autoRefresh: boolean
   autoRefreshInterval: number
+  namespacesEnabled: boolean
+  userConfiguredPrefix: string
 }
 
 type SysInfoFrameProps = {
@@ -87,13 +88,9 @@ export class SysInfoFrame extends Component<
     results: false,
     success: false,
     autoRefresh: false,
-    autoRefreshInterval: 20 // seconds
-  }
-  metricsSettings?: {
-    namespacesEnabled: boolean
-    userConfiguredPrefix: string
-    jmxDisabled: boolean
-    metricsDisabled: boolean
+    autoRefreshInterval: 20, // seconds
+    namespacesEnabled: false,
+    userConfiguredPrefix: 'neo4j'
   }
 
   componentDidMount(): void {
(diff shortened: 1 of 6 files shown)


src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
```diff
@@ -45,6 +44,8 @@ import { SysInfoTable } from './SysInfoTable'
 import { Bus } from 'suber'
 import { GlobalState } from 'shared/globalState'
 import { Frame } from 'shared/modules/stream/streamDuck'
+import { ExclamationTriangleIcon } from '../../../components/icons/Icons'
+import styled from 'styled-components'
```
jharris4, line 48: should this be in a separate styled.tsx file like we do elsewhere?

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
```diff
@@ -107,26 +153,33 @@ export class SysInfoFrame extends Component<
         this.timer && clearInterval(this.timer)
       }
     }
+
+    if (prevProps.useDb !== this.props.useDb) {
+      this.getSysInfo()
+    }
   }
 
   getSysInfo(): void {
```
jharris4, line 162: let's make this an arrow function / class property to get rid of the need to bind it...

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
```diff
@@ -81,20 +83,64 @@ export class SysInfoFrame extends Component<
...
-    autoRefreshInterval: 20 // seconds
+    autoRefreshInterval: 20, // seconds
+    namespacesEnabled: false,
+    userConfiguredPrefix: 'neo4j'
   }
-  helpers = this.props.hasMultiDbSupport ? helpers : legacyHelpers
 
   componentDidMount(): void {
-    this.getSysInfo()
+    this.getSettings()
```
jharris4, line 95: It could cause problems that we only fetch the settings once on mount, but should work fine for the most common cases

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
```diff
@@ -81,20 +83,64 @@ export class SysInfoFrame extends Component<
...
+
+                  return newState
+                },
+                {}
+              )
+
+              this.setState(newState)
+              resolve()
+            } else {
+              reject()
```
jharris4, line 133: maybe return the "failed to run listConfig" message here, and a different error when not connected?

src/browser/modules/Stream/SysInfoFrame/helpers.tsx
```diff
@@ -21,71 +21,142 @@
...
+
+function flatten<T>(acc: T[], curr: T[]): T[] {
+  return acc.concat(curr)
+}
+
+export const responseHandler = (setState: (newState: any) => void) =>
   function(res: any): void {
     if (!res || !res.result || !res.result.records) {
-      setState({ success: false })
       return
```
jharris4, line 157: no error message here?

src/browser/modules/Stream/SysInfoFrame/helpers.tsx
```diff
@@ -159,22 +230,32 @@ export const responseHandler = (
...
+      { label: 'Committed Read', value: tx.committed_read },
+      { label: 'Committed Write', value: tx.committed_write }
     ]
 
+    const valuesMissing = [
+      storeSizes,
+      pageCache,
+      idAllocation,
+      transactions
+    ].some(type => type.some((item: any) => !item.value))
```
jharris4, line 250: Should this maybe be an explicit check for undefined?

jharris4 (approved):
Definitely an improvement on how it was before, nice work!

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
OskarDamkjaer: Opted to to do this to spam the server less, detecting server changes is an area of weakness for sure.

OskarDamkjaer:
Did some more manual testing of this branch (esp compared to current productino) to make sure the setting are handled properly. It worked well, so merging now

PR #1480, opened 2021-07-23
[WIP] Use read transactions when able

OskarDamkjaer:
Currently we use write transactions on all our metadata queries, even if they could be reads, this causes issues in a cluster when there's no writer available (since the driver will attempt to route writeTransactions to "write" server).

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
+++ b/src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
@@ -28,7 +28,7 @@ import { withBus } from 'react-suber'
 import { ExplorerComponent } from '../../D3Visualization/components/Explorer'
 import { StyledVisContainer } from './VisualizationView.styled'
 
-import { CYPHER_REQUEST } from 'shared/modules/cypher/cypherDuck'
+import { ROUTED_CYPHER_READ_REQUEST } from 'shared/modules/cypher/cypherDuck'
 import { NEO4J_BROWSER_USER_ACTION_QUERY } from 'services/bolt/txMetadata'
 import { getMaxFieldItems } from 'shared/modules/settings/settingsDuck'
 import { resultHasTruncatedFields } from 'browser/modules/Stream/CypherFrame/helpers'
@@ -119,11 +119,10 @@ export class Visualization extends Component<any, VisualizationState> {
     return new Promise((resolve, reject) => {
       this.props.bus &&
         this.props.bus.self(
-          CYPHER_REQUEST,
+          ROUTED_CYPHER_READ_REQUEST,
           {
             query: query,
-            queryType: NEO4J_BROWSER_USER_ACTION_QUERY,
-            useDirectReadTransaction: true
+            queryType: NEO4J_BROWSER_USER_ACTION_QUERY
           },
           (response: any) => {
             if (!response.success) {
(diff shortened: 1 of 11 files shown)


src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
```diff
@@ -119,8 +119,11 @@ export class Visualization extends Component<any, VisualizationState> {
     return new Promise((resolve, reject) => {
       this.props.bus &&
         this.props.bus.self(
-          CYPHER_REQUEST,
-          { query: query, queryType: NEO4J_BROWSER_USER_ACTION_QUERY },
+          ROUTED_CYPHER_READ_REQUEST,
```
OskarDamkjaer, line 122: Risk on consistency, as the read might go to a core that's not got the write update from the main query yet

src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
```diff
@@ -154,7 +157,7 @@ export class Visualization extends Component<any, VisualizationState> {
     return new Promise<any>((resolve, reject) => {
       this.props.bus &&
         this.props.bus.self(
-          CYPHER_REQUEST,
+          ROUTED_CYPHER_READ_REQUEST,
```
OskarDamkjaer, line 160: Risk for consistency issues, as the write query to would be on a different core from the read query and it might not have gotten the update yet. Should be handled by using the same session through something called "bookmarks"

src/browser/modules/Stream/SysInfoFrame/SysInfoFrame.tsx
```diff
@@ -169,7 +169,7 @@ export class SysInfoFrame extends Component<
     if (bus && isConnected && useDb) {
       this.setState({ lastFetch: Date.now() })
       bus.self(
-        CYPHER_REQUEST,
+        ROUTED_CYPHER_READ_REQUEST,
```
OskarDamkjaer, line 172: It might make sense to use write here to make sure it goes to the leader, which usually has the stats we care about.

OskarDamkjaer: We do a routed -> If we're on a cluster we say stats are from the LEADER server. We could say browser only supports listing sysinfo from leader, other wise use opts mgmr

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -468,7 +468,7 @@ const clusterRole = (store: any) =>
         return resolve(null)
       }
       bolt
-        .directTransaction(
+        .routedReadTransaction(
```
OskarDamkjaer, line 471: this one we'll need to do as a direct query. otherwise the role will be scrablmed

OskarDamkjaer: Anyway when running with Neo4j protocol the current role loses it's meaning, as it could have been re-routed on the server side (or the client side)

OskarDamkjaer:
it'd make sense to change the main CYPHER_REQUEST to use routed reads if connected via neo4j rather than bolt

src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
OskarDamkjaer: Leaving this as is for that reason

OskarDamkjaer:
While these changes probably would be useful, it's hard to gauge what unintended side effects they might have. Closing this for now and will re-open if we prioritize this sort of work in the future

PR #1481, opened 2021-07-23
Sysinfo 3.x Fixes

OskarDamkjaer:
Changes:
- Fixes bug introduced in the :sysinfo frame refresh were we didn't fetch any metrics for 3.x databases
- No longer shows database table if server lacks multidb support


jharris4 (approved):
Nice, glad we've got this fixed!

PR #1483, opened 2021-07-26
Add exports to individual folder

OskarDamkjaer:
So that users don't need to select all scripts in a folder manually. Also allows for exporting them as a single cypher file, which has been requested internally.

### DIFF
--- a/src/browser/components/SavedScripts/SavedScriptsFolder.tsx
+++ b/src/browser/components/SavedScripts/SavedScriptsFolder.tsx
@@ -42,11 +42,13 @@ import {
   ContextMenuItem
 } from './styled'
 import { Folder } from 'shared/modules/favorites/foldersDuck'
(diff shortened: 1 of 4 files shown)


src/browser/components/SavedScripts/SavedScriptsFolder.tsx
```diff
@@ -116,6 +119,24 @@ function SavedScriptsFolder({
...
+        data-testid="contextMenuExport"
+        onClick={() => exportScripts('ZIPFILE')}
+        key="exportZip"
+      >
+        Export scripts as .zip file
+      </ContextMenuItem>
+    ),
+    exportScripts && (
+      <ContextMenuItem
+        data-testid="contextMenuRemove"
```
jharris4, line 134: looks like a wrong name here? `Remove` should be `Export` or some variant of that?

jharris4 (approved):
Neat extra little feature

PR #1484, opened 2021-07-28
Emphasized the log in / logout experience

renetapopova:
Trello card: [add browser doc on the log in / logout experience](https://trello.com/c/bDKwUq7d/3400-add-browser-doc-on-the-log-in-logout-experience)


martin-neotech (approved):
Thanks.

PR #1487, opened 2021-08-04
do not allow sysinfo command when using system db

jharris4:
Screen Shot 2021-08-04 at 9 36 31 AM

### DIFF
--- a/src/shared/services/commandInterpreterHelper.ts
+++ b/src/shared/services/commandInterpreterHelper.ts
@@ -48,7 +48,8 @@ import {
   getRemoteContentHostnameAllowlist,
   getDatabases,
   fetchMetaData,
-  getAvailableSettings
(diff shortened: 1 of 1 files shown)


src/shared/services/commandInterpreterHelper.ts
```diff
@@ -354,13 +354,27 @@ const availableCommands = [
...
     exec(action: any, put: any, store: any) {
-      put(
-        frames.add({
-          useDb: getUseDb(store.getState()),
-          ...action,
-          type: 'sysinfo'
-        })
-      )
+      const useDb = getUseDb(store.getState())
+      if (useDb === 'system') {
```
oskarhane, line 358: Use existing constant for this please

oskarhane (approved):
🌤️ LGTM

PR #1492, opened 2021-08-12
Fix redirect bug

OskarDamkjaer:
Fix typos and a bug that caused sso_redirect param not to work

### DIFF
--- a/src/shared/modules/auth/index.js
+++ b/src/shared/modules/auth/index.js
@@ -1,6 +1,6 @@
 import {
   getCredentialsFromAuthResult,
-  getInitializationParameters,
+  getInitialisationParameters,
   temporarilyStoreUrlSearchParams
 } from './common'
 import {
@@ -140,7 +140,7 @@ export const handleAuthFromRedirect = SSOProviders =>
       code,
       state,
       error
-    } = getInitializationParameters()
+    } = getInitialisationParameters()
 
     authLog('Handling auth redirect from SSO server')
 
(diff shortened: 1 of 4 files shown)


src/shared/modules/auth/index.js
```diff
@@ -26,7 +26,7 @@ import {
 } from './settings'
 
 export const authRequestForSSO = async selectedSSOProvider => {
-  authLog('Initializing auth redirect request for SSO')
+  authLog('Initialising auth redirect request for SSO')
```
jharris4, line 29: heh, the original spelling was correct since we mostly use American spelling here! :-)

src/shared/modules/discovery/discoveryDuck.ts
```diff
@@ -274,7 +274,7 @@ export const discoveryOnStartupEpic = (some$: any, store: any) => {
           authLog(err.message)
         }
       } else if (wasRedirectedBackFromSSOServer()) {
-        authLog('Handling auth_flow_step redirect')
+        authLog('Initialising auth_flow_step redirect')
```
jharris4, line 277: Use a z here too for yankee spelling...

src/shared/modules/auth/settings.js
```diff
@@ -13,7 +13,7 @@ export const mandatoryKeysForSSOProviderParams = [
   'scope'
 ]
 
-export const searchParamsToRemoveAfterAutoRedirect = ['cmd', 'arg']
+export const searchParamsToRemoveAfterAutoRedirect = ['sso_redirect']
```
jharris4, line 16: woops, good catch!

PR #1495, opened 2021-08-13
Fix export button for history and add tests

OskarDamkjaer:
Frame refactor in last release seemed to have broken exporting history. This PR fixes it, cleans up the file and adds tests to make sure it keeps working.

Preview @ http://export_history.surge.sh

### DIFF
--- a/e2e_tests/integration/data-export.spec.ts
+++ b/e2e_tests/integration/data-export.spec.ts
@@ -80,7 +80,14 @@ describe('Data export', () => {
   it('can export history', () => {
     cy.executeCommand(':clear')
     cy.executeCommand(':history')
-    cy.get('[data-testid="exportGrassButton"]').should('not.exist')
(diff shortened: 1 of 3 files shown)


e2e_tests/integration/data-export.spec.ts
```diff
@@ -76,4 +76,11 @@ describe('Data export', () => {
       })
     })
   })
+
+  it('can export history', () => {
+    cy.executeCommand(':clear')
+    cy.executeCommand(':history')
+    cy.get('[data-testid="exportGrassButton"]').should('not.exist')
+    cy.get('[data-testid="exportHistoryButton"]').should('not.exist')
```
jharris4, line 84: hmm, this tests looks like it needs fixing?

OskarDamkjaer: yes good catch

jharris4 (approved):
LGTM

PR #1496, opened 2021-08-13
Add postinstall hook to run update licenses

OskarDamkjaer:


OskarDamkjaer:
Hook won't work on windows, let's try another approach

PR #1499, opened 2021-08-16
Unbreak export of favorites

OskarDamkjaer:
Preview @ https://unbreak_export.surge.sh


jharris4 (approved):
Nice fix. Somewhat unrelated nit: it'd be nice to have `cursor: pointer;` on the menu button next to each favorite

PR #1503, opened 2021-08-18
Fix issues with trailing spaces in commands

OskarDamkjaer:
Preview @ http://trailing_space.surge.sh

Previously ":history " and ":queries " would result in an error


jharris4 (approved):
LGTM

PR #1504, opened 2021-08-18
Add params to background explain query

OskarDamkjaer:
Our background explain query we do to get warnings, was not using the client side params we have. So it'll give warnings even as queries are correct. To reproduce 

1. Run `:param p => 1`
2. Type `RETURN $p`
3. Get warning on $p not being defined, although the query runs fine

I think there's a better way to solve this than passing the params down to the monaco component, but that's a change for when we get rid of the Bus


jharris4 (approved):
LGTM

PR #1505, opened 2021-08-18
List external guides

OskarDamkjaer:
Once you've clicked the back arrow on an external guide, it wasn't possible to find your way back. This PR adds a list of remote guides to the indexpage of the sidebar guides. Previously the index page was itself a guide, but I've come to realize that was a bit to cute. I've refactored indexpage to use a "GuidePicker" component instead.

Preview @https://list_external_guides.surge.sh

Easiest way to test this: `:guide graph-data-science` and it'll show up in the sidebar


src/browser/modules/Sidebar/GuidesDrawer.tsx
```diff
@@ -59,33 +69,49 @@ function GuidesDrawer({
...
-  backToAllGuides: () => dispatch(startGuide()),
-  gotoSlide: (slideIndex: number) => dispatch(gotoSlide(slideIndex))
+const mapStateToProps = (state: GlobalState) => ({
+  currentGuide: getCurrentGuide(state),
+  remoteGuides: getRemoteGuides(state)
+})
+const mapDispatchToProps = (dispatch: Dispatch<Action>) => ({
+  backToAllGuides: () => dispatch(resetGuide()),
+  gotoSlide: (slideIndex: number) => dispatch(gotoSlide(slideIndex)),
+  setGuide: (guide: Guide) => dispatch(setGuide(guide)),
```
jharris4, line 112: would be nice if we were consistent about naming of `setGuide` -> `setCurrentGuide` ...

src/shared/modules/guides/guidesDuck.ts
```diff
@@ -22,61 +22,120 @@ import docs from 'browser/documentation'
...
 }
 
-const defaultGuide: Guide = {
-  ...docs.guide.chapters.index,
-  currentSlide: 0
-}
 export interface GuideState {
-  guide: Guide
+  currentGuide: Guide | null
+  remoteGuides: Guide[]
```
jharris4, line 43: We'll need to make sure this is saved/loaded to/from localStorage, and that it doesn't break when it is missing in localstorage (it seems guides were not persisted in previous browser versions)

src/browser/AppInit.tsx
```diff
@@ -66,6 +66,7 @@ applyKeys(
   'documents',
   'folders',
   'grass',
+  'guides',
```
jharris4: Unfortunately we can't serialize the guides like this, as they contain Jsx elements which don't serialize properly and cause a crash when you try to view guides after reloading...

jharris4 (approved):
I'm approving the changes for listing the remote guides, even though the persistence to local storage isn't done yet.

PR #1507, opened 2021-08-18
Remove <Render/> component

OskarDamkjaer:
This PR removes the <Render/> component helper in favour of more classical react patterns to make the code base easier to understand for new contributors

Preview @ https://remove_render_comp.surge.sh


src/browser/modules/Stream/ParamsFrame.tsx
```diff
@@ -47,14 +47,14 @@ const ParamsFrame = ({ frame }: any) => {
     typeof frame.success === 'undefined' ? null : (
       <StyledStatsBar>
         <Ellipsis>
-          <Render if={frame.success === true}>
```
jharris4, line 50: just a nit, but this could be combined into ternary, no?

jharris4:
Looking good aside from a small nit. I don't see the commit to remove the Render component now that it's unused though?

jharris4 (approved):
LGTM

src/browser/modules/Stream/ParamsFrame.tsx
OskarDamkjaer: I wasn't confident with this, with the weird "undefined" string check, so leaving it

PR #1508, opened 2021-08-19
added sso content and updated content-nav

AlexicaWright:

### DIFF
--- a/docs/modules/ROOT/pages/browser-sso.adoc
+++ b/docs/modules/ROOT/pages/browser-sso.adoc
@@ -1,0 +1,112 @@
+[[browser-sso]]
+= Neo4j Browser Single Sign-On (SSO)
+:description: This section describes SSO for Browser.
+
+Browser provides basic support for some Single Sign On (SSO) providers under specific Enterprise environment configurations.
+Contact your Neo4j representative for additional assistance.
+Your organization needs an existing compatible SSO provider (local or external) already configured.
+
+Supported providers at this time include OpenID Connect (OIDC) OAuth 2.0 providers Okta, KeyCloak, and Microsoft Azure AD.
+SSO has been tested on Neo4j 4.2.x and 4.3.x deployments.
+Browser supports two authorization flows:
+* Authorization Code flow with PKCE
+* Implicit flow
+
+[IMPORTANT]
+====
+It is strongly advised to use PKCE to ensure security.
+Further information about OpenID Connect and OAuth can be found at https://openid.net/connect/.
+====
+
+In addition to having an SSO provider already configured at your organization, configuring SSO for Browser and Neo4j requires a separate OAuth plugin, available from your Neo4j representative.
+With these prerequisites in place, two additional steps are required:
+
+. The https://neo4j.com/docs/operations-manual/current/configuration/neo4j-conf/#neo4j-conf[`neo4j.conf`] file should be updated with the following information (be sure to avoid duplicate entries in the `neo4j.conf` file):
++
+[source, properties]
+----
+dbms.security.auth_enabled=true
+dbms.security.authentication_providers=native,plugin-org.neo4j.auth.openid.OpenIdPlugin
+dbms.security.authorization_providers=native,plugin-org.neo4j.auth.openid.OpenIdPlugin
+plugins.auth.openid.configuration=https://login.org.com/.well-known/openid-configuration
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/browser-sso.adoc
```diff
@@ -0,0 +1,112 @@
...
+[source, properties]
+----
+dbms.jvm.additional=-Dorg.apache.commons.logging.Log=org.apache.commons.logging.impl.SimpleLog
+dbms.jvm.additional=-Dorg.apache.commons.logging.simplelog.showdatetime=true
+dbms.jvm.additional=-Dorg.apache.commons.logging.simplelog.log.org.apache.http=DEBUG
+----
+
+Browser needs to be aware of the identity providers available for use.
+This is done by specifying a URL parameter discoveryURL that specifies a URL to a json file containing the SSO providers.
+Example for Browser: `http://<browser-server-host>:<http-port>?discoveryURL=http://webhost.com/public/discovery.json`.
```
jharris4, line 49: @OskarDamkjaer can you please confirm this example url is correct?

OskarDamkjaer:
I think it would make sense to mention that the `:debug` command also contains logging from our SSO implementation and is helpful for debugging

docs/modules/ROOT/pages/browser-sso.adoc
OskarDamkjaer: Looks like correct example to me!

OskarDamkjaer: We could mention that it's planned to be picked up from the server in the future

docs/modules/ROOT/pages/browser-sso.adoc
```diff
@@ -0,0 +1,112 @@
...
+// It contains all the possible parameters you can provide.
+// You most likely do not need all the parameters.
+// If you are unsure, please consult your Neo4j representative to avoid misconfiguration.
+//
+// [source, parameters]
+// ----
+//
+// {
+// 	// other discovery entries
+// 	// e.g. "bolt": "bolt://localhost:7687"
```
OskarDamkjaer, line 62: While not strictly required, I think it's a good idea to have the "bolt" field included in your discovery data. If it's not present here, in the built-in discovery or in the connectURL param we don't know where to login to when we come back from the SSO server

AlexicaWright: This section is about the `discovery.json` file and that doesn't apply to Browser, if I'm not mistaken? Where should I include it?

docs/modules/ROOT/pages/browser-sso.adoc
AlexicaWright: What is planned to be picked up from the server?

OskarDamkjaer: There's a built in discovery.json as well when the browser is hosted by neo4j, it currently only has `bolt` and some other fields but in the future it'll be configurable to have SSOProviders as well. But we can wait with documenting that until it's available in 4.4

docs/modules/ROOT/pages/browser-sso.adoc
OskarDamkjaer: Ah sorry, yes these do apply to us as well!

docs/modules/ROOT/pages/browser-sso.adoc
```diff
@@ -0,0 +1,112 @@
...
+dbms.jvm.additional=-Dorg.apache.commons.logging.simplelog.showdatetime=true
+dbms.jvm.additional=-Dorg.apache.commons.logging.simplelog.log.org.apache.http=DEBUG
+----
+
+Browser needs to be aware of the identity providers available for use.
+This is done by specifying a URL parameter discoveryURL that specifies a URL to a json file containing the SSO providers.
+Example for Browser: `http://<browser-server-host>:<http-port>?discoveryURL=http://webhost.com/public/discovery.json`.
+
+// The `discovery.json` file must contain entries tailored to your organization’s specific SSO solution.
+// Below is a reference discovery file for the ID provider (IDP) _Keycloak_ containing one SSO provider.
```
OskarDamkjaer, line 52: Set up to be running on port "18080", it'd be worth to add here I think

OskarDamkjaer (approved):
Looks good, thanks!

PR #1512, opened 2021-08-21
fix monaco fullscreen resize bug

jharris4:
http://fix-monaco-fullscreen-bug.surge.sh/

Fixes the issue where multi-line main editor in fullscreen didn't take up the full height, as shown in this screenshot:
Screen Shot 2021-08-23 at 9 30 36 AM


oskarhane (approved):
Looks good!

Is there a way to test this? Like go into fullscreen and check the height or something?
Something to think about going forward maybe, but not a blocker for this fix.

jharris4:
@oskarhane Agreed that it would be great to have an e2e test to catch any regressions on this in the future, we'll look to get that added soon!

PR #1518, opened 2021-09-06
Better coverage for fullscreen

OskarDamkjaer:
Improves our cypress coverage of fullscreen for both the main editor and re-usable frame which are prone to bugs.


package.json
```diff
@@ -104,7 +104,7 @@
     "copy-webpack-plugin": "^6.0.3",
     "cross-env": "^6.0.3",
     "css-loader": "^1.0.0",
-    "cypress": "^7.0.1",
+    "cypress": "8.3.1",
```
jharris4, line 107: why the strict version here, and not `^` ?

e2e_tests/integration/fullscreen.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    cy.get(EditorArea)
+      .invoke('height')
+      .should('equal', 286)
+
+    cy.get(EditorArea)
+      .invoke('width')
+      .should('equal', viewportWidth - 80) // normal editor width
+  })
+
+  it('re-usable fram can also use fullscreen', () => {
```
jharris4, line 99: typo

e2e_tests/integration/fullscreen.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    cy.get('[id^=monaco-]')
+      .eq(1)
+      .invoke('height')
+      .should('equal', 276)
+
+    cy.get('[id^=monaco-]')
+      .eq(1)
+      .invoke('width')
+      .should('equal', viewportWidth - 184) // normal editor width
```
jharris4, line 161: Can we at least move these numbers to constants with somewhat descriptive names? This would both make it easier to understand, and to change in the future.

package.json
OskarDamkjaer: No reason, was just copying the snippet cypress gave me

e2e_tests/integration/fullscreen.spec.ts
OskarDamkjaer: WIll do!

e2e_tests/integration/fullscreen.spec.ts
jharris4: so that's a thumbs up to leaving the typo in there? lol

jharris4 (approved):
The typo in the test name is not a blocker, so I'm approving this :-)

PR #1522, opened 2021-09-13
fix: Remove routing context when converts from neo4j to bolt protocol

bigmontz:
Browser during the loggin is creating direct connection to the informed `connect url`, but it's not removing the query string from the url. Because of this, browser is getting a `Parameters are not supported with none route scheme..` error. The routing context params are not support in direct connection by design.

The solution for this issue is removing the routing context (query string) from the url when transalate from `neo4j` to `bolt`.


jharris4:
I deployed this to surge so we can get some testing done to see if it fixes the issue without breaking anything like `:sysinfo` command etc: http://policy-query-param-fix.surge.sh/

oskarhane (approved):
LGTM

PR #1523, opened 2021-09-13
Add luxon to fix duration formatting issue

OskarDamkjaer:
Changes:
- Adds luxon as a date/duration management lib
- Formats durations better for both exports and cyper result frame

Before:
CleanShot 2021-09-13 at 17 19 28

After:
CleanShot 2021-09-13 at 17 19 46

Preview @ https://format_duration.surge.sh


package.json
```diff
@@ -30,7 +30,8 @@
     "e2e-local": "CYPRESS_E2E_TEST_ENV=\"local\" cypress run",
     "e2e-local-open": "CYPRESS_E2E_TEST_ENV=\"local\" cypress open",
     "format": "prettier-eslint --write 'src/**/!(*.min).{js,jsx,ts,tsx,css,json}' 'e2e_tests/**/*.{js,jsx,ts,tsx,css,json}' 'scripts/**/*.{js,jsx,ts,tsx,css,json}' 'build_scripts/**/*.{js,jsx,ts,tsx,css,json}'",
-    "jest": "jest --coverage",
```
jharris4, line 33: This will affect CI, so we should either just add a new `jest-no-cov` script and leave the existing `jest` as it was, or else we should check what the change will mean/do on CI.

src/browser/modules/Stream/CypherFrame/helpers.ts
```diff
@@ -443,9 +444,13 @@ export function mapNeo4jValuesToPlainValues(values: any): any {
  */
 function neo4jValueToPlainValue(value: any) {
   switch (get(value, 'constructor')) {
+    case neo4j.types.Duration:
```
jharris4, line 447: Can we move this code that converts the value to a duration string to a shared util function or something?

jharris4 (approved):
LGTM

src/browser/modules/Stream/CypherFrame/helpers.ts
OskarDamkjaer: Done!

PR #1524, opened 2021-09-13
Redirect play if able

OskarDamkjaer:
The next step in moving off playframes is to automatically redirect built-in guides to the sidebar guides

preview @ http://format_duration.surge.sh


src/shared/services/commandInterpreterHelper.ts
```diff
@@ -590,6 +591,28 @@ const availableCommands = [
     name: 'play',
     match: (cmd: any) => /^play(\s|$)/.test(cmd),
     exec(action: any, put: any, store: any) {
+      // Built in play guides where migrated to
+      // use the guide command instead
+      const legacyBuiltInGuides = [
+        'concepts',
+        'cypher',
+        'index',
```
jharris4, line 599: should this be here?

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -590,6 +591,28 @@ const availableCommands = [
     name: 'play',
     match: (cmd: any) => /^play(\s|$)/.test(cmd),
     exec(action: any, put: any, store: any) {
+      // Built in play guides where migrated to
+      // use the guide command instead
+      const legacyBuiltInGuides = [
```
jharris4, line 596: can we remove the play command source code now if they're never used anymore?

src/shared/services/commandInterpreterHelper.ts
OskarDamkjaer: Removing

jharris4 (approved):
LGTM

PR #1528, opened 2021-09-15
Include pathname in redirect

OskarDamkjaer:
The SSO-redirect accidentally removes the pathname from the URL as well, breaking SSO on aura. This fixes it, tested locally with browser hosted on /browser


oskarhane (approved):
Looks good to me - though we should think about adding tests in the shared lib later. 🧔

PR #1532, opened 2021-09-20
Remove browsersync banner

OskarDamkjaer:
To make room for the new analytics consent banner we're removing the browser sync banners. Preview @ https://remove_browser_sync_banner.surge.sh


src/browser/modules/Main/Main.tsx
```diff
@@ -83,8 +81,6 @@ const Main = React.memo(function Main(props: any) {
           Server is taking a long time to respond...
         </WarningBanner>
       )}
-      {props.utilizeBrowserSync && <SyncReminderBanner />}
```
jharris4, line 86: Seems like this prop is still being passed to the component even though not used anymore?

jharris4 (approved):
LGTM

PR #1535, opened 2021-09-23
Get desktopAllowTracking and desktopTrackingId from DesktopApi

jharris4:


OskarDamkjaer (approved):
Nice, thanks for getting this part out of the way

PR #1538, opened 2021-09-27
SSO lib integration

OskarDamkjaer:
Integration of the new SSO library built by @tbwiss (thanks) rebased against current master with some small typescript updates. Tested against local keycloak installation.


oskarhane (approved):
LGTM - Could you test against an external ipd before merging?

OskarDamkjaer:
Tested again w/ azure. Merging

PR #1542, opened 2021-10-01
Rework product analytics consent

OskarDamkjaer:
This PR reworks our product analytics consent settings, we now coalesce settings from Aura, Desktop, a new setting in neo4j.conf and a new setting in browser to determine if we can send user data and crash reports. To make it clear what setting takes prevalence and be confident that they are followed, I've created the util function `usedTelemetrySettingSource`. 

The browser setting also shows where the to go to change the current settings (if not in browser) and shows a banner on the first five mounts, informing users of the new setting. Have also reached out to PM to get a review to test from the statistics end.

### DIFF
--- a/src/browser/modules/App/App.tsx
+++ b/src/browser/modules/App/App.tsx
@@ -129,7 +129,6 @@ export function App(props: any) {
               browserVersion: version,
               ...originalData
             }
-            console.log(category, label, data, props.telemetrySettings)
             eventMetricsCallback &&
               eventMetricsCallback.current &&
               eventMetricsCallback.current({ category, label, data })
@@ -186,7 +185,6 @@ export function App(props: any) {
             allowSendReports: false,
             allowSendStats: false
           }
-          console.log(args[1])
           updateDesktopUDCSettings({
             allowCrashReportsInDesktop: allowSendReports,
             allowUserStatsInDesktop: allowSendStats,
--- a/src/browser/modules/App/App.tsx
+++ b/src/browser/modules/App/App.tsx
@@ -81,11 +81,7 @@ import { getMetadata, getUserAuthStatus } from 'shared/modules/sync/syncDuck'
 import ErrorBoundary from 'browser-components/ErrorBoundary'
 import { getExperimentalFeatures } from 'shared/modules/experimentalFeatures/experimentalFeaturesDuck'
 import FeatureToggleProvider from '../FeatureToggle/FeatureToggleProvider'
-import {
-  inDesktop,
-  inWebEnv,
-  URL_ARGUMENTS_CHANGE
-} from 'shared/modules/app/appDuck'
+import { inWebEnv, URL_ARGUMENTS_CHANGE } from 'shared/modules/app/appDuck'
 import useDerivedTheme from 'browser-hooks/useDerivedTheme'
 import FileDrop from 'browser-components/FileDrop/FileDrop'
 import DesktopApi from 'browser-components/desktop-api/desktop-api'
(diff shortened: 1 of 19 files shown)


src/browser/modules/App/App.tsx
```diff
@@ -83,180 +87,280 @@ import {
...
+        allowUserStats: this.props.allowUdcInAura
+      },
+      BROWSER_SETTING: {
+        allowCrashReporting: this.props.browserAllowCrashReports,
+        allowUserStats: this.props.browserAllowUserStats
+      },
+      NEO4J_CONF: {
+        allowCrashReporting: this.props.neo4jConfAllowsUdc,
+        allowUserStats: this.props.neo4jConfAllowsUdc
+      }
```
oskarhane, line 151: I'm a big fan of configuration over conditionals, so nice one here!

src/browser/modules/App/App.tsx
```diff
@@ -83,180 +87,280 @@ import {
...
+}
 
-export function App(props: any) {
-  const [derivedTheme, setEnvironmentTheme] = useDerivedTheme(
-    props.theme,
-    LIGHT_THEME
-  )
-  // @ts-expect-error ts-migrate(7053) FIXME: No index signature with a parameter of type 'strin... Remove this comment to see the full error message
-  const themeData = themes[derivedTheme] || themes[LIGHT_THEME]
+export class App extends React.Component<any, AppState> {
```
oskarhane, line 117: What's the reason for moving to a class component here?

src/browser/modules/App/App.tsx
```diff
@@ -83,180 +87,280 @@ import {
...
+      openSettingsDrawer,
+      setConsentBannerShownCount,
+      setEnvironmentTheme,
+      store,
+      syncConsent,
+      telemetrySettingSource,
+      themeData,
+      titleString,
+      useDb
+    } = this.props
```
oskarhane, line 212: Out of scope for this PR, but this number of props here hints that this god-component might be subject for being split into multiple more narrow scoped components.

oskarhane (approved):
Great!
Is there a way to write more tests on this to make sure it's not broken in the future?

Either E2E tests (should be doable to fake an Aura env?), or maybe component tests on the settings sidebar with a fake store.
The `:debug` output might be useful in the E2E tests.

src/browser/modules/App/App.tsx
OskarDamkjaer: I was thinking the same and left a mental note (it's also missing prop types)

src/browser/modules/App/App.tsx
OskarDamkjaer: The metrics data event handler is now dependent on the redux state to know if we can send event or not, since it was registered in a useEffect, the handler would need to be unregistered and reregistered on every mount state had changed. At first I wasn't confident that would work well with segment and moved to a class component where the callback reference could be reused, but looking at the final implementation we could have stayed as a function component.

src/browser/modules/App/App.tsx
```diff
@@ -256,7 +288,7 @@ export function App(props: any) {
   )
 }
 
-const mapStateToProps = (state: any) => {
+const mapStateToProps = (state: GlobalState) => {
```
oskarhane, line 291: 👍

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -697,6 +707,7 @@ export const serverConfigEpic = (some$: any, store: any) =>
           return Rx.Observable.of(null)
         })
     })
+    .do(() => store.dispatch(update({ serverConfigDone: true })))
```
oskarhane, line 710: Not sure I'm a big fan of using this as a one-time thing - because this epic runs at an interval and the value can change between runs. 
Using it like this kind of implies that it only runs once.
A name change might be good enough. `serverConfigReads: Int` maybe to really emphasize that it's a recurring thing?

I'm going to approve this anyway now, but give it a think and change if you want.

oskarhane: This new state opens up for browser to be in a pending state for other areas as well (all other configs just above this) rather than having a local default to fallback on.

src/browser/modules/Main/styled.tsx
```diff
@@ -66,6 +65,26 @@ export const NotAuthedBanner = styled(Banner)`
...
+  background-color: ${props => props.theme.auth};
+  display: flex;
+  justify-content: space-between;
+`
+export const DismissConsentBanner = styled.span`
+  &:hover {
+    cursor: pointer;
+  }
+  &:after {
+    content: 'X';
```
oskarhane, line 78:
```suggestion
    content: '\00D7';
    font-size: 1.4rem;
```

oskarhane (commented):
Looks good. I think you should make the CSS change I suggest below, but other than that 👍

oskarhane (approved):
🐝

src/shared/modules/dbMeta/dbMetaDuck.ts
OskarDamkjaer: I think serverConfigReads is a better implementation, good idea!

PR #1559, opened 2021-10-13
Fixed navigation

martin-neotech:
Fixed navigation and links.


AlexicaWright:
👍   I very much like this! It's perfectly clear and I will implement this on the Bloom and Desktop docs also. Thank you @martin-neotech !

PR #1561, opened 2021-10-13
Fix :auto for multistatement queries

OskarDamkjaer:
Multistatements with :auto would get ignored since they were not in the list of allowed multistatement commands. Preview @ http://fix_auto_multlistatement.surge.sh 

This query can be used to verify that the fix works (given at least two nodes in the db it creates two nodes with label Test). Make sure you've got multistatments enabled (setting in the sidebar)
```
:auto MATCH (n)
WITH n
LIMIT 1
CREATE (t:Test {zid: id(n)})
RETURN t;
:auto MATCH (n)
WITH n
SKIP 1
LIMIT 1
CREATE (t:Test {zid: id(n)})
RETURN t;
```

The customer who reported the issue also tested it in production for the issue they ran into and confirmed the solution.

### DIFF
--- a/e2e_tests/integration/multistatements.spec.ts
+++ b/e2e_tests/integration/multistatements.spec.ts
@@ -153,10 +153,14 @@ RETURN t;`
       .contains('SUCCESS')
     cy.executeCommand('match (n: MultiStmtTest) return n.zid')
     cy.get('[role="cell"]').then(cells => {
-      cy.wrap(parseInt(cells[0].textContent || '')).should(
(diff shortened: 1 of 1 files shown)


jharris4:
This appears to fix at least the basic use cases, so was there anything we want to do to test that it works for any more complicated cases?

OskarDamkjaer:
Single statement :auto is tested in `loadcsv.spec.ts`, I'll add a test for multi-statement :auto as well

PR #1581, opened 2021-11-01
Update SSO to use 4.4 discovery and refresh tokens

OskarDamkjaer:

### DIFF
--- a/src/shared/modules/connections/connectionsDuck.ts
+++ b/src/shared/modules/connections/connectionsDuck.ts
@@ -102,7 +102,7 @@ export type Connection = {
   SSOProviders?: SSOProvider[]
 }
 
-const initialState: ConnectionReduxState = {
+export const initialState: ConnectionReduxState = {
   allConnectionIds: [],
   connectionsById: {},
   activeConnection: null,
@@ -451,7 +451,7 @@ export type DiscoverableData = {
   supportsMultiDb?: boolean
   host?: string
   encrypted?: string
-  hasForceUrl?: boolean
+  hasForceURL?: boolean
   SSOError?: string
   attemptSSOLogin?: boolean
   SSOProviders?: SSOProvider[]
@@ -505,7 +505,8 @@ export const startupConnectEpic = (action$: any, store: any) => {
       )
 
       if (
-        !(discovered && discovered.hasForceUrl) && // If we have force url, don't try old connection data
+        //TODO Considder the SSO implications of this
+        !(discovered && discovered.hasForceURL) && // If we have force url, don't try old connection data
         shouldTryAutoconnecting(savedConnection)
       ) {
         try {
--- a/src/shared/modules/connections/connectionsDuck.ts
+++ b/src/shared/modules/connections/connectionsDuck.ts
@@ -637,19 +637,67 @@ export const connectionLostEpic = (action$: any, store: any) =>
     // Only retry in web env and if we're supposed to be connected
     .filter(() => inWebEnv(store.getState()) && isConnected(store.getState()))
     .throttleTime(5000)
-    .do((action: any) => {
+    .do(() => store.dispatch(updateConnectionState(PENDING_STATE)))
+    .mergeMap((action: any) => {
       if (action.error.code === 'Neo.ClientError.Security.TokenExpired') {
         const SSOProviders = getActiveConnectionData(store.getState())
           ?.SSOProviders
         if (SSOProviders) {
-          handleRefreshingToken(SSOProviders).then(credentials => {
-            store.dispatch(discovery.updateDiscoveryConnection(credentials))
-          })
+          return Rx.Observable.fromPromise(
+            (async function() {
+              //TODO can throw?
+              const credentials = await handleRefreshingToken(SSOProviders)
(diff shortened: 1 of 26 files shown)


src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -513,6 +519,7 @@ export const startupConnectEpic = (action$: any, store: any) => {
       // merge with discovery data if we have any and try again
       if (discovered) {
         store.dispatch(discovery.updateDiscoveryConnection(discovered))
+        console.log('discovered: ' + JSON.stringify(discovered))
```
jharris4, line 522: console.log to be removed here?

src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -532,7 +539,8 @@ export const startupConnectEpic = (action$: any, store: any) => {
                 store.dispatch(setActiveConnection(discovery.CONNECTION_ID))
                 resolve({ type: STARTUP_CONNECTION_SUCCESS })
               })
-              .catch(() => {
+              .catch(err => {
+                console.log('error: ' + JSON.stringify(err))
```
jharris4, line 543: console.log to be removed here?

src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -630,7 +638,66 @@ export const connectionLostEpic = (action$: any, store: any) =>
     .filter(() => inWebEnv(store.getState()) && isConnected(store.getState()))
     .throttleTime(5000)
     .do(() => store.dispatch(updateConnectionState(PENDING_STATE)))
-    .mergeMap(() => {
+    .mergeMap((action: any) => {
+      if (action.error.code === 'Neo.ClientError.Security.TokenExpired') {
```
jharris4, line 642: we're planning to refactor so this code is not duplicated, right?

src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -630,7 +638,66 @@ export const connectionLostEpic = (action$: any, store: any) =>
...
+        const SSOProviders = getActiveConnectionData(store.getState())
+          ?.SSOProviders
+        if (SSOProviders) {
+          return Rx.Observable.fromPromise(
+            (async function() {
+              //TODO can throw?
```
jharris4, line 648: yeah, if the `handleRefreshingToken` throws, we should bail out of the reconnection attempts, and go straight to the connection lost state. TBD is whether we should specifically report this back to the user somehow, possible with custom error text?

src/shared/modules/discovery/discoveryDuck.ts
```diff
@@ -164,111 +158,57 @@ export const discoveryOnStartupEpic = (some$: any, store: any) => {
...
-
-      if (action.forceURL) {
-        const { username, protocol, host } = getUrlInfo(action.forceURL)
-
-        const discovered = {
-          username,
-          requestedUseDb: action.requestedUseDb,
-          host: `${protocol ? `${protocol}//` : ''}${host}`,
-          supportsMultiDb: !!action.requestedUseDb,
-          encrypted: action.encrypted,
```
jharris4, line 188: I'm a bit fuzzy on while all these discovered properties are needed, and whether they're actually used anywhere, but I guess it's fine for now.

src/shared/modules/connections/connectionsDuck.ts
OskarDamkjaer: done

src/shared/modules/connections/connectionsDuck.ts
OskarDamkjaer: yes

src/shared/modules/discovery/discoveryHelpers.test.ts
```diff
@@ -0,0 +1,178 @@
...
+    expect(discoveryData).toBeTruthy()
+    expect(discoveryData?.host).toEqual(boltHost)
+    expect(discoveryData?.source).toEqual(DISCOVERY_ENDPOINT)
+    expect(discoveryData?.SSOProviders).toEqual([])
+    expect(discoveryData?.SSOError).toEqual(undefined)
+    expect(discoveryData?.neo4jVersion).toEqual(neo4jVersion)
+    expect(discoveryData?.urlMissing).toEqual(false)
+  })
+
+  test('finds and priotises sso providers from session storage/connect form when all discovery sources are present, but doesnt merge when hosts differ', async () => {
```
jharris4, line 90: should probably add the new discoveryConnection source to this test?

src/shared/modules/discovery/discoveryDuck.ts
OskarDamkjaer: Yeah this turned into a bit of a rabbit hole trying to understand where they're all used. I'm somewhat certain that the encrypted key is not used here and that the restApi is used for the :http command. But didn't have enough time to dig into it fully

jharris4 (approved):
This is working really well in testing so far, so I'm going to merge it so we can have a release candidate.

We may want to address some of the minor remaining review comments later though...

PR #1582, opened 2021-11-02
Neo4j 4.4 config_clients.allow_telemetry

martin-neotech:
New configuration to opt out from product analytics by default.


OskarDamkjaer:
If it's set to off in neo4j conf, it can't be set to true in browser so we could drop the `, per default`

AlexicaWright (approved):
LGTM 👍  Thank you @martin-neotech

PR #1583, opened 2021-11-02
Fix issue where fullscreen change didn't cause re-render

OskarDamkjaer:
A performance optimization caused the VisualizationView.tsx not to rerender as the fullscreen prop changed. Causing this issue:

https://user-images.githubusercontent.com/10564538/139863358-d61f4741-6a25-4f9f-aca8-cf91aa44fc5a.mp4

Preview of the fix @ http://zoombuttons_fullscreen_issue.surge.sh

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
+++ b/src/browser/modules/Stream/CypherFrame/VisualizationView.tsx
@@ -53,6 +53,7 @@ export class Visualization extends Component<any, VisualizationState> {
   shouldComponentUpdate(props: any, state: VisualizationState) {
     return (
       this.props.updated !== props.updated ||
+      this.props.fullscreen !== props.fullscreen ||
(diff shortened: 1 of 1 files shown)


FallingCeilingS (approved):
I haven't pulled the code locally but the fix looks reasonable to me.

A general question: do we plan to refactor our class components to functional components as possible to separate container and presentational components in the future?

jharris4:
@FallingCeilingS choosing class vs functional components is mostly a matter of style, but also in some cases using one leads to cleaner & simpler code than the other.

For working on style improvements in the future, we would like to refactor some of our components to be "dumb" instead of "smart".

"dumb" in this context that all data or callbacks are passed in as props, and any redux type logic is separated into wrapping "smart" components.

jharris4 (approved):
LGTM

PR #1584, opened 2021-11-02
Changed show more links from span to button for accessibility

eijawerner:
also since drawer background is always dark no matter theme, then use dark theme on ShowAllOrMore to make sure link color has better visibility.

### DIFF
--- a/src/browser/modules/D3Visualization/components/OverviewPane.test.tsx
+++ b/src/browser/modules/D3Visualization/components/OverviewPane.test.tsx
@@ -53,7 +53,7 @@ describe('<OverviewPane />', () => {
 
   const getRelTypes: (length: number) => GraphStatsRelationshipTypes = length =>
     Array.from({ length: length }).reduce(
-      (acc: GraphStatsLabels, _current, index) => {
+      (acc: GraphStatsRelationshipTypes, _current, index) => {
         return {
           ...acc,
           ['relType' + index]: { count: 1, properties: { key: 'abc' } }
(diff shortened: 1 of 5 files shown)


src/browser/modules/D3Visualization/components/OverviewPane.test.tsx
```diff
@@ -53,7 +53,7 @@ describe('<OverviewPane />', () => {
 
   const getRelTypes: (length: number) => GraphStatsRelationshipTypes = length =>
     Array.from({ length: length }).reduce(
-      (acc: GraphStatsLabels, _current, index) => {
+      (acc: GraphStatsRelationshipTypes, _current, index) => {
```
OskarDamkjaer, line 56: Nice that you remembered to get this in

src/browser/modules/DBMSInfo/MetaItems.tsx
```diff
@@ -152,12 +154,14 @@ const LabelItems = ({
       <DrawerSectionBody className={wrapperStyle}>
         {labelItems}
       </DrawerSectionBody>
-      <ShowMoreOrAll
-        total={totalNumItems}
-        shown={labels.length}
-        moreStep={moreStep}
-        onMore={onMoreClick}
-      />
+      <ThemeProvider theme={dark}>
```
OskarDamkjaer, line 157: At first I was thinking we'd want to pass the theme as a prop to avoid using a ThemeProvider but then we'd have to add a prop to ShowMorOrAll, so I thnk this solution works better

OskarDamkjaer (approved):

jharris4 (approved):
LGTM

PR #1588, opened 2021-11-08
Animate opening/closing of Node Inspector

eijawerner:
also cleanup of Navigation.tsx, removed unused stuff and fixed so showing content also on close.

### DIFF
--- a/src/browser/components/TabNavigation/Navigation.test.tsx
+++ b/src/browser/components/TabNavigation/Navigation.test.tsx
@@ -58,7 +58,7 @@ describe('<Navigation />', () => {
     const { getByTestId, queryByTestId, rerender } = render(
       <Navigation
         onNavClick={onNavClick}
-        openDrawer={openDrawer}
+        selectedDrawerName={openDrawer}
         topNavItems={topNavItems}
       />
     )
@@ -70,45 +70,13 @@ describe('<Navigation />', () => {
     rerender(
       <Navigation
         onNavClick={onNavClick}
-        openDrawer={openDrawer}
+        selectedDrawerName={openDrawer}
         topNavItems={topNavItems}
       />
     )
 
     // expect documents drawer to be open
     expect(queryByTestId('documents-content')).toBeTruthy()
-  })
-
-  it('should close drawer when button is clicked on open drawer', () => {
-    let openDrawer = 'documents'
-    const onNavClick = (clickedDrawer: any) => {
-      openDrawer = toggleDrawer(openDrawer, clickedDrawer)
-    }
-
-    // render with documents drawer open
-    const { getByTestId, queryByTestId, rerender } = render(
-      <Navigation
-        onNavClick={onNavClick}
-        openDrawer={openDrawer}
-        topNavItems={topNavItems}
-      />
-    )
-
-    // click documents button
-    fireEvent.click(getByTestId('documents-icon'))
-
-    // rerender with updated openDrawer value
-    rerender(
-      <Navigation
-        onNavClick={onNavClick}
-        openDrawer={openDrawer}
-        topNavItems={topNavItems}
-      />
-    )
-
-    // expect drawer to be closed
-    expect(queryByTestId('documents-content')).toBeNull()
-    expect(queryByTestId('dbms-content')).toBeNull()
   })
 
   it('should switch drawer when different button is clicked than currently open', () => {
@@ -121,7 +89,7 @@ describe('<Navigation />', () => {
     const { getByTestId, queryByTestId, rerender } = render(
       <Navigation
         onNavClick={onNavClick}
-        openDrawer={openDrawer}
+        selectedDrawerName={openDrawer}
         topNavItems={topNavItems}
       />
     )
@@ -133,7 +101,7 @@ describe('<Navigation />', () => {
     rerender(
       <Navigation
         onNavClick={onNavClick}
-        openDrawer={openDrawer}
+        selectedDrawerName={openDrawer}
         topNavItems={topNavItems}
       />
     )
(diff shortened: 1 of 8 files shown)


src/browser/components/TabNavigation/Navigation.tsx
```diff
@@ -94,37 +87,44 @@ class Navigation extends Component<NavigationProps, NavigationState> {
...
+    let newTransitionState: DrawerTransitionState = prevState.transitionState
+    if (prevProps.selectedDrawerName !== this.props.selectedDrawerName) {
+      if (this.props.selectedDrawerName) {
         if (
           prevState.transitionState === Closed ||
           prevState.transitionState === Closing
         ) {
-          newState.transitionState = Opening
+          newTransitionState = Opening
+          lastSelectedDrawerName = null
```
OskarDamkjaer, line 99: Personal preference but I find this easier to read as two setState calls rather than creating and mutating local variables and then only calling once

src/browser/components/TabNavigation/Navigation.test.tsx
```diff
@@ -79,38 +79,6 @@ describe('<Navigation />', () => {
     expect(queryByTestId('documents-content')).toBeTruthy()
   })
 
-  it('should close drawer when button is clicked on open drawer', () => {
```
OskarDamkjaer, line 82: Note: this is left out as it's better tested in e2e tests

src/browser/components/TabNavigation/Navigation.tsx
```diff
@@ -156,48 +159,47 @@ class Navigation extends Component<NavigationProps, NavigationState> {
...
+      topNavItems,
+      this.props.selectedDrawerName
+    )
     const bottomNavItemsList = buildNavList(
       bottomNavItems,
-      this.state.drawerContent
+      this.props.selectedDrawerName
     )
 
+    const isDrawerVisible = () => this.state.transitionState !== Closed
```
OskarDamkjaer, line 181: We could simplify this to be a constant rather than a function, right? `const drawerIsVisible = this.state.transitionState !== Closed`. WDYT?

src/browser/components/TabNavigation/Navigation.tsx
```diff
@@ -156,48 +159,47 @@ class Navigation extends Component<NavigationProps, NavigationState> {
...
+
     return (
       <StyledSidebar>
         <StyledTabsWrapper>
           <StyledTopNav>{topNavItemsList}</StyledTopNav>
           <StyledBottomNav>{bottomNavItemsList}</StyledBottomNav>
         </StyledTabsWrapper>
-        <StyledDrawer
-          width={width}
-          ref={ref => {
```
OskarDamkjaer, line 191: nice that we could get rid of this 🙏

src/browser/components/TabNavigation/Navigation.tsx
```diff
@@ -156,48 +159,47 @@ class Navigation extends Component<NavigationProps, NavigationState> {
...
-              ref.removeEventListener('transitionend', this._onTransitionEnd)
-              ref.addEventListener('transitionend', this._onTransitionEnd)
-            }
-          }}
-        >
-          {getContentToShow(this.state.drawerContent)}
+        <StyledDrawer width={width} onTransitionEnd={this.onTransitionEnd}>
+          {isDrawerVisible() &&
+            getContentToShow(
+              this.props.selectedDrawerName || this.state.lastSelectedDrawerName
```
OskarDamkjaer, line 201: Not sure how, but it'd be nice if we could explain why lastSelectedDrawerName is needed. Perhaps a comment? Or some other name? Thoughts?

src/browser/modules/Sidebar/Sidebar.tsx
```diff
@@ -176,7 +176,7 @@ const Sidebar = ({
 
   return (
     <TabNavigation
-      openDrawer={openDrawer}
+      selectedDrawerName={openDrawer}
```
OskarDamkjaer, line 179: It'd be nice to rename openDrawer to selectedDrawerName in the rest of the code as well 🤔 something we could do in a future PR

src/browser/modules/D3Visualization/components/NodeInspectorPanel.tsx
```diff
@@ -70,44 +68,40 @@ export class NodeInspectorPanel extends Component<NodeInspectorPanelProps> {
           )}
         </StyledNodeInspectorTopMenuChevron>
 
-        {expanded && (
-          <StyledNodeInspectorContainer
+        <NodeInspectorDrawer width={width} openDrawer={expanded}>
```
OskarDamkjaer, line 71: Got a bit confused here because we've used "openDrawer" to be a drawer name in the past, anyway I think `isOpen` would be a better name, but openDrawer also works fine

OskarDamkjaer (approved):
Happy to see these changes! Opening/closing feels much smoother now. I've left a few thoughts but don't think any changes are required

src/browser/components/TabNavigation/Navigation.tsx
eijawerner: yes I was thinking of naming it something with closing, like closingDrawerName, since it is only used when a drawer is closing to keep the content until width is 0 and it is totally invisible. WDYT?

OskarDamkjaer: It's somewhat long but calling it `drawerToShowWhileClosing` could do the trick?

OskarDamkjaer: I'll leave you to pick the name

PR #1589, opened 2021-11-08
Improve connection validation performance

FallingCeilingS:
Previously, we used `CALL db.indexes` to verify that a connection is still valid. This can be expensive, so we replace the existing logic of `validateConnection` by calling `driver.verifyConnectivity()`. Also, this PR includes some type fixes.
More type fixes can be addressed via subsequent PRs to ensure this PR is small.

### DIFF
--- a/src/shared/services/bolt/boltConnection.ts
+++ b/src/shared/services/bolt/boltConnection.ts
@@ -77,9 +77,7 @@ export const validateConnection = (
           })
           .catch((e: { code: string; message: string }) => {
             session.close()
-            // Only invalidate the connection if not available
-            // or not authed
-            // or credentials have expired
+            // Only invalidate bolt connection error
             if (!e.code || isBoltConnectionErrorCode(e.code)) {
               rej(e)
             } else {
--- a/e2e_tests/integration/multistatements.spec.ts
+++ b/e2e_tests/integration/multistatements.spec.ts
@@ -41,6 +41,9 @@ describe('Multi statements', () => {
   })
 
   it('can run multiple statements (non open by default)', () => {
+    // Avoid the problem when executing the script under system database:
+    // Neo.DatabaseError.Statement.ExecutionFailed: Not a recognised system command or procedure. This Cypher command can only be executed in a user database: RETURN 1 AS `1`
+    cy.executeCommand(':use neo4j')
     cy.executeCommand(':clear')
     cy.executeCommand(validQuery)
     cy.get('[data-testid="frame"]', { timeout: 10000 }).should('have.length', 1)
--- a/e2e_tests/integration/multistatements.spec.ts
+++ b/e2e_tests/integration/multistatements.spec.ts
@@ -43,7 +43,7 @@ describe('Multi statements', () => {
   it('can run multiple statements (non open by default)', () => {
     // Avoid the problem when executing the script under system database:
     // Neo.DatabaseError.Statement.ExecutionFailed: Not a recognised system command or procedure. This Cypher command can only be executed in a user database: RETURN 1 AS `1`
-    cy.executeCommand(':use neo4j')
+    cy.useNeo4jDatabase()
     cy.executeCommand(':clear')
     cy.executeCommand(validQuery)
     cy.get('[data-testid="frame"]', { timeout: 10000 }).should('have.length', 1)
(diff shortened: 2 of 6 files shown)


src/shared/services/bolt/boltConnection.ts
```diff
@@ -50,44 +49,24 @@ export const hasMultiDbSupport = async (): Promise<boolean> => {
...
-            // or credentials have expired
-            if (!e.code || isBoltConnectionErrorCode(e.code)) {
-              rej(e)
-            } else {
-              res(driver)
-            }
-          })
+    .verifyConnectivity()
+    .then((result: ServerInfo) => {
+      console.log(result)
```
OskarDamkjaer, line 63: Console.log leftover from debugging?

src/shared/services/bolt/boltConnection.ts
```diff
@@ -50,44 +49,24 @@ export const hasMultiDbSupport = async (): Promise<boolean> => {
...
+      resolve(driver)
+    })
+    .catch(e => {
+      if (!e.code || isBoltConnectionErrorCode(e.code)) reject(e)
```
OskarDamkjaer, line 66: I think it'd be good to keep the comment note about why we only care about isBoltConnectionErrorCOde

OskarDamkjaer: Also reading through the driver source to see if we still need to do this check

OskarDamkjaer: Looking at the driver code -> https://github.dev/neo4j/neo4j-javascript-driver/blob/6388d879ceb2b0dba91c2551bfe0868c16694ae8/packages/core/src/driver.ts 
it seems this method can throw without setting it's code to any of the ones we categorize as "boltConnectionErrorCodes". I'm starting to think we should always reject on error, regardless of errorcode

FallingCeilingS:
E2E tests are failing. Need to investigate why it happened.

src/shared/services/bolt/boltConnection.ts
```diff
@@ -50,44 +49,25 @@ export const hasMultiDbSupport = async (): Promise<boolean> => {
...
-            // Only invalidate the connection if not available
-            // or not authed
-            // or credentials have expired
-            if (!e.code || isBoltConnectionErrorCode(e.code)) {
-              rej(e)
-            } else {
-              res(driver)
-            }
-          })
+    .verifyConnectivity()
```
jharris4, line 63: Unfortunately it seems that something in the session creation with the database parameter of system is having side effects. With this change running some of the tests results in a "NotSystemDatabase" type of error when running commands when Browser thinks it is using the system database.

I'm not sure why this is happening, but for now I think we need to revert to the old logic, and just switch try switching the query from `CALL db.indexes()` to `CALL db.ping()`

FallingCeilingS: Tested printing out `driver.session()`. It seems like in both cases they're the same:
- In `master` branch:
Screenshot 2021-11-09 at 11 47 50
- In this PR before reverting the changes:
Screenshot 2021-11-09 at 11 50 39

This is the problem we have no idea yet. Will probably revisit it in the future.

FallingCeilingS:
Document an interesting behaviour here (not sure if anyone else can replicate):
- Serve docker for database `docker run --env NEO4J_AUTH=neo4j/newpassword --env=NEO4J_ACCEPT_LICENSE_AGREEMENT=yes -p7474:7474 -p7687:7687 neo4j:4.3.6-enterprise`.
- Run browser front-end.
- Execute command

```cql
DROP user noroles
```
and
```cql
CREATE USER noroles SET PASSWORD "pw" CHANGE REQUIRED
```

- Disconnect with the DB and log in to `noroles`.

When we call `CALL db.indexes()`, an exception is thrown and then a new frame is added to force user change the current password:

https://github.com/neo4j/neo4j-browser/blob/325b8de9746b6338b36724984f431cba778d11de/src/shared/services/bolt/boltConnection.ts#L61-L91

Screenshot 2021-11-11 at 10 44 08

If we changed the DB call to `CALL db.ping()`, the behaviour is different. Please refer to the video between 1:30 to 1:31. There's a yellow warning bar appeared shortly showing "Connection to server lost. Reconnecting...".

https://user-images.githubusercontent.com/26452483/141284719-93e52a38-c766-4713-b85c-9301c0769eef.mp4

Screenshot 2021-11-11 at 10 53 15

And the exception can be caught by one of the Promises. But no frame pops up to ask the user to update the password.

Screenshot 2021-11-11 at 10 50 48

It needs more investigation to figure out why it's happening. But it's not the highest priority for now.

OskarDamkjaer:
Let's reopen this when/if we decide to spend time investigating the issues we ran into

PR #1591, opened 2021-11-09
Fix typescript issues

eijawerner:

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/CodeView.test.tsx
+++ b/src/browser/modules/Stream/CypherFrame/CodeView.test.tsx
@@ -67,7 +67,7 @@ describe('CodeViews', () => {
   describe('CodeStatusbar', () => {
     test('displays no statusBarMessage', () => {
       // Given
-      const props = { result: {}, maxRows: 0 }
+      const props = { result: null, maxRows: 0, maxFieldItems: 0 }
 
       // When
       const { container } = render(<CodeStatusbar {...props} />)
@@ -84,10 +84,11 @@ describe('CodeViews', () => {
           summary: {
             resultAvailableAfter: neo4j.int(5),
             resultConsumedAfter: neo4j.int(5)
-          },
-          maxRows: 100,
-          records: [{ res: 'xx3' }]
-        }
+          } as any,
+          records: [{ res: 'xx3' }] as any
+        },
+        maxRows: 100,
+        maxFieldItems: 0
       }
 
       // When
--- a/src/browser/modules/Stream/CypherFrame/relatable-view.tsx
+++ b/src/browser/modules/Stream/CypherFrame/relatable-view.tsx
@@ -16,7 +16,7 @@
  */
 
 import React, { useMemo } from 'react'
-import { isInt } from 'neo4j-driver'
+import { isInt, QueryResult } from 'neo4j-driver'
 import Relatable from '@relate-by-ui/relatable'
 import { get, head, map, slice } from 'lodash-es'
 import { Icon } from 'semantic-ui-react'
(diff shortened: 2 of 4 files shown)


src/browser/modules/Stream/CypherFrame/CodeView.test.tsx
```diff
@@ -84,10 +84,11 @@ describe('CodeViews', () => {
           summary: {
             resultAvailableAfter: neo4j.int(5),
             resultConsumedAfter: neo4j.int(5)
-          },
-          maxRows: 100,
-          records: [{ res: 'xx3' }]
-        }
+          } as any,
```
FallingCeilingS, line 87: Do we need `as any` in such places? I don't see TSlint warning without this.

src/browser/modules/Stream/CypherFrame/relatable-view.tsx
```diff
@@ -44,6 +44,8 @@ import {
 } from './relatable-view.styled'
 import { stringifyMod, unescapeDoubleQuotesForDisplay } from 'services/utils'
 import { GlobalState } from 'shared/globalState'
+import { BrowserRequestResult } from 'project-root/src/shared/modules/requests/requestsDuck'
+import Record from 'neo4j-driver-core/types/record'
```
OskarDamkjaer, line 48: any reason we've not included this in the import on line 19?

src/browser/modules/Stream/CypherFrame/CodeView.test.tsx
eijawerner: yes I get typescript errors otherwise. Probably we need to sync our settings then.

src/browser/modules/Stream/CypherFrame/relatable-view.tsx
```diff
@@ -52,14 +54,24 @@ const RelatableView = connect((state: GlobalState) => ({
...
   maxRows,
-  result,
-  maxFieldItems
-}: any) {
-  const { records = [] } = result
+  maxFieldItems,
+  result
+}: RelatableViewComponentProps): JSX.Element | null {
+  const records = useMemo(
+    () => (result ? (result as QueryResult).records || [] : []),
```
OskarDamkjaer, line 69: I'd prefer if we'd avoid casting the type like so 
```typescript
() => (result && 'records' in result ? result.records : []),
```

src/shared/services/utils.ts
```diff
@@ -312,10 +312,10 @@ export const arrayToObject = (array: any) =>
 export const stringifyMod = (
   value: any,
   modFn: any = null,
-  pretty: any = false,
+  pretty: boolean | number = false,
   skipOpeningIndentation = false
-): any => {
-  const prettyLevel = !pretty ? false : pretty === true ? 1 : parseInt(pretty)
+): string => {
+  const prettyLevel = !pretty ? false : pretty === true ? 1 : +pretty
```
OskarDamkjaer, line 318: Thinking this'd be easier to follow as `  const prettyLevel = isNumber(pretty) ? pretty : pretty ? 1 : 0`

but the whole method is pretty messy anyhow

OskarDamkjaer (approved):
Have a look through my comments to see if there's anything you agree with, on the whole good to go imo!

src/browser/modules/Stream/CypherFrame/relatable-view.tsx
```diff
@@ -91,7 +103,10 @@ function getColumns(records: any, maxFieldItems: any) {
   }))
 }
 
-function CypherCell({ cell }: any) {
+type CypherCellProps = {
+  cell: any
```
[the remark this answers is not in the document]
eijawerner: I tried to do that, but I basically ended up with almost any type, since StringifyMod accepts basically anything. Probably we need to rewrite the code to make it more explicitly clear what types that are accepted here but I do feel that it requires someone with more insight into the code at this moment.

src/browser/modules/Stream/CypherFrame/CodeView.test.tsx
FallingCeilingS: @eijawerner - my bad. Checked with your branch and it should be typed anyway.

PR #1592, opened 2021-11-10
Cut long property values

eijawerner:
Cut long property values in the Node properties panel and possible to expand by clicking 'Show all'.

Screenshot 2021-11-17 at 14 57 21

### DIFF
--- a/src/browser/modules/D3Visualization/components/DetailsPane.test.tsx
+++ b/src/browser/modules/D3Visualization/components/DetailsPane.test.tsx
@@ -23,7 +23,9 @@ import { render, screen, waitFor } from '@testing-library/react'
 import {
   DETAILS_PANE_STEP_SIZE,
   DetailsPaneComponent,
-  MAX_LENGTH_SMALL
+  MAX_LENGTH_LARGE,
+  MAX_LENGTH_SMALL,
+  WIDE_VIEW_THRESHOLD
 } from './DetailsPane'
 import { VizItem, VizNodeProperty } from './types'
 
@@ -50,11 +52,13 @@ describe('<DetailsPane />', () => {
     properties?: VizItem[]
     labels?: string[]
     type?: 'node' | 'relationship'
+    width?: number
   }
   const renderComponent = ({
     properties = [],
     labels = [],
-    type = 'node'
+    type = 'node',
+    width = 200
   }: RenderComponentProps) => {
     let mockVizItem: VizItem
     switch (type) {
@@ -82,7 +86,7 @@ describe('<DetailsPane />', () => {
         frameHeight={0}
         graphStyle={mockGraphStyle}
         vizItem={mockVizItem}
-        nodeInspectorWidth={200}
+        nodeInspectorWidth={width}
       />
     )
   }
@@ -136,7 +140,10 @@ describe('<DetailsPane />', () => {
       type: 'string',
       value: fullText
     }
-    renderComponent({ properties: [mockProperty] })
+    renderComponent({
+      properties: [mockProperty],
+      width: WIDE_VIEW_THRESHOLD - 1
+    })
 
     const expectedCutValue = fullText.slice(0, MAX_LENGTH_SMALL) + '...'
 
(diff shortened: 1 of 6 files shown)


src/browser/modules/D3Visualization/components/DetailsPane.test.tsx
```diff
@@ -123,4 +128,36 @@ describe('<DetailsPane />', () => {
       screen.getByRole('button', { name: 'Show 2 more' })
     ).toBeInTheDocument()
   })
+
+  test('should handle show more on long property value', async () => {
```
jharris4, line 132: It's great that you've added this test!

Do you think it's worth adding tests for values both larger and smaller than the `WIDE_VIEW_THRESHOLD` ?

Maybe one test would do `nodeInspectorWidth={WIDE_VIEW_THRESHOLD - 1}` and the other with `WIDE_VIEW_THRESHOLD + 1` or something?

eijawerner: Sounds like a good idea. I'll add a test for that.

src/browser/modules/D3Visualization/components/styled.tsx
```diff
@@ -371,3 +371,10 @@ export const PaneBodySectionHeaderWrapper = styled.div`
   display: flex;
   flex-direction: column;
 `
+
+export const StyledExpandValueButton = styled.button`
```
OskarDamkjaer, line 375: Can't help but think we must have this exact component somewhere else also 🤔

OskarDamkjaer: Found it! It's StyledShowMoreButton! Perhaps we could move that and reuse it? 🤔  Not sure it's the right move to de-duplicate when there's only two uses, let me know what you think

OskarDamkjaer:
It'd be nice to have a screenshot and a sentence describing the change in the PR description to give outside contributors (and ourselves in the future) some context to understand the PR

src/browser/modules/D3Visualization/components/DetailsPane.tsx
```diff
@@ -39,17 +40,54 @@ import { StyleableRelType } from './StyleableRelType'
...
+export const MAX_LENGTH_LARGE = 300
+type ExpandableValueProps = {
+  value: string
+  type: string
+  wideMode: boolean
+}
+function ExpandableValue({ value, wideMode }: ExpandableValueProps) {
+  const [expanded, setExpanded] = useState(false)
+
+  const maxLength = wideMode ? MAX_LENGTH_LARGE : MAX_LENGTH_SMALL
```
OskarDamkjaer, line 53: I think it's nice to have the max string length be dependant on how much space is available, but I was thinking it'd be easier & simpler if we calculated some maxLength from the width? So rather than having a wideMode prop we could send just the "width" and perhaps say maxLenght = width/3 or something like that? Would get rid of a few constants and be more flexible. WDYT?

src/browser/modules/D3Visualization/components/DetailsPane.tsx
```diff
@@ -39,17 +40,54 @@ import { StyleableRelType } from './StyleableRelType'
...
+
+  const handleExpandClick = () => {
+    setExpanded(true)
+  }
+
+  let valueShown = expanded
+    ? value
+    : value.slice(0, Math.min(maxLength, value.length))
+  const valueIsTrimmed = valueShown.length !== value.length
+  valueShown += valueIsTrimmed ? '...' : ''
```
OskarDamkjaer, line 63: There is a unicode character for elipsis we could use here "…" not sure if it's preferred, but it's an option at least!

src/browser/modules/D3Visualization/components/DetailsPane.tsx
```diff
@@ -39,17 +40,54 @@ import { StyleableRelType } from './StyleableRelType'
...
+
+  const maxLength = wideMode ? MAX_LENGTH_LARGE : MAX_LENGTH_SMALL
+
+  const handleExpandClick = () => {
+    setExpanded(true)
+  }
+
+  let valueShown = expanded
+    ? value
+    : value.slice(0, Math.min(maxLength, value.length))
```
OskarDamkjaer, line 61: From testing it doesn't matter if slice is called with a value that's too big, so we could do `value.slice(0, maxLength)`. Doing it that way is a bit simpler but maybe less clear. I'll leave it to your preference to pick

src/browser/modules/D3Visualization/components/DetailsPane.tsx
```diff
@@ -39,17 +40,54 @@ import { StyleableRelType } from './StyleableRelType'
 import { upperFirst } from 'services/utils'
 import { ShowMoreOrAll } from 'browser-components/ShowMoreOrAll/ShowMoreOrAll'
 
+export const MAX_LENGTH_SMALL = 150
+export const MAX_LENGTH_LARGE = 300
+type ExpandableValueProps = {
+  value: string
+  type: string
```
OskarDamkjaer, line 47: This prop seems unused

OskarDamkjaer (commented):
I think it was the right move to create a nested component, easier to test and easier to make re-usable if we need to in the future! Left a few comments, do have a look and tell me what you think

src/browser/modules/D3Visualization/components/styled.tsx
eijawerner: yes I thought about it, but I think it is not necessary for now given the simplicity of the code. I think if we should do a general component it could be a ButtonStyledAsLink or something where also the color is sent it for a more useful reusability.

src/browser/modules/D3Visualization/components/DetailsPane.tsx
eijawerner: Sure!

src/browser/modules/D3Visualization/components/DetailsPane.tsx
eijawerner: sure, since slice is such a common method and it is clear in the documentation of that that it will be the whole string in case the end is larger than the length I think it is better to remove the Math.min for simplicity.

src/browser/modules/D3Visualization/components/DetailsPane.tsx
eijawerner: yes, you are right, it is from the future, is going to be used in my next PR :D. Make sense to remove it for now though. 👍

src/browser/modules/D3Visualization/components/DetailsPane.tsx
OskarDamkjaer: For a followup PR

OskarDamkjaer (approved):

PR #1594, opened 2021-11-11
Local Storage for retaining remote guides

FallingCeilingS:
We are able to load remote guides. However, when reloading the Browser, added remote guides are lost. Thus, we need to retain remote guides in `localStorage`. 
Also, to keep the size of cache small, we aim not to store contents of guides. Same with Redux states. Instead, when we click the remote guide, we fetch from remote server.
This PR also involves some code clean up in relevant areas.

Test links:
- https://guides.neo4j.com/got
- https://guides.neo4j.com/restaurant_recommendation
- https://guides.neo4j.com/sandbox/movies/index.html

More guide links:
https://neo4j.com/developer/guide-create-neo4j-browser-guide/

Further code cleanup around the affected area and relevant unit tests and E2E tests will be handled in subsequent PRs.

### DIFF
--- a/src/browser/modules/Sidebar/GuideDrawer.tsx
+++ b/src/browser/modules/Sidebar/GuideDrawer.tsx
@@ -29,9 +29,9 @@ import {
   FetchGuideAction,
   UpdateGuideAction,
   getCurrentGuide,
-  gotoSlide,
   getRemoteGuides,
   resetGuide,
+  gotoSlide,
   setCurrentGuide,
   fetchRemoteGuide,
   updateRemoteGuides
@@ -51,7 +51,7 @@ import GuidePicker from './GuidePicker'
 type GuideDrawerProps = {
   currentGuide: Guide | null
   remoteGuides: Guide[]
-  backToAllGuides: () => void
+  backToAllGuides: () => SetGuideAction
   gotoSlide: (slideIndex: number) => GotoSlideAction
   setCurrentGuide: (guide: Guide) => SetGuideAction
   fetchRemoteGuide: (identifier: string) => FetchGuideAction
--- a/src/browser/modules/Sidebar/GuideDrawer.tsx
+++ b/src/browser/modules/Sidebar/GuideDrawer.tsx
@@ -89,7 +89,7 @@ function GuideDrawer({
             {currentGuide.title}
           </GuideTitle>
           <GuideCarousel
-            slides={currentGuide.slides}
+            slides={currentGuide.slides ?? []}
             currentSlideIndex={currentGuide.currentSlide}
             gotoSlide={gotoSlide}
             scrollToTop={() =>
--- a/src/browser/modules/Sidebar/GuideDrawer.tsx
+++ b/src/browser/modules/Sidebar/GuideDrawer.tsx
@@ -24,6 +24,7 @@ import { connect } from 'react-redux'
 
 import {
   Guide,
+  RemoteGuide,
   getCurrentGuide,
   getRemoteGuides,
   resetGuide,
@@ -46,12 +47,12 @@ import GuidePicker from './GuidePicker'
 
 type GuideDrawerProps = {
   currentGuide: Guide | null
-  remoteGuides: Guide[]
+  remoteGuides: RemoteGuide[]
   backToAllGuides: () => void
   gotoSlide: (slideIndex: number) => void
   setCurrentGuide: (guide: Guide) => void
   fetchRemoteGuide: (identifier: string) => void
-  updateRemoteGuides: (newList: Guide[]) => void
+  updateRemoteGuides: (newList: RemoteGuide[]) => void
 }
 
 function GuideDrawer({
@@ -116,7 +117,7 @@ const mapDispatchToProps = (dispatch: Dispatch<Action>) => ({
   setCurrentGuide: (guide: Guide) => dispatch(setCurrentGuide(guide)),
   fetchRemoteGuide: (identifier: string) =>
     dispatch(fetchRemoteGuide(identifier)),
-  updateRemoteGuides: (updatedList: Guide[]) =>
+  updateRemoteGuides: (updatedList: RemoteGuide[]) =>
     dispatch(updateRemoteGuides(updatedList))
 })
 const ConnectedGuideDrawer = connect(
--- a/src/browser/modules/Sidebar/GuideDrawer.tsx
+++ b/src/browser/modules/Sidebar/GuideDrawer.tsx
@@ -18,17 +18,22 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { Dispatch, useRef } from 'react'
-import { Action } from 'redux'
+import React, { useRef } from 'react'
+import { Action, Dispatch } from 'redux'
 import { connect } from 'react-redux'
 
 import {
-  getCurrentGuide,
   Guide,
+  GotoSlideAction,
+  SetGuideAction,
+  FetchGuideAction,
+  UpdateGuideAction,
+  getCurrentGuide,
   gotoSlide,
   getRemoteGuides,
   resetGuide,
   setCurrentGuide,
+  fetchRemoteGuide,
   updateRemoteGuides
 } from 'shared/modules/guides/guidesDuck'
 import { GlobalState } from 'shared/globalState'
@@ -45,19 +50,21 @@ import GuidePicker from './GuidePicker'
 
 type GuideDrawerProps = {
   currentGuide: Guide | null
-  backToAllGuides: () => void
-  gotoSlide: (slideIndex: number) => void
   remoteGuides: Guide[]
-  setCurrentGuide: (guide: Guide) => void
-  updateRemoteGuides: (newList: Guide[]) => void
+  backToAllGuides: () => void
+  gotoSlide: (slideIndex: number) => GotoSlideAction
+  setCurrentGuide: (guide: Guide) => SetGuideAction
+  fetchRemoteGuide: (identifier: string) => FetchGuideAction
+  updateRemoteGuides: (newList: Guide[]) => UpdateGuideAction
 }
 
 function GuideDrawer({
   currentGuide,
+  remoteGuides,
   backToAllGuides,
   gotoSlide,
-  remoteGuides,
   setCurrentGuide,
+  fetchRemoteGuide,
   updateRemoteGuides
 }: GuideDrawerProps): JSX.Element {
   const scrollRef = useRef<HTMLDivElement>(null)
@@ -81,6 +88,7 @@ function GuideDrawer({
         <GuidePicker
           remoteGuides={remoteGuides}
           setCurrentGuide={setCurrentGuide}
+          fetchRemoteGuide={fetchRemoteGuide}
           updateRemoteGuides={updateRemoteGuides}
         />
       ) : (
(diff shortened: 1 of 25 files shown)


src/shared/modules/commands/helpers/play.ts
```diff
@@ -18,11 +18,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// Need to refactor
```
jharris4, line 21: what is this comment for? It's hard to tell without more detail...

jharris4:
@FallingCeilingS Really nice work on this, the code looks great, and nice cleanup along the way as well!

I'll want to test this out a little next week, and @OskarDamkjaer should probably have a look too since he knows this code really well.

src/shared/modules/commands/helpers/play.ts
FallingCeilingS: @jharris4 - sorry for the confusion. IMO the original `fetchRemoteGuide` API request has a conflict with the name with `fetchRemoteGuide` method that returns Redux action. I renamed it as `fetchRemoteGuideAsync` but not sure if it's semantic enough.
Also, the type needs to be specified in this file at some point.
This method seems related to `Guide`. So, I don't think it should be in `play.ts`. IMO either the file should be renamed like `guideApi.ts`, or consolidated into `guideResolverHelper.ts`. It helps enhance the readability.

src/shared/modules/guides/guidesDuck.ts
```diff
@@ -72,6 +86,32 @@ interface AddGuideAction {
...
+  GlobalState
+> = (action$, store$) =>
+  action$.ofType(FETCH_GUIDE).flatMap(action => {
+    const initialSlide = tryGetRemoteInitialSlideFromUrl(
+      (<FetchGuideAction>action).identifier
+    )
+    return Observable.fromPromise(
+      resolveGuide((<FetchGuideAction>action).identifier, store$.getState())
+    ).mergeMap(({ title, identifier, slides, isError }) => {
+      const guide: Guide = {
```
OskarDamkjaer, line 100: I've also been thinking a bit about if handling this with redux observable is the best way to go about things.. What we could do is this: store guides to localstorage (except the slides). Then when we open a guide that has slides null we treat that as if the slides are "loading", then we call the resolveGuide method updating the slide in redux and voloua!

OskarDamkjaer: It's get rid of a lot of moving parts, let's discuss later

OskarDamkjaer (approved):

OskarDamkjaer (changes requested):
Nice changes! Let's look through my comments together and see what's the best way forward is :D

src/shared/modules/commands/helpers/play.ts
FallingCeilingS: @OskarDamkjaer - how about `requests.ts.` or `api.ts` or `async.ts`? I'm wondering all API calls can be consolidated in one file in the future?

src/browser/modules/Sidebar/GuideDrawer.tsx
```diff
@@ -45,19 +50,21 @@ import GuidePicker from './GuidePicker'
 
 type GuideDrawerProps = {
   currentGuide: Guide | null
-  backToAllGuides: () => void
-  gotoSlide: (slideIndex: number) => void
   remoteGuides: Guide[]
-  setCurrentGuide: (guide: Guide) => void
-  updateRemoteGuides: (newList: Guide[]) => void
+  backToAllGuides: () => SetGuideAction
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - I tried to revert this change, but the TSlint started complaining:
Screenshot 2021-11-17 at 11 18 06
Screenshot 2021-11-17 at 11 16 46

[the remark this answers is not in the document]
OskarDamkjaer: Yeah the GuidePickers props also need to return void rather than actions

src/shared/services/guideResolverHelper.tsx
```diff
@@ -85,11 +92,15 @@ function htmlTextToSlides(html: string): JSX.Element[] {
   return [<Slide key="first" html={html} isSidebarSlide />]
 }
 
-async function resolveRemoteGuideFromURL(
-  guideName: string,
-  state: any
-): Promise<{ slides: JSX.Element[]; title: string; isError?: boolean }> {
-  const url = guideName
+async function resolveRemoteGuideByUrl(
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - as I've touched quite a few files, I'd prefer to include further clean up in subsequent PR including adding test cases.

src/shared/services/guideResolverHelper.tsx
```diff
@@ -43,27 +43,34 @@ import {
...
+  const isUrl = identifier.startsWith('http')
   if (isUrl) {
-    return await resolveRemoteGuideFromURL(guideName, state)
+    return await resolveRemoteGuideByUrl(identifier, state)
   }
 
-  if (isGuideChapter(guideName)) {
-    return chapters[guideName]
+  if (isGuideChapter(identifier)) {
+    return { ...chapters[identifier], identifier }
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - good point. I'd include this change in subsequent PR. Also, let's consider to consolidate duplicated document elements across our code-base in the future.

src/shared/services/guideResolverHelper.tsx
```diff
@@ -43,27 +43,34 @@ import {
...
   }
 
   try {
-    return await resolveRemoteGuideFromName(guideName, state)
+    return await resolveRemoteGuideByName(identifier, state)
   } catch (e) {}
 
-  return guideUnfound
+  return { ...guideUnfound, identifier }
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - same as https://github.com/neo4j/neo4j-browser/pull/1594#discussion_r751348321.

src/shared/modules/guides/guidesDuck.ts
```diff
@@ -95,10 +135,12 @@ export default function reducer(
       return { ...state, remoteGuides: action.updatedGuides }
 
     case ADD_GUIDE:
+      // Use title as the unique identifier doesn't seem correct
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - I'll do it in subsequent PR.

src/shared/modules/guides/guidesDuck.ts
```diff
@@ -72,6 +86,32 @@ interface AddGuideAction {
   guide: Guide
 }
 
+export const fetchRemoteGuideEpic: Epic<
+  FetchGuideAction | SetGuideAction | OpenAction | AddGuideAction,
+  GlobalState
+> = (action$, store$) =>
+  action$.ofType(FETCH_GUIDE).flatMap(action => {
+    const initialSlide = tryGetRemoteInitialSlideFromUrl(
+      (<FetchGuideAction>action).identifier
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - good point! We need to introduce `typesafe-actions` dependency in our project.

```ts
import { isOfType } from 'typesafe-actions';
```

I'd like to include it in the subsequent PR.

src/shared/modules/guides/guidesDuck.ts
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - as discussed offline, there's misunderstanding on this topic. I've already eliminated `slides` from `remoteGuides`.

src/shared/modules/guides/guidesDuck.ts
FallingCeilingS: @OskarDamkjaer - as discussed offline, we will leave the behaviour for now. Better to have an improved behaviour in the future.

src/shared/modules/commands/helpers/play.ts
FallingCeilingS: @OskarDamkjaer - as discussed offline, we agreed to keep modules as small as possible. So, we just rename it as `playAndGuides.ts`.

OskarDamkjaer: Settled on this being called playOrGuides.ts offline

src/shared/services/localstorage.ts
```diff
@@ -72,8 +76,10 @@ export function getAll(): Partial<GlobalState> {
           ...(current as typeof settingsInitialState),
           playImplicitInitCommands: true
         }
+      } else if (key === 'guides') {
+        out[key] = { ...(current as GuideState), currentGuide: null }
```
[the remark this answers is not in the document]
OskarDamkjaer: From offline discussion I learned: remoteGuides never have slides stored in redux, so we've updated the types to represent that. And we changed the code to not store guides in localstorage as well

src/shared/modules/commands/helpers/playAndGuides.ts
```diff
@@ -22,15 +22,18 @@ import { hostIsAllowed } from 'services/utils'
...
   return new Promise<void>((resolve, reject) => {
-    if (!hostIsAllowed(url, allowlist)) {
+    if (!hostIsAllowed(url, allowlistStr)) {
       return reject(
         new Error('Hostname is not allowed according to server allowlist')
       )
     }
     resolve()
-  }).then(() => {
+  }).then(async () => {
```
OskarDamkjaer, line 36: This is not needed right? We'll only need to add these when using await

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -505,32 +501,16 @@ const availableCommands = [
...
   {
     name: 'guide',
-    match: (cmd: any) => /^guide(\s|$)/.test(cmd),
-    exec(action: any, put: any, store: any) {
-      const guideName = action.cmd.substr(':guide'.length).trim()
-      if (!guideName) {
-        put(resetGuide())
-        put(open('guides'))
+    match: (cmd: string) => /^guide(\s|$)/.test(cmd),
+    exec(action: ExecuteSingleCommandAction, dispatch: Dispatch<Action>) {
```
OskarDamkjaer, line 505: This action is not necessarily `ExecuteSingleCommandAction`, it could also be a SYSTEM_COMMAND_QUEUED action or some other action (such as on line 257 where the action is created without an action creator). For now I'm not sure what the best way to handle this is. I think we can leave it like it is now or change the type to be {cmd: string} instead. Anyway, this problem will go away when commandInterpreterHelper is fully typed

OskarDamkjaer (changes requested):
This looks really good now! I left a few comments that you don't need to act on unless you want to, except for the comment on fetchRemoteGuideEpic, where running `:guide movies` makes it show up in the remote guide list

src/shared/services/commandInterpreterHelper.ts
FallingCeilingS: @OskarDamkjaer - I think I'll create a generic type/interface `CommandAction`. Then, any types can extend this one. And we can use this generic action in this file.
Will do in subsequent PR.

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -505,32 +501,16 @@ const availableCommands = [
...
-          }
-          put(setCurrentGuide(guide))
-          if (!guide.isError) {
-            put(addGuideIfExternal(guide))
-          }
-
-          put(open('guides'))
-        }
-      )
+      dispatch(fetchRemoteGuide(guideIdentifier))
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - agree. Will handle it in subsequent PR.

OskarDamkjaer (approved):
Looks good! Looking forward to seeing your follup PRs as well! ⭐

FallingCeilingS:
Thanks @jharris4 @OskarDamkjaer for reviewing! I'm going to merge it in!

PR #1595, opened 2021-11-12
Show current manuals instead of old if not connected to server

eijawerner:

### DIFF
--- a/src/browser/modules/DBMSInfo/DatabaseKernelInfo.tsx
+++ b/src/browser/modules/DBMSInfo/DatabaseKernelInfo.tsx
@@ -25,7 +25,8 @@ import {
   getEdition,
   getStoreSize,
   getClusterRole,
-  getDatabases
+  getDatabases,
+  Database
 } from 'shared/modules/dbMeta/dbMetaDuck'
 import {
   executeCommand,
@@ -48,6 +49,15 @@ import {
 } from './styled'
 import { getUsedDbName } from 'shared/modules/features/versionedFeatures'
 
+type DatabaseKernelInfo = {
+  role: any
+  version: string | null
+  edition: string | null
+  dbName: any
+  storeSize: any
+  onItemClick: any
+  databases: Database[]
+}
 export const DatabaseKernelInfo = ({
   role,
   version,
@@ -56,7 +66,7 @@ export const DatabaseKernelInfo = ({
   storeSize,
   onItemClick,
   databases
-}: any) => {
+}: DatabaseKernelInfo) => {
   return (
     <DrawerSection className="database-kernel-info">
       <DrawerSubHeader>DBMS</DrawerSubHeader>
--- a/src/browser/modules/Sidebar/Documents.tsx
+++ b/src/browser/modules/Sidebar/Documents.tsx
@@ -43,7 +43,7 @@ import {
 import { formatDocVersion } from './docsUtils'
 
 export const shouldLinkToNewRefs = (v: string): boolean => {
-  if (!semver.valid(v)) return false
+  if (!semver.valid(v)) return true
   return semver.gte(v, '3.5.0-alpha01')
 }
 
@@ -160,7 +160,7 @@ const Documents = (props: DocumentsProps) => {
           props.trackCannyFeatureRequest()
           window.open(CANNY_FEATURE_REQUEST_URL, '_blank')
         }}
-      ></StyledFeedbackButton>
+      />
       <StyledFullSizeDrawerBody>
         <DocumentItems header="Useful commands" items={useful} />
         <DocumentItems header="Documentation links" items={docs} />
@@ -171,7 +171,7 @@ const Documents = (props: DocumentsProps) => {
 }
 
 const mapStateToProps = (state: GlobalState) => {
-  const version = getVersion(state)
+  const version = getVersion(state) || 'current'
   return {
     version,
     urlVersion: formatDocVersion(version)
--- a/src/browser/modules/Stream/Queries/QueriesFrame.tsx
+++ b/src/browser/modules/Stream/Queries/QueriesFrame.tsx
@@ -36,7 +36,8 @@ import {
 } from 'shared/modules/cypher/cypherDuck'
 import {
   getConnectionState,
-  CONNECTED_STATE
+  CONNECTED_STATE,
+  ConnectionState
 } from 'shared/modules/connections/connectionsDuck'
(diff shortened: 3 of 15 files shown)


src/browser/modules/Sidebar/Documents.tsx
```diff
@@ -41,8 +41,8 @@ import {
   StyledHeaderContainer
 } from './styled'
 
-export const formatDocVersion = (v = ''): string => {
-  if (!semver.valid(v)) {
+export const formatDocVersion = (v: string | null = ''): string => {
```
FallingCeilingS, line 44: `v` is probably `undefined` rather than `null`? Please correct me if I'm wrong.

FallingCeilingS (changes requested):
The fix looks good to me! I only have two minor comments.

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -97,8 +97,8 @@ export function getMetaInContext(state: any, context: any) {
   }
 }
 
-export const getVersion = (state: any) =>
-  (state[NAME] || {}).server ? (state[NAME] || {}).server.version : 0
+export const getVersion = (state: any): string | null =>
```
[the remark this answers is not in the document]
eijawerner: Thanks, I'll fix!

src/browser/modules/Sidebar/Documents.tsx
eijawerner: It is actually null in the initialState in redux.

src/shared/modules/features/versionedFeatures.ts
```diff
@@ -31,70 +32,71 @@ export const FIRST_MULTI_DB_SUPPORT = NEO4J_4_0
...
     return getUseDb(state)
   }
   return getActiveDbName(state)
 }
 
-export const getDefaultBoltScheme = (serverVersion: any) => {
-  serverVersion = guessSemverVersion(serverVersion)
+export const getDefaultBoltScheme = (serverVersion: string | null) => {
+  const serverVersionGuessed = guessSemverVersion(serverVersion)
   const pre4 = 'bolt://'
```
FallingCeilingS, line 95: It's not related to the PR. But can you try to rename it to a more descriptive one?

FallingCeilingS (changes requested):
Changes look good to me! Nice to have tests updated too! I commented some area that are optional to change in this PR. Also I haven't tested the code locally, and will do it later.
Finding the reason why tests are failing is awesome!

src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -97,9 +97,9 @@ export function getMetaInContext(state: any, context: any) {
   }
 }
 
-export const getVersion = (state: any) =>
-  (state[NAME] || {}).server ? (state[NAME] || {}).server.version : 0
-export const getEdition = (state: any) => state[NAME].server.edition
+export const getVersion = (state: GlobalState): string | null =>
+  (state[NAME] || {}).server ? (state[NAME] || {}).server.version : null
```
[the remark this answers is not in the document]
eijawerner: I thought so too, but checked with Hane who added these kind of logics and he thought it can be so keeping it for now.

src/browser/modules/DBMSInfo/DatabaseKernelInfo.tsx
```diff
@@ -48,6 +49,15 @@ import {
...
 import { getUsedDbName } from 'shared/modules/features/versionedFeatures'
 
+type DatabaseKernelInfo = {
+  role: any
+  version: string | null
+  edition: string | null
+  dbName: any
+  storeSize: any
+  onItemClick: any
+  databases: Database[]
```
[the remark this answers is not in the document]
eijawerner: yes, but this PR never ends if I handle all type problems, already I think it got a lot bigger than planned :). So I decided to put a line here so the PR is not too big.

src/browser/modules/Stream/Queries/QueriesFrame.tsx
```diff
@@ -59,10 +60,27 @@ import FrameError from '../../Frame/FrameError'
...
+}
 
-export class QueriesFrame extends Component<any, QueriesFrameState> {
+type QueriesFrameProps = {
+  frame?: any
+  bus: any
+  availableProcedures: any
+  connectionState: number
+  neo4jVersion: string | null
+}
```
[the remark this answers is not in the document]
eijawerner: Same as above

src/browser/modules/Stream/SchemaFrame.tsx
```diff
@@ -37,9 +37,15 @@ import {
 } from 'browser-components/DataTables'
 import Directives from 'browser-components/Directives'
 import { NEO4J_BROWSER_USER_ACTION_QUERY } from 'services/bolt/txMetadata'
+import { GlobalState } from 'project-root/src/shared/globalState'
 
-const Indexes = ({ indexes, neo4jVersion }: any) => {
+type IndexesProps = {
+  indexes: any
```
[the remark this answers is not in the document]
eijawerner: same as above

src/shared/modules/features/versionedFeatures.ts
eijawerner: Which one do you mean, the pre4 variable?

FallingCeilingS: @eijawerner - yes!

OskarDamkjaer: As it's unrelated to the main PR, it seems to me this shouldn't hold the PR from being merged? Seeing how we want to get a release candidate out soon. Also do you have any suggestion for a better name?

FallingCeilingS: @OskarDamkjaer - agree. I'd say we can rename it as `prefix` from this context.

FallingCeilingS (approved):
Changes look good to me! Those minor comments shouldn't block the merge process and can always be revisited in the future. 💯

src/shared/modules/features/versionedFeatures.ts
eijawerner: Is it ok if I leave it as it is for this PR? I think it is ok even if it is not perfect (and the pattern is used in several places in the file), and also as you said it is not really part of this PR.

PR #1596, opened 2021-11-15
Update some test deps

eijawerner:


src/browser/modules/Main/tests/slow-connection-pending.test.ts
```diff
@@ -19,6 +19,7 @@
  *
  */
 
+import '@testing-library/react-hooks/dont-cleanup-after-each.js'
```
OskarDamkjaer, line 22: Probably worth getting rid of these when and if we need to update the tests for other reasons in the future

OskarDamkjaer (approved):
🚀

PR #1597, opened 2021-11-15
Fix change log badge/notification doesn't appear until sidebar is opened

FallingCeilingS:
The canny badge doesn't show when connected to the database:
Screenshot 2021-11-15 at 17 10 20

But when clicking the "Help & Resources" icon, it turns up:
Screenshot 2021-11-15 at 17 11 43

The expected behaviour should be that it appears once connected to DB.

This PR also involves some code cleanups.


src/browser/components/TabNavigation/Navigation.tsx
```diff
@@ -51,6 +50,7 @@ export interface NavItem {
   name: string
   title: string
   icon: (isOpen: boolean) => JSX.Element
+  // Need to clean up code and properly type content
```
OskarDamkjaer, line 53: while I agree it's something we'd need to do, it feel unrelated to this change

jharris4: Yeah, we have the ideal of keeping PRs small and scoped to only the task at hand, but we also have a desire to gradually add types as we do other work.

Maybe we want to say that it's ok to add types, but only if it's in a file that is already being modified by the PR?

e2e_tests/integration/0.index.spec.ts
```diff
@@ -45,13 +46,59 @@ describe('Neo4j Browser', () => {
...
+    cy.get('[data-testid="frame"]').contains('the Bacon Path')
   })
+
   it('can connect', () => {
     const password = Cypress.config('password')
     cy.connect('neo4j', password)
   })
 
+  it('can show Canny badge in "Document" icon and correspond side bar drawer', () => {
+    // The Canny badge should appear when initially connected to the database
```
OskarDamkjaer, line 58: what I worry about with these tests are that they'll fail if canny is unavailable/tests run without network. Also not sure if we can run them twice, since the changelog has already been read the second time? Not a super big problem per se, but it could be really annoying when debugging

src/browser/modules/App/App.tsx
```diff
@@ -145,6 +146,12 @@ export function App(props: any) {
     props.bus && props.bus.send(initAction.type, initAction)
   }, [props.bus])
 
+  useEffect(() => {
+    window.Canny && window.Canny('initChangelog', cannyOptions)
+    return () => window.Canny && window.Canny('closeChangelog')
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
OskarDamkjaer, line 152: Why is this comment needed?

src/browser/modules/Sidebar/Sidebar.tsx
```diff
@@ -79,6 +79,7 @@ const Sidebar = ({
 }: SidebarProps) => {
   const topNavItemsList: NavItem[] = [
     {
+      // Consider use constant variable to store those keys
```
OskarDamkjaer, line 82: What exactly do you mean here? Store the names and titles?

OskarDamkjaer: Eija has suggested we start using this lib: https://www.npmjs.com/package/i18n to store all our user facing texts as variables

src/browser/services/canny.ts
```diff
@@ -29,9 +29,14 @@ export const cannyOptions = {
 export const CANNY_FEATURE_REQUEST_URL =
   'https://neo4j-browser.canny.io/feature-requests'
 
+interface CannyOptions {
```
OskarDamkjaer, line 32: 🙏

e2e_tests/integration/0.index.spec.ts
OskarDamkjaer: In the general case I prefer end2end tests for this kind of thing, but because of the canny-issues I just mentioned I think these could work better as unit tests of App.tsx where we mock the 'https://canny.io/sdk.js' url and see that it gets called.

OskarDamkjaer (commented):
Good to see the various cleanup changes and adding new tests, left a comment or two. Looking forward to hear your thoughts on them

src/browser/components/TabNavigation/Navigation.tsx
OskarDamkjaer: I personally don't mind unrelated type touch-ups but I was mostly saying it's not _needed_  for this pr

OskarDamkjaer: I notice I was not very clear in my comment

FallingCeilingS: @jharris4 @OskarDamkjaer - it's unrelated. I just added a comment so it reminds us to clean it up at some point. Also I noticed these components are a bit over complicated as many layers are nested. So, IMO probably better to touch it when we want to refactor our components.

src/browser/modules/Sidebar/Sidebar.tsx
OskarDamkjaer: I think it'd be nice to use something like i18n to store all our user facing text as variables 🤔

src/browser/components/TabNavigation/Navigation.tsx
OskarDamkjaer: Generally we don't leave any todo/reminder comments in the code, we'd rather put a card in trello. While I prefer keeping the source clean from those kinds of comments, I'm open to discussing it w/ you and Eija if you guys think we should start doing so!

e2e_tests/integration/0.index.spec.ts
FallingCeilingS: @OskarDamkjaer - do you think it's better to remove those test cases? Or add some conditions that we only run it when the network and Canny is available? Probably the former option is more straightforward?

OskarDamkjaer: I expanded on this but forgot to send the comment! I generally like e2e tests for these kinds of changes, but I was thinking it'd be more practical to do a unit test in this case. I'm thinking if we write a unit test for App.tsx where the we mock url for the canny sdk and ensure that it gets pinged we'd have confidence that the bug you fixed doesn't come back

src/browser/modules/App/App.tsx
FallingCeilingS: @OskarDamkjaer - TSlint is complaining about the dependency list. Any better idea to get rid of it?
Screenshot 2021-11-17 at 10 27 00

src/browser/modules/Sidebar/Sidebar.tsx
FallingCeilingS: @OskarDamkjaer - I'm not sure if we need to hard-code the item name as the key in any cases (I haven't found any hard-code so far). But if we need in any cases, IMO it's better to use some variables like:

```ts
const DOCUMENTS = 'documents'
```

```ts
const bottomNavItemsList: NavItem[] = [
  {
    name: DOCUMENTS,
    // rest of the code
  }
]
```

But if there's not any cases to get access to those keys manually, we can just leave it as it is.

src/browser/modules/App/App.tsx
OskarDamkjaer: Like the warning says -> the window.Canny is an unneeded dependency since mutating it doesn't re-render the component (meaning it's not a prop or state and only those two can go into dependency arrays)

OskarDamkjaer: Use effect generally has a lot of pitfalls, I don't know if you've read it before, but I do recommend reading this blogpost about it if you're interested https://overreacted.io/a-complete-guide-to-useeffect/

OskarDamkjaer: So an empty dependency array is the right move for this component + making sure that window.Canny is defined before calling the function

OskarDamkjaer: Wait now that I think of it. What we should do is move this logic to the `CannyLoader` component in `canny.ts`. It makes a lot more sense that it'd live there _and_ we'd get around the issue of window.canny sometimes not having been loaded yet!

src/browser/modules/Sidebar/Sidebar.tsx
OskarDamkjaer: The guide navigation item already exists as a variable "GUIDE_DRAWER_ID" but it's not used in this file yet. I think it'd make sense to use that variable and add the others as well. Then we could also update the `NavItem` type to also use these types/variables.

OskarDamkjaer: It'd be nice to have done, but not really needed for this PR unless you really want to do it

FallingCeilingS: @OskarDamkjaer - I'd prefer to leave it in the future!

e2e_tests/integration/0.index.spec.ts
```diff
@@ -45,13 +46,67 @@ describe('Neo4j Browser', () => {
...
   })
+
   it('can connect', () => {
     const password = Cypress.config('password')
     cy.connect('neo4j', password)
   })
 
+  it('can show Canny badge in "Document" icon and correspond side bar drawer', () => {
+    cy.window().then(win => {
+      if (win.Canny && win.IsCannyLoaded && typeof win.Canny === 'function') {
```
OskarDamkjaer, line 59: Good to not fail any releases if the canny sdk were to not load properly due to network issues or simliar

OskarDamkjaer (approved):
LGTM!

FallingCeilingS:
Thanks @OskarDamkjaer for reviewing! I'm going to merge it in!

PR #1598, opened 2021-11-15
Remove circular dependency in manual link

OskarDamkjaer:
Move the formatDocVersion dep to a utils file to avoid a circular dependency


eijawerner (approved):
⭐

PR #1602, opened 2021-11-18
Catch ts errors when running tests

eijawerner:
removed RowExpandToggle.tsx since getting error and is not used.


OskarDamkjaer (approved):
Don't know how these type errors slipped us by in the past.. Good catch!

PR #1605, opened 2021-11-18
Support new alias feature from neo4j 4.4

OskarDamkjaer:
Adds support for the 4.4 concept of aliases. 
To create an alias run `create alias myAlias for database neo4j`. 

You should not be able to run `:use myAlias` and see your aliases when running `:dbs`. The "real" name should still be displaced in editor.

Preview @http://support_aliases.surge.sh

### DIFF
--- a/src/browser/modules/App/App.tsx
+++ b/src/browser/modules/App/App.tsx
@@ -34,8 +34,8 @@ import { utilizeBrowserSync } from 'shared/modules/features/featuresDuck'
 import { getOpenDrawer, open } from 'shared/modules/sidebar/sidebarDuck'
 import { getErrorMessage } from 'shared/modules/commands/commandsDuck'
 import {
-  shouldAllowOutgoingConnections,
-  getDatabases
+  findDatabaseByNameOrAlias,
+  shouldAllowOutgoingConnections
 } from 'shared/modules/dbMeta/dbMetaDuck'
 import {
   getActiveConnection,
@@ -154,7 +154,7 @@ export function App(props: any) {
     codeFontLigatures,
     connectionState,
     consentBannerShownCount,
-    databases,
+    databaseIsUnavailable,
     defaultConnectionData,
     drawer,
     errorMessage,
@@ -268,7 +268,7 @@ export function App(props: any) {
                       lastConnectionUpdate={lastConnectionUpdate}
                       errorMessage={errorMessage}
                       useDb={useDb}
-                      databases={databases}
+                      databaseIsUnavailable={databaseIsUnavailable}
                       showUdcConsentBanner={
                         telemetrySettings.source === 'BROWSER_SETTING' &&
                         consentBannerShownCount <= 5
@@ -291,6 +291,11 @@ export function App(props: any) {
 }
 
 const mapStateToProps = (state: GlobalState) => {
+  const useDb = getUseDb(state)
+  const databaseIsUnavailable =
+    useDb === null ||
+    findDatabaseByNameOrAlias(state, useDb)?.status !== 'online'
+
   const connectionData = getActiveConnectionData(state)
   return {
     experimentalFeatures: getExperimentalFeatures(state),
@@ -311,8 +316,8 @@ const mapStateToProps = (state: GlobalState) => {
     browserSyncAuthStatus: getUserAuthStatus(state),
     loadSync: utilizeBrowserSync(state),
     isWebEnv: inWebEnv(state),
-    useDb: getUseDb(state),
-    databases: getDatabases(state),
+    useDb,
+    databaseIsUnavailable,
     telemetrySettings: getTelemetrySettings(state),
     consentBannerShownCount: getConsentBannerShownCount(state)
   }
--- a/src/browser/modules/App/App.tsx
+++ b/src/browser/modules/App/App.tsx
@@ -263,7 +263,6 @@ export function App(props: any) {
                   </ErrorBoundary>
                   <StyledMainWrapper id={MAIN_WRAPPER_DOM_ID}>
                     <Main
-                      activeConnection={activeConnection}
                       connectionState={connectionState}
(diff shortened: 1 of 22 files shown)


src/browser/modules/App/App.tsx
```diff
@@ -154,7 +154,7 @@ export function App(props: any) {
     codeFontLigatures,
     connectionState,
     consentBannerShownCount,
-    databases,
+    databaseIsUnavailable,
```
FallingCeilingS, line 157: How about `isDatabaseAvailable`?

src/shared/services/commandInterpreterHelper.ts
```diff
@@ -219,7 +219,9 @@ const availableCommands = [
         const cleanDbName = unescapeCypherIdentifier(normalizedName)
 
         const dbMeta = getDatabases(store.getState()).find(
-          (db: any) => db.name.toLowerCase() === cleanDbName
+          (db: Database) =>
+            db.name.toLowerCase() === cleanDbName ||
+            db.aliases?.find(alias => alias.toLowerCase() === cleanDbName)
```
FallingCeilingS, line 224: How about reuse `findDatabaseByNameOrAlias` method from `dbMetaDuck`?

FallingCeilingS (changes requested):
Good job!
I have two minor comments in the diff. As we paired together last Friday, I can understand why you made those changes much quicker.
In addition, it would be great to include unit test and E2E test on this scenario if you've confirmed the UI behaviour with PM?

src/shared/services/commandInterpreterHelper.ts
FallingCeilingS: So, we can just check if `findDatabaseByNameOrAlias` returns `undefined`.

src/browser/modules/App/App.tsx
OskarDamkjaer: that's better!

src/shared/services/commandInterpreterHelper.ts
OskarDamkjaer: Good catch

OskarDamkjaer:
Thanks for the comments, made a few improvements now. Also pushed the last part of the feature, only have testing left to do. I'll re-request your review when ready 🙏

src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -505,7 +505,6 @@ export const startupConnectEpic = (action$: any, store: any) => {
       )
 
       if (
-        //TODO Considder the SSO implications of this
         !(discovered && discovered.hasForceURL) && // If we have force url, don't try old connection data
```
OskarDamkjaer, line 508: Old comment left in by mistake

src/shared/modules/discovery/discoveryHelpers.test.ts
```diff
@@ -105,8 +105,6 @@ describe('getAndMergeDiscoveryData', () => {
     expect(discoveryData).toBeTruthy()
     expect(discoveryData?.host).toEqual(boltHost)
 
-    // expect(discoveryData?.source).toEqual(DISCOVERY_ENDPOINT)
-    // Todo test source via logs
```
OskarDamkjaer, line 109: another old comment left in by mistake

tsconfig.json
```diff
@@ -17,6 +17,7 @@
     "noFallthroughCasesInSwitch": true,
     "jsx": "react",
     "outDir": "./build",
+    "noEmit": true,
```
OskarDamkjaer, line 20: Set tsc to not emit any javascript when it runs, since we don't use tsc for compiling but we get weird errors when running the tests if we have accidentally have generated some js code

OskarDamkjaer:
As soon as it's released!

src/browser/modules/DBMSInfo/DatabaseSelector.tsx
```diff
@@ -39,32 +40,38 @@ const EMPTY_OPTION = 'Select db to use'
...
   if (!selectedDb) {
-    databasesList = ([] as any[]).concat(
-      [{ name: EMPTY_OPTION, status: null }],
-      databases
-    )
+    databasesList.unshift({ name: EMPTY_OPTION })
   }
-  const uniqDatabases = uniqBy(databasesList, 'name')
+  // When connected to a cluster, we get duplicates for each member
+  const uniqDatabases = uniqBy(databases, 'name')
```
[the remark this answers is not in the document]
OskarDamkjaer: Fixed, thanks!

src/browser/modules/DBMSInfo/DatabaseSelector.tsx
```diff
@@ -39,32 +40,38 @@ const EMPTY_OPTION = 'Select db to use'
...
-  let databasesList = databases
+  const databasesList: (Partial<Database> & {
+    name: string
+  })[] = databases
   if (!selectedDb) {
-    databasesList = ([] as any[]).concat(
-      [{ name: EMPTY_OPTION, status: null }],
-      databases
-    )
+    databasesList.unshift({ name: EMPTY_OPTION })
```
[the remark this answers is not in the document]
OskarDamkjaer: The empty option is used as a failsafe for when we don't have a database selected, which should never happen with the current way start up is done. Our handling of what database is still somewhat complex, so I'd rather leave it in

eijawerner (approved):

src/browser/modules/Stream/styled.tsx
```diff
@@ -390,3 +390,7 @@ export const AuraPromoLink = styled.a`
   text-decoration: none;
   margin-right: 5px;
 `
+
+export const AliasText = styled.span`
+  font-size: 13px;
```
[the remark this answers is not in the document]
OskarDamkjaer: Done!

OskarDamkjaer:
Merging as soon as the tests come back green

PR #1609, opened 2021-11-19
Show current manuals instead of old if not connected to server

eijawerner:


eijawerner:
Only opened this PR for testing out something, closing it.

PR #1618, opened 2021-12-01
Refactor frame template

OskarDamkjaer:
Currently the component tree for our frames looked like this (example where we have a single :history frame).
`Stream` -> `:history-frame` -> `frameTemplate` (containing state about fullscreen) -> `frameEditor` (containing `monaco`) which meant that when a frame was re-run (and it's frame type changed) it's children (frame template and frame editor) would unmount => meaning `monaco` would unmount, reinitialize, we'd lose fullscreen state. This PR adds a component between the `Stream.tsx` and the `Frame`s that holds the fullscreen state and the `monaco` editor, to avoid this issue.


FallingCeilingS:
I think this branch should target to `cooldown` for now?

src/browser/modules/Stream/FrameContainer.tsx
```diff
@@ -0,0 +1,166 @@
...
+    isCollapsed,
+    setExportItems
+  }
+  const FrameComponent = getFrameComponent(props.frameData)
+
+  return (
+    <StyledFrame
+      className={isFullscreen ? 'is-fullscreen' : ''}
+      data-testid="frame"
+      fullscreen={isFullscreen}
```
FallingCeilingS, line 123:
```suggestion
      ifFullscreen={isFullscreen}
```

src/browser/modules/Stream/FrameContainer.tsx
```diff
@@ -0,0 +1,166 @@
...
+    >
+      <FrameTitlebar
+        frame={frame}
+        pinned={props.frameData.isPinned}
+        fullscreen={isFullscreen}
```
FallingCeilingS, line 128:
```suggestion
        isFullscreen={isFullscreen}
```

src/browser/modules/Stream/FrameContainer.tsx
```diff
@@ -0,0 +1,166 @@
...
+        fullscreenToggle={toggleFullscreen}
+        collapse={isCollapsed}
```
FallingCeilingS, line 130:
```suggestion
        isCollapse={isCollapsed}
```

src/browser/modules/Stream/Stream.tsx
```diff
@@ -21,71 +21,15 @@
 import { connect } from 'react-redux'
```
FallingCeilingS, line 21: Nice to see the restructuring eliminates functions that shouldn't be this component's responsibility.

src/browser/modules/Stream/FrameContainer.tsx
```diff
@@ -0,0 +1,166 @@
...
+    setFullscreen(full => !full)
+    setCollapsed(false)
+  }
+
+  return {
+    isCollapsed,
+    isFullscreen,
+    toggleCollapse,
+    toggleFullscreen
+  }
```
FallingCeilingS, line 165: This is a hook. I noticed there's a `src/browser/hooks` folder contains several hooks, and some hooks files inside of components files or folders. Better to consider the consistency of file structure and do some refactor.
This is not related to this PR. Just write it down so we don't forget it.

FallingCeilingS (changes requested):
Great job for refactoring. I have a few comments for those changes. I'll revisit it once failing tests are fixed.

src/browser/modules/Frame/FrameTemplate.tsx
```diff
@@ -18,25 +18,18 @@
...
   StyledFrameBody,
   StyledFrameContents,
   StyledFrameStatusbar,
   StyledFrameMainSection,
-  StyledFrameAside,
-  ContentContainer
+  StyledFrameAside
 } from './styled'
 
 type FrameTemplateProps = {
```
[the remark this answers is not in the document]
OskarDamkjaer: Nice catch!

src/browser/modules/Stream/FrameContainer.tsx
OskarDamkjaer: Nice! refactored all of these.

src/browser/modules/Frame/FrameTemplate.tsx
```diff
@@ -18,25 +18,18 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
```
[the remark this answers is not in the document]
OskarDamkjaer: Good idea! Makes it easier to understand the component I think

src/browser/modules/Frame/DefaultFrame.tsx
```diff
@@ -20,7 +20,11 @@
 import React from 'react'
 import FrameTemplate from './FrameTemplate'
 
-const DefaultFrame = ({ frame }: any): JSX.Element => {
+const DefaultFrame = ({
+  frame,
+  isCollapsed,
+  isFullscreen
+}: any): JSX.Element => {
```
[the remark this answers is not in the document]
OskarDamkjaer: I suspect this frame is actually not used. I've removed it and will see if tests still pass.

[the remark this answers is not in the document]
OskarDamkjaer: as well as doing some manual testing

e2e_tests/integration/style.spec.ts
```diff
@@ -50,7 +50,7 @@ describe(':style', () => {
     cy.executeCommand(':clear')
     cy.executeCommand(':style')
     // can't trigger hover with cypress so we can't know it's visible
-    cy.get('[data-testid="exportGrassButton"]').should('exist')
+    cy.get('[data-testid="exportGraSSButton"]').should('exist')
```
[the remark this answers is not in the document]
OskarDamkjaer: It's stylized like this for the export => grass is short for Graph Style Sheet

OskarDamkjaer:
Had to do a force push to fix a merge conflict with the snapshot testing, but I should have addressed all your comments now! :)

src/browser/modules/Stream/FrameContainer.tsx
FallingCeilingS: I guess you forgot to make changes here?

src/browser/modules/Frame/DefaultFrame.tsx
```diff
@@ -1,39 +0,0 @@
-/*
```
FallingCeilingS, line 1: 👍

src/browser/modules/Frame/FrameBodyTemplate.tsx
```diff
@@ -0,0 +1,103 @@
...
+  isFullscreen: boolean
+  isCollapsed: boolean
+}
+
+function FrameBodyTemplate({
+  contents,
+  onResize = () => {
+    /*noop*/
+  },
+  sidebar,
```
FallingCeilingS, line 49:
```suggestion
  sideBar,
```

src/browser/modules/Frame/FrameBodyTemplate.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  return (
+    <>
+      <StyledFrameBody
+        fullscreen={isFullscreen}
+        collapsed={isCollapsed}
+        removePadding={removePadding}
+        hasSlides={hasSlides}
+      >
+        {sidebar && sidebar()}
```
FallingCeilingS, line 79:
```suggestion
        {sideBar && sideBar()}
```

FallingCeilingS (changes requested):
LGTM! I've tested locally and it works well. The editor is no longer re-mounted when switching between frame bodies and among normal and fullscreen views. Great job! 💯 
However, it seems like the collapse button doesn't work anymore:
Screenshot 2021-12-14 at 10 50 44
Can you please investigate it?
Also, I have some minor comments. Since this is a large PR, I'm wondering if @eijawerner wants to look at it too?

src/browser/modules/Stream/FrameContainer.tsx
eijawerner: I disagree that we should move all hooks there, I think we should only move those that are used at multiple places to the general folder. Similar to how we only move reusable components to the root, and keep those only used locally close to where it is used.

eijawerner:
Good catch that you found that the collapse is not working @FallingCeilingS! I tested now and there is more to it, when switching to another tab after pressing the collapse it collapses, then there is no way to get back to see the tabs again...

https://user-images.githubusercontent.com/11065688/146147249-4265bb4f-77cf-4327-9c8a-e0d6539b6920.mov

FallingCeilingS:
> Good catch that you found that the collapse is not working @FallingCeilingS! I tested now and there is more to it, when switching to another tab after pressing the collapse it collapses, then there is no way to get back to see the tabs again...
> 
>  Screen.Recording.2021-12-15.at.08.57.43.mov

@eijawerner - nice catch! 
@OskarDamkjaer - it would be great to add relevant test cases after you fixed this problem.

src/browser/modules/Stream/FrameContainer.tsx
OskarDamkjaer: Ah missed it in my search and replace.. thanks!

OskarDamkjaer: Got a few extra unrelated ones now as well, to make the whole code base consistent

src/browser/modules/Frame/FrameBodyTemplate.tsx
OskarDamkjaer: I'd prefer to keep it as one word, like in the rest of the codebase

OskarDamkjaer:
I've found the cause of the error was an odd pattern used in the CypherFrame. I've refactored it to avoid duplicating it's fullscreen and collapsed state in the component, and worked around the need to pass the frame height as a prop

eijawerner (changes requested):
Nice cleanup and fix of bug. 👍 However when I tested this PR I discovered an issue in Safari with zoom in fullscreen not working anymore. Please have a look at that before I am ready to approve this PR!

eijawerner (approved):
Great work! 🏅

FallingCeilingS (approved):
I've tested locally and it looks good to me! Thanks for putting efforts on refactoring. 💯 
In the future, we can probably think about running our E2E tests against different browsers we support so we can identify problems much quicker: https://docs.cypress.io/guides/guides/launching-browsers#Disabled-Barriers

PR #1621, opened 2021-12-03
Color for * nodes sometimes not updated in property details

eijawerner:
In property details the color of * are sometimes not updated when changing color, while it updates correctly in the sidebar.

image

Steps to reproduce bug:
1. Do a style reset with command: ":style reset"
2. fetch not all but one or some nodes, for example: MATCH (n:Person_0) RETURN n LIMIT 25 (then after that first fetch of some, if fetching all after that it still not works, but if fetch all first after a style reset it is working)
3. then try to change color of * in node properties, the color is not updated in the property details view.

### DIFF
--- a/e2e_tests/integration/viz.spec.ts
+++ b/e2e_tests/integration/viz.spec.ts
@@ -44,4 +44,32 @@ describe('Viz rendering', () => {
     )
     cy.executeCommand('MATCH (a:TestLabel) DETACH DELETE a')
   })
+  it('can change default color of nodes', () => {
+    const selectorNodeLabelAll =
+      '[data-testid="property-details-overview-node-label-*"]'
+    const defaultColor = 'rgb(165, 171, 182)'
+    const orangeColor = 'rgb(247, 151, 103)'
+    cy.executeCommand(':clear')
+    cy.executeCommand(':style reset')
+    cy.executeCommand(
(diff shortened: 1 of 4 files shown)


FallingCeilingS (commented):
Changes look quite straightforward. Can you also add relevant E2E tests?

e2e_tests/integration/viz.spec.ts
```diff
@@ -44,4 +44,32 @@ describe('Viz rendering', () => {
     )
     cy.executeCommand('MATCH (a:TestLabel) DETACH DELETE a')
   })
+  it('can change default color of nodes', () => {
+    const selectorNodeLabelAll =
+      '[data-testid="property-details-overview-node-label-*"]'
```
FallingCeilingS, line 49: There are some `data-testid`s written with camel case, while some are kebab case, in this project. I think we need to unify all of them, or even using variables to set these ids, in the future.

FallingCeilingS: Probably we can clean them up in your subsequent PR.

FallingCeilingS (approved):
Changes look good to me! Great job! Looking forward to seeing your subsequent PR for more clean-ups!
Let's merge it in once TC checks are passed!

PR #1622, opened 2021-12-13
Don't print warnings when running tests

OskarDamkjaer:
We still have 4000 warnings left as part of our typescript migration, so it makes no sense to print them all every time we run the tests. When we get down to a manageable number, we'll turn them on again


FallingCeilingS (approved):
Good to see the warnings are gone for now and the `console.error` still here to remind us to fix problems.

eijawerner (approved):

PR #1623, opened 2021-12-13
Visualisation performance optimisation 1

FallingCeilingS:
This is one small part of visualisation performance optimisation.
We are having performance issues of iterative layout/rendering of nodes. When running profiling tool in Chrome, we found a lot of expensive calculations and probably can be optimised.
One of the expensive calculation is `fitOnFixedNumberOfLines` in `fitCaptionInCircle` in `formatNodeCaption` method. `fitOnFixedNumberOfLines` uses canvas to measure text length and greedy algorithm to fit as many words in one line. `fitCaptionInCircle` finds the best number of lines and baseline (position) of each line to draw text inside of circle. This procedure called [typesetting](https://en.wikipedia.org/wiki/Computers_and_Typesetting). 
Although the algorithm itself in `fitOnFixedNumberOfLines` has the time complexity of `O(N^2)` and cannot be optimised anymore, we can still use other methods to improve the calculation and user experience.

Screenshot 2021-12-13 at 13 41 00

1. The existing text measuring requires using canvas so that we know if we should add the word in the current line or break into a new line (this saves us time to re-implementing the [Knuth-Plass Line Breaking Algorithm](http://litherum.blogspot.com/2015/07/knuth-plass-line-breaking-algorithm.html)). But the canvas was re-created every time we wanted to measure the text. We can reuse it and get the same canvas context.
2. We called for getting stylesheets, like `font-size` and `icon-code` in the iteration body, which involves some style re-calculation (`GraphStyle.forNode`, `GraphStyle.calculateStyle` and `StyleElement.applyRules`). It is inefficient. We can move it outside of the loop.
3. The adding node icon feature, which is written 5 years ago, is never surfaced. So, we can remove the code to reduce unnecessary calculation (see the screenshot below).
image
4. We called `viz.update()` to re-draw elements. But it was called multiple times. For now, we just set conditions to control which part should be re-drawn rather than the whole svg. Will do more refactor in the future.
5. This PR involves some TypeScript clean up.

TODO in future optimisation:
1. Introduce time-slicing for splitting CPU intensive micro tasks (https://javascript.info/event-loop).
2. Optimise layout and physics calculation.
3. Introduce canvas to draw relationships (it seems like relationships layout is the bottleneck of the visualisation).
4. Code clean up.


src/browser/modules/D3Visualization/GraphEventHandler.ts
```diff
@@ -55,7 +55,7 @@ export class GraphEventHandler {
     }
     this.selectedItem = item
     item.selected = true
-    this.graphView.update()
+    this.graphView.update({ updateNodes: true, updateRelationships: true })
```
OskarDamkjaer, line 58: What do you think of having both these options as true by default, so you can just call .update() when you want a full update, and then optionally use the parameter when needed?

src/browser/modules/D3Visualization/lib/visualization/components/Node.ts
```diff
@@ -0,0 +1,86 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+type NodeProperties = { [key: string]: number | string | boolean }
```
OskarDamkjaer, line 21: I'm not sure, but could a property not also be an array? Do you know?

OskarDamkjaer: I see now that this file is based on a file I've created but a rename caused it to show up as new 😅

src/browser/modules/D3Visualization/mapper.ts
```diff
@@ -20,22 +20,22 @@
...
   const graph = new Graph()
   graph.addNodes(mapNodes(nodes))
   graph.addRelationships(mapRelationships(relationships, graph))
   ;(graph as any).display = { initialNodeDisplay: 300, nodeCount: 1 }
   return graph
 }
 
-export function mapNodes(nodes: any) {
+export function mapNodes(nodes: any): Node[] {
   return nodes.map(
```
OskarDamkjaer, line 39: I'm not sure I understand what this mapping function does. Since it's returning Node[] but is seems to also be taking Node[], when called by createGraph. Is there a type error hidden here somewhere?

src/browser/modules/D3Visualization/lib/visualization/utils/textMeasurement.ts
```diff
@@ -17,43 +17,43 @@
...
-export default function(text: any, fontFamily: any, fontSize: any) {
+export default function(
+  text: string,
+  fontFamily: string,
+  fontSize: number,
+  canvas2DContext: CanvasRenderingContext2D
+): number {
   const font = `normal normal normal ${fontSize}px/normal ${fontFamily}`
-  return cache()(text + font, () => measureUsingCanvas(text, font))
+  return cacheTextWidth()(`[${font}][${text}]`, () =>
```
OskarDamkjaer, line 56: We talked about taking the shortcut of using width of the widest available letter, did you decide not to pursuit that option or did you just not want to include it in this work?

src/browser/modules/D3Visualization/lib/visualization/components/visualization.ts
```diff
@@ -267,11 +269,16 @@ const vizFn = function(
     return latestStats
   }
 
-  viz.update = function() {
+  viz.update = function(options: {
+    updateNodes: boolean
```
OskarDamkjaer, line 273: This method seems to not use the new options?

src/browser/modules/D3Visualization/lib/visualization/utils/textMeasurement.ts
FallingCeilingS: @OskarDamkjaer - I need to check with @gregoryking if we have scenarios with different languages. If it only contains English, other latin languages and numbers, we can simply optimise this method. Otherwise, we need more investigation if it's doable for multi-languages, i.e. Arabic. Chinese, Japanese, Thai, etc. Potential problems are:
- A letter/character has different widths in different fonts (since we don't allow users to change font, it should be fine).
- Even one font can be rendered with slightly different widths in different browsers (that's why we need to create canvas to calculate the width under the runtime of certain browser).
- Symbols have different widths (including symbols' padding). So, we need to consider this case.

So, I think better **_not_** to include this work in this PR. But I'll keep it on my radar when I get back from @gregoryking 💪 .

src/browser/modules/D3Visualization/lib/visualization/components/visualization.ts
FallingCeilingS: `options` is passed into `geomotry.onGraphChange(graph, options)` in line 280.
https://github.com/FallingCeilingS/neo4j-browser/blob/visualisation-performance-optimisation-2/src/browser/modules/D3Visualization/lib/visualization/components/visualization.ts#L280

src/browser/modules/D3Visualization/mapper.ts
FallingCeilingS: @OskarDamkjaer - this is the correct type for `createGraph`. In this method, it calls `graph.addNodes(mapNodes(nodes))` assigns mapped `Node[]` data to it's class variable `_nodes`.

https://github.com/neo4j/neo4j-browser/blob/947678af576733051414d286cb16e3fa75ba062b/src/browser/modules/D3Visualization/mapper.ts#L30-L36

https://github.com/neo4j/neo4j-browser/blob/947678af576733051414d286cb16e3fa75ba062b/src/browser/modules/D3Visualization/lib/visualization/components/graph.ts#L74-L82

src/browser/modules/D3Visualization/lib/visualization/renders/init.ts
```diff
@@ -61,51 +62,25 @@ const nodeCaption = new Renderer({
...
     const text = selection
       .selectAll('text.caption')
-      .data((node: any) => node.caption)
+      .data((node: Node) => node.caption)
 
     text
       .enter()
       .append('text')
-      // .classed('caption', true)
+      // Classed element ensures duplicated data will be removed before adding
```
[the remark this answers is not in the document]
FallingCeilingS: @OskarDamkjaer - `selection.exit().remove()` [is typically used to remove “superfluous” elements corresponding to old data](https://github.com/d3/d3-selection/blob/v3.0.0/README.md#selection_exit). That is, we can use this method to remove old nodes' captions when updating nodes drawing, i.e. change nodes' radius. Otherwise, the existing node captions will still here:
Screenshot 2021-12-21 at 00 07 09
Screenshot 2021-12-21 at 00 07 20
As you can see the old caption is not removed when updating the node radius. So, those captions were overlapped together.

The mechanism is that we select all elements with the selector `text.caption`. So, `D3` is able to remove old elements and re-render new elements.
https://github.com/neo4j/neo4j-browser/blob/947678af576733051414d286cb16e3fa75ba062b/src/browser/modules/D3Visualization/lib/visualization/renders/init.ts#L60-L86

Previously, we have such code:
https://github.com/neo4j/neo4j-browser/blob/947678af576733051414d286cb16e3fa75ba062b/src/browser/modules/D3Visualization/lib/visualization/renders/init.ts#L92-L117
`nodeIcons` selects all `text` elements and remove old ones when updating, and that's why we don't need explicitly specify class of `nodeCaption`. However, `nodeIcon` is removed in this PR. So, we ran into the problem like the screenshot above. In this case, we need to uncomment the line 69 to set `.classed('caption', true)`.

src/browser/modules/D3Visualization/lib/visualization/utils/textMeasurement.ts
FallingCeilingS: @OskarDamkjaer - from this Trello card, we also need to consider emoji: https://trello.com/c/DcFUZlDU.

OskarDamkjaer: Sounds good, thanks!

src/browser/modules/D3Visualization/mapper.ts
OskarDamkjaer: `nodes` in `graph.addNodes(mapNodes(nodes))` is declared to be of type `Node[]` on line 30, so that we should be able to remove the  "any" from `addNodes`? When I do so I get type errors
CleanShot 2021-12-21 at 15 50 30@2x

OskarDamkjaer: So to me it seems there are multiple shapes a node can have, and the point of "mapNodes" is to map from one representation of nodes to another.

OskarDamkjaer: Looking into it, it seems that the mapping function is meant to take the type of node returned from `extractNodesAndRelationshipsFromRecordsForOldVis` and map it into our own Node class

OskarDamkjaer: Digging deeper, the nodes passed to the createGraph constructor come from the state in VisualizationView.tsx

OskarDamkjaer: So they aren't of type "Node", I'm thinking the best way forward is either to make sure everything we say has the `Node` type really is a Node of leave it as any

FallingCeilingS: @OskarDamkjaer - it should be like
```ts
// graph.ts
  addNodes(nodes: Node[]) { 
    // ...
 } 
```
But for `mapNodes()`, the parameter has different shape and it's mapped to `Node[]` as the returned value:
```ts
// mapper.ts
  mapNode(nodes: UnmappedNodeFromQueryResult[]): Node[] {
    // ...
  }
```

src/browser/modules/D3Visualization/lib/visualization/renders/init.ts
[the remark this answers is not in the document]
OskarDamkjaer: Ah interesting, thanks for explaining!

src/browser/modules/D3Visualization/lib/visualization/components/visualization.ts
OskarDamkjaer: ah, sorry It seems I've become blind

src/browser/modules/D3Visualization/lib/visualization/utils/pairwiseArcsRelationshipRouting.ts
```diff
@@ -21,20 +21,30 @@ import measureText from './textMeasurement'
...
+import GraphStyle from 'browser/modules/D3Visualization/graphStyle'
 
 export default class PairwiseArcsRelationshipRouting {
-  style: any
-  constructor(style: any) {
+  style: GraphStyle
+  canvas: HTMLCanvasElement
+  constructor(style: GraphStyle) {
     this.style = style
+    this.canvas = document.createElement('canvas')
```
OskarDamkjaer, line 31: Just a note: All the text measurement before seemed to reuse the same canvas, rather than create their own: `d3.select('canvas#textMeasurementCanvas)`.  I'm assuming this is a convenience tradeoff you've made that just uses a bit more memory but isn't a problem?

src/browser/modules/D3Visualization/lib/visualization/components/graphGeometry.ts
```diff
@@ -50,118 +60,138 @@ export default class NeoD3Geometry {
...
-    for (
-      let iLine = 0, end = lineCount - 1, asc = end >= 0;
-      asc ? iLine <= end : iLine >= end;
-      asc ? iLine++ : iLine--
-    ) {
-      const line = emptyLine(lineCount, iLine)
+    const wordMeasureWidthList: number[] = []
+    words.forEach((word: string) =>
+      wordMeasureWidthList.push(measure(`${word}`))
+    )
```
OskarDamkjaer, line 158: Doesn't really matter (don't need to change it if you don't feel like it), but is there any reason not write this as:

```typescript
const wordMeasureWidthList = words.map(measure)
```
or
```typescript
const wordMeasureWidthList = words.map(w => measure(w))
```

OskarDamkjaer (changes requested):
Just a type error left to fix, otherwise looks good! :D

src/browser/modules/D3Visualization/lib/visualization/utils/pairwiseArcsRelationshipRouting.ts
FallingCeilingS: @OskarDamkjaer - that's correct. It's not the bottleneck. Even I don't do the refactor, it should be fine.

src/browser/modules/D3Visualization/mapper.ts
FallingCeilingS: @OskarDamkjaer - thanks for pointing it out. I've distinguished two different `Node` types in the latest commit!

src/browser/modules/D3Visualization/lib/visualization/utils/pairwiseArcsRelationshipRouting.ts
FallingCeilingS: @OskarDamkjaer - `getContext` needs some time:
Screenshot 2021-12-21 at 22 52 53
So, it's slightly more efficient to pass the same `context` to the method to save time.

But, again, this part is not the bottleneck. Even we don't optimise it, it won't affect the performance too much.

OskarDamkjaer (approved):
Nice! Good to get these perf updates in 🚀

PR #1624, opened 2021-12-14
Improve performance of queries that return lots of nodes

nglgzz:
This change should improve the rendering time for query results that contain lots of nodes. I used the following query throughout my tests:

```cypher
WITH range(1,100000) as range 
UNWIND range as item 
CREATE (n:Stuff {iter: item}) 
RETURN n
```

and measured:

1. The time it takes from dispatching the `requests/SENT` to the `requests/UPDATED` actions in Redux. 
2. The time it takes to execute some snippets/functions that could be potential bottlenecks. 

### Before

1. **36.42s**

2. - **1ms** - commandInterpreterHelper.ts - exec - handleCypherCommand
   - **525ms** - boltWorker.ts - onmessage - cypherResponseMessage
   - **636ms** - setup-bolt-worker.ts - setupBoltWorker - addTypesAsField
   - **1ms** - boltWorker.ts - onmessage - cypherResponseMessage
   - **43790ms** - commandInterpreterHelp.ts - exec - updateQueryResult

### After

1. **03.60s**

2. - **4ms** - commandInterpreterHelper.ts - exec - handleCypherCommand
   - **517ms** - boltWorker.ts - onmessage - cypherResponseMessage
   - **295ms** - setup-bolt-worker.ts - setupBoltWorker - addTypesAsField
   - **0ms** - boltWorker.ts - onmessage - cypherResponseMessage
   - **944ms** - commandInterpreterHelp.ts - exec - updateQueryResult

### Notes

The time measurements can vary quite a lot between different runs, but even if they're not precise they can give an idea of the impact of each potential bottleneck.

Further improvements in this area could be achieved by optimizing the serialization of the results between the worker and main threads (eg. using shared memory), or the visualization/state management (more investigation would be needed for improving the visualization though).

### DIFF
--- a/src/browser/modules/D3Visualization/components/Explorer.tsx
+++ b/src/browser/modules/D3Visualization/components/Explorer.tsx
@@ -43,15 +43,15 @@ import { Action, Dispatch } from 'redux'
 const deduplicateNodes = (nodes: any) => {
   return nodes.reduce(
     (all: any, curr: any) => {
-      if (all.taken.indexOf(curr.id) > -1) {
+      if (all.taken[curr.id]) {
         return all
       } else {
         all.nodes.push(curr)
-        all.taken.push(curr.id)
+        all.taken[curr.id] = true
         return all
       }
     },
-    { nodes: [], taken: [] }
+    { nodes: [], taken: {} }
   ).nodes
 }
 


OskarDamkjaer:
Lovely to see such a nice performance improvement 😍 I was wondering if there was some neat way we could make sure we don't regress again on this function... I went looking for some good way to add a "benchmarking" unit test, but the problem is that it depends too much on outside factors (like how fast your laptop happens to be). 

For the future I think it'd be interesting if we leave some ways in the code so that manual testing of performance is easier something like the code checking if `node.ENV === "TEST"` and then running `performance.mark`. 🤔

OskarDamkjaer (approved):
Lovely! Let's merge on green!

FallingCeilingS (approved):
It's a simple change but it significantly optimised time complexity from `O(N^2)` to `O(N)`! Looking forward to seeing more optimisations!

PR #1628, opened 2021-12-20
Bump prettier and add automatic import sorting

OskarDamkjaer:


FallingCeilingS (approved):
Looks good to me! 💯 We can also bump TypeScript to ^4.5.x at some point so the prettier and eslint can leverage the latest features of TS.

OskarDamkjaer:
Will merge after investigating why testcases in #1629 fail

PR #1629, opened 2021-12-20
Format codebase using new prettier config

OskarDamkjaer:
Depends on #1628


FallingCeilingS (commented):
It's a good idea to sort imports. However, seems like this plugin only allows us sort imports in alphabetical way. IMO ideally we should be able to categorise imports with external, internal, etc. And in each category, we can sort it alphabetically. It would be beneficial for developers to find dependencies in the file quickly and makes code more organised.

So, how about trying [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import)? It's more powerful for sorting by our defined orders: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

Thoughts?

OskarDamkjaer:
@FallingCeilingS I found some extra setting to allow us to split between external and internal packages :D

FallingCeilingS:
@OskarDamkjaer - sorry for causing so many conflicts by #1623 😢. I hope this is not a super complicated to re-run the fix!

PR #1631, opened 2021-12-22
Visualisation performance optimisation 2: refactor svg append

FallingCeilingS:
This PR is the preparation for `D3.js` version upgrade (@eijawerner is looking into this) and future optimisation.

We have unnecessary prototype extending for `d3.selection().enter().appendSVG()` for loading text-based SVG icons to node menu context items. We can definitely use native `d3` methods to complete this job. After removing this one, we are able to avoid errors thrown when upgrading to `d3` version 4. (first commit)

In the meanwhile, this PR includes some code cleanups and type correction. (second commit)


src/browser/modules/D3Visualization/lib/visualization/renders/menu.ts
```diff
@@ -102,42 +100,51 @@ const createMenuItem = function(
...
+    new DOMParser().parseFromString(rawSvgIcon, 'application/xml')
+      .documentElement.firstChild as HTMLElement,
+    true
+  )
   const icon = iconPath
     .enter()
-    .appendSVG(rawSvgIcon)
+    .append('g')
+    .html(svgIcon.innerHTML)
+    .classed('icon', true)
```
FallingCeilingS, line 119: This is the core change for implementing `appendSVG` method that was removed in this PR.
The rest of the diff is code cleanups and refactors.

src/browser/modules/D3Visualization/lib/visualization/renders/icons.ts
```diff
@@ -18,7 +18,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const icons: any = {
+const icons: Record<string, string> = {
```
eijawerner, line 21: Set type ``{'Expand / Collapse': string, Unlock: string ....}`` instead

src/browser/modules/D3Visualization/lib/visualization/renders/menu.ts
```diff
@@ -157,60 +164,58 @@ const createMenuItem = function(
...
 
 const donutExpandNode = new Renderer({
-  onGraphChange(selection: any, viz: any) {
+  onGraphChange(selection: d3.Selection<NodeVisualisationModel>, viz: any) {
     return createMenuItem(
       selection,
       viz,
       'nodeDblClicked',
       2,
-      'expand_node',
```
OskarDamkjaer, line 183: Why does this class name need to change? It's referred to in our style sheets, it'll need to be changed there as well

OskarDamkjaer: We'll merge and then send a followup PR

OskarDamkjaer (approved):
Minor thing to fix, we'll approve and do it in a followup PR

eijawerner (commented):
Nice cleanup and great to get rid of the svg append stuff! ⭐  I just have comment about a type.

PR #1635, opened 2021-12-29
Fix hidden type issues causing crashes

OskarDamkjaer:
One of the types that made up the `VizItem` type was not defined, which in turn made the `VizItem` type `any` which is fixed in this pr. 

There was also a bug where some relationships with only one node were not removed, which I've also fixed.
I've also renamed the Node.ts file to match the class it's containing "VizNode"


nglgzz (approved):
Great fix! :+1:

PR #1637, opened 2022-01-10
Show hint about :auto on CALL {...} IN TRANSACTIONS

OskarDamkjaer:
We do a few regex checks on error messages to see if the error was caused by a missing `:auto` prefix, if so we add a hint to the error message.

This PR adds a check for if the error is the new `CALL IN {} TRANSACTIONS` which needs :auto and updates the check for `PERIODIC COMMIT` (that has the new error message`"A query with 'PERIODIC COMMIT' can only be executed in an implicit transaction but tried to execute in an explicit transaction."`).

I also add a few tests to verify the behaviour, updated the `:help auto` to use the new syntax and added a helper method to cypress for faster tests.

CleanShot 2022-01-10 at 14 41 32

URL for testing: update_warning.surge.sh
Do note that a 4.4 db is needed.

### DIFF
--- a/e2e_tests/integration/auto-prefix.spec.ts
+++ b/e2e_tests/integration/auto-prefix.spec.ts
@@ -36,13 +36,14 @@ describe(':auto prefix in browser', () => {
 
   it('adding :auto enables running periodic commit', () => {
     cy.executeCommand(':auto USING PERIODIC COMMIT RETURN "Laverre";')
-    cy.getFrames().contains('"Laverre"')
+    cy.getFrames().contains('ERROR')
+    cy.getFrames().contains('expected "LOAD"')
   })
 
   if (Cypress.config('serverVersion') >= 4.4) {
     it('shows help link when running CALL IN TRANSACTIONS without :auto', () => {
       cy.executeCommand(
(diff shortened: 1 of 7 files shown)


e2e_tests/integration/auto-prefix.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  it('shows help link when running period commit without :auto', () => {
+    cy.executeCommand('USING PERIODIC COMMIT RETURN "Verdanturf"')
+    cy.getFrames().contains('ERROR')
+    cy.getFrames().contains(':auto')
+  })
+
+  it('adding :auto enables running periodic commit', () => {
+    cy.executeCommand(':auto USING PERIODIC COMMIT RETURN "Laverre";')
+    cy.getFrames().contains('ERROR')
```
nglgzz, line 39: Could you add a comment explaining why we're looking for an error here?

OskarDamkjaer: sure, it definitely looks a bit odd

nglgzz (approved):
Looks good to me :+1:

PR #1639, opened 2022-01-13
Visualization: tweak interaction and positioning of nodes

nglgzz:
Add some tweaks to get the updated visualization to behave closer to the older version:

- Lock nodes on hover and unlock them on mouse leave.  
- Increase the strength of the force repelling nodes.
- Accelerate simulation by running multiple ticks between each render (the difference here can mainly be seen when dragging nodes, the neighboring nodes follow more quickly than before).

### DIFF
--- a/src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
+++ b/src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
@@ -71,8 +71,11 @@ const layout: AvailableLayouts = {
         )
 
       const simulation = forceSimulation<VizNode, Relationship>()
-        .force('charge', forceManyBody().strength(-300))
-        .on('tick', () => render())
+        .force('charge', forceManyBody().strength(-400))
+        .on('tick', () => {
+          simulation.tick(10)
+          render()
+        })
         .stop()
 
       const update = function (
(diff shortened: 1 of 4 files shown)


src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
```diff
@@ -71,8 +71,11 @@ const layout: AvailableLayouts = {
         )
 
       const simulation = forceSimulation<VizNode, Relationship>()
-        .force('charge', forceManyBody().strength(-300))
-        .on('tick', () => render())
+        .force('charge', forceManyBody().strength(-400))
+        .on('tick', () => {
+          simulation.tick(10)
```
OskarDamkjaer, line 76: Did you consider doing a variable amount of ticks here? I wonder if ticking as many times as possible within 1/60th of a second (or 1/100th of a second if we want to leave some time for the rest of the code to run) would make it even better?

OskarDamkjaer (approved):
Great work fiddling with the last few details! It's like they say when 90% is done, only 90% remain :D I left a suggestion, feel free to ignore or test in a future PR

src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
OskarDamkjaer: Nice! Good to know

PR #1641, opened 2022-01-17
Prevent simulation running when entering/leaving frame fullscreen

eijawerner:
It looks wierd that when entering and leaving frame fullscreen the nodes jumps to new positions (because the simulation is run and updated), prevent that by only calling resize.


nglgzz (approved):
:+1:

PR #1642, opened 2022-01-17
Cleanup of svg dimensions

eijawerner:
now both width and height are clearly defined to same as available space, and (0, 0) is in the center of the svg.

### DIFF
--- a/src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
+++ b/src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
@@ -35,11 +35,7 @@ import Relationship from './Relationship'
 import VizNode from './VizNode'
 
 type ForceLayout = {
-  update: (
-    graph: Graph,
-    size: [number, number],
-    animate?: boolean
-  ) => Simulation<VizNode, Relationship>
+  update: (graph: Graph, animate?: boolean) => Simulation<VizNode, Relationship>
   simulation: Simulation<VizNode, Relationship>
 }
 export type Layout = { init: (render: () => void) => ForceLayout }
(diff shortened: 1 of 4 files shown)


eijawerner:
Before (0, 0) point (green) was slightly off the centre (red), now centre point is in (0, 0). Also in the old the width of the viewbox was set to a smaller value than it actually got, resulting in weird behaviors, probably that caused the issues with going to frame fullscreen in Safari that now works. 🎉 
Screenshot 2022-01-17 at 08 34 52

src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
```diff
@@ -35,11 +35,7 @@ import Relationship from './Relationship'
 import VizNode from './VizNode'
 
 type ForceLayout = {
-  update: (
-    graph: Graph,
-    size: [number, number],
-    animate?: boolean
-  ) => Simulation<VizNode, Relationship>
+  update: (graph: Graph, animate?: boolean) => Simulation<VizNode, Relationship>
```
FallingCeilingS, line 38: `animate?` is not used anywhere. Can you also remove it?

FallingCeilingS (approved):
LGTM if tested locally! 💯

src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
eijawerner: Nice catch! It is actually used, since when I tried to remove it I see that it complains in Vizualization.ts row 324 where it is used. I guess it is just mapped to a different name (precompute), though it feels strange that typescript doesn't complain so probably need to specify typescript better here?? However since this is out of scope I'll leave as it is for this PR.

PR #1644, opened 2022-01-19
Fix issue with prettier and tsx files

eijawerner:
Changing back to default for importOrderParserPlugins since not working on tsx files.
the default for importOrderParserPlugins is [typescript, jsx]. Without the jsx the tsx files are not working,
however for queryPlan.ts it is not working for some strange reason to include jsx so adding an override only for .ts-files without the jsx.


nglgzz (approved):

OskarDamkjaer:
Nice!

PR #1645, opened 2022-01-19
Reduce icon size to avoid scroll on single line values in table

eijawerner:
- Fixed so single line object values will not have an unnecessary scroll by shrinking to copy icon
- Also, reduced the spacing around object values in tables
Before:
before
After:
Screenshot 2022-01-20 at 10 05 26


OskarDamkjaer (approved):
Nice simple, singles 👍

PR #1646, opened 2022-01-20
Visualization: tweak duration of the simulation

nglgzz:
- Remove the timeout that stops the simulation and adjust alpha values to make the simulation cool down on its own.
- Clicking on nodes no longer restarts the simulation.
- Reduce the number of precomputed ticks.
- Fixed a bug when dragging nodes, sometimes they would get unlocked when moving the cursor outside of the node.
- Remove re-render when clicking on the visualization background when no items are selected.

Preview available here: http://browser-visualization-tweaks.surge.sh

### DIFF
--- a/src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
+++ b/src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
@@ -30,7 +30,7 @@ import Graph from './Graph'
 import GraphGeometry from './GraphGeometry'
 import Relationship from './Relationship'
 import VizNode from './VizNode'
-import { Layout, clearSimulationTimeout, setSimulationTimeout } from './layout'
+import { Layout } from './layout'
 import GraphStyle from 'browser/modules/D3Visualization/graphStyle'
 
 export type MeasureSizeFn = () => { width: number; height: number }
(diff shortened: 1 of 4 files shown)


src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -278,18 +278,40 @@ const vizFn = function (
     }
 
     function dragHandler(simulation: Simulation<VizNode, Relationship>) {
+      let initialPos: [number, number]
+      let restartedSim = false
+      const tolerance = 25
+
+      const dist = (a: [number, number], b: [number, number]) =>
+        Math.pow(a[0] - b[0], 2) + Math.pow(a[1] - b[1], 2)
```
OskarDamkjaer, line 286: If we're going for euclidean distance we're missing `Math.sqrt()` here right? (this still kind of works though)

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -278,18 +278,40 @@ const vizFn = function (
...
-        event.subject.fy = event.y
+      function dragged(
+        event: D3DragEvent<SVGGElement, VizNode, any>,
+        node: VizNode
+      ) {
+        // This is to prevent clicks/double clicks from restarting the simulation
+        if (dist(initialPos, [event.x, event.y]) > tolerance && !restartedSim) {
+          // Set alphaTarget to a value higher than alphaMin so the simulation
+          // isn't stopped while nodes are being dragged.
+          simulation.alphaTarget(0.09).alpha(1).restart()
```
OskarDamkjaer, line 301: If you double click on a node or do a super short drag, the wiggle of surrounding nodes is still a bit jarring imo. I not necessarily a problem, but maybe setting the alpha to a lower value when dragging would make it better (?) (bit out of my depths on the simulation/target/alpha stuff tbh)  

https://user-images.githubusercontent.com/10564538/150495328-01f9e362-87d0-427c-a955-e8792fbbfa75.mp4

OskarDamkjaer: It'd be nice if "unrelated" nodes moved less when dragging as well, perhaps solved by the same change

src/browser/modules/D3Visualization/lib/visualization/components/layout.ts
```diff
@@ -40,20 +40,7 @@ type ForceLayout = {
...
-}
-export const clearSimulationTimeout = () => {
-  if (simulationTimeout) {
-    clearTimeout(simulationTimeout)
-    simulationTimeout = null
-  }
-}
-
-const SIMULATION_STARTING_TICKS = 800
+const SIMULATION_STARTING_TICKS = 300
```
OskarDamkjaer, line 43: Because the frames still take a little while to show up, it's easy to run into this issue:

https://user-images.githubusercontent.com/10564538/150496195-bad6113d-8cdb-401c-b501-a4c55fa83c61.mp4

I don't think changing the starting ticks more would solve the problem and I don't think it's something you'd need to look into for the d3 changes though.

OskarDamkjaer (approved):
Really nice we got rid of the timeout and got a "natural" way to end the simulation. It feels much better to play around with now imo, nice work! 🎉

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
nglgzz: Yeah, that's right. I took this from the code we had before and I removed the `Math.sqrt()`. I don't remember if the difference was noticeable, but the reasoning here was to remove unnecessary computation, since this function is run loads of times while dragging. I squared the tolerance too, so it is essentially the same.

OskarDamkjaer: Ah neat. Perhaps that's what the previous implementation did as well. I "fixed" it and added the sqrt. perhaps we should rename it to distanceIsh or add a comment so others don't "fix" it again

PR #1650, opened 2022-01-24
Display clustering data in :sysinfo frame again

OskarDamkjaer:
The list of causal clustering members in :sysinfo was mistakenly removed for 4.X in this [PR](https://github.com/neo4j/neo4j-browser/pull/997/files#diff-8510e7fbadeaa77c3c8cb21ca95c15c86c461eeae12a68b7f117e4ed3f1255e3L124) and I removed mistakenly removed it for 3.X in this pr #1472. 

I've split the SysInfoFrame into two components, one new (refactored and respects new settings) and one old (calls old metric endpoints and only works for 3.5). It led to a bit of duplication but this way we can refactor and improve the new SysInfoFrame without worrying that we break any of the old APIs. Some of the old code is a bit arcane, I based it on how things were in 4.3.1 when this was working for the customer. https://github.com/neo4j/neo4j-browser/tree/1ba490209c4aa6c392b6e857e3928a160b0840d5

I've made sure the High Availability clustering stats, as well as cluster member stats for 3.5 clusters now work properly again.  Screenshot of high availablity stats: CleanShot 2022-01-25 at 17 28 01
Video of all five clustering members behaving correctly.

https://user-images.githubusercontent.com/10564538/151018188-6d840a07-19d7-4c60-ae23-866ea7110a90.mp4


OskarDamkjaer:
A note: in neo4j 4.3 a change was made to that results in the causal clustering table to show up even when running single instance. Working with greg on how to make it less confusing

src/browser/modules/Stream/SysInfoFrame/helpers.tsx
```diff
@@ -270,3 +272,43 @@ export const responseHandler = (setState: (newState: any) => void) =>
...
+      return
+    }
+    const mappedResult = mapSysInfoRecords(res.result.records)
+    const mappedTableComponents = mappedResult.map((ccRecord: any) => {
+      const httpUrlForMember = ccRecord.addresses.filter((address: any) => {
+        return (
+          !address.includes(window.location.href) &&
+          (window.location.protocol.startsWith('file:')
+            ? address.startsWith('http://')
+            : address.startsWith(window.location.protocol))
```
nglgzz, line 289: No need to change this since it was there before, but it's quite hard to understand what we're trying to do here.

nglgzz (approved):
Great change to have two versions of this component instead of mixing the legacy and current code in one place!

src/browser/modules/Stream/SysInfoFrame/helpers.tsx
OskarDamkjaer: yeah for sure. I didn't want to change it because it's been working for a long time before I removed it

PR #1651, opened 2022-01-24
Fixed so possible to set params using multiline

eijawerner:
Now it is possible to use :params command with multiple lines Screenshot 2022-01-24 at 15 05 34

### DIFF
--- a/e2e_tests/integration/params.spec.ts
+++ b/e2e_tests/integration/params.spec.ts
@@ -21,10 +21,8 @@
 /* global Cypress, cy, before */
 
 describe(':param in Browser', () => {
-  before(function() {
-    cy.visit(Cypress.config('url'))
-      .title()
-      .should('include', 'Neo4j Browser')
+  before(function () {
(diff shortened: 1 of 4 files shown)


OskarDamkjaer (approved):
Nice that you found such a clean fix! 🙏

e2e_tests/integration/params.spec.ts
```diff
@@ -121,4 +119,17 @@ function runTests() {
     cy.resultContains('point({srid:4326, x:12.9082, y:57.7346})')
   }
   // })
+
+  // :params
+  // it('can set :params with multiple lines syntax', () => {
```
OskarDamkjaer, line 124: Is this meant to be commented out ?

eijawerner: yes, for now, this is because the file was rewritten when wanted to run all test cases with and without web worker. Probably the file can be rewritten so we still can keep the 'it' so we can get better test output. But I will not do that work in this PR.

OskarDamkjaer (approved):
Good to get a few more tests and typescript fixes along for the ride!

PR #1652, opened 2022-01-24
Refactor visualization code

nglgzz:
I realize these are a lot of changes, but these files are very interdependent, so it was easiest to modify them all at once. Below is a breakdown of the changes.

**Force layout:**
  - Convert the layout function into a class, since it was being used as a class anyway.
  - Extract all constants to a separate file to get a better overview of all the parameters we're using.
  - Split the update of nodes and relationships, and also split the functions for restarting or precomputing the simulation.

**Visualization:**
  - Similar to the force layout, this file now exports a class instead of a function. 
  - Extract the mouse event handlers to a separate file. 
  - Update the graph event handlers to update nodes and relationships, and to restart the force simulation only when needed. One noticeable change here is that now the simulation isn't restarted when deselecting a node.
  
**Graph component:** ~~the component now uses React hooks instead of a class.~~ cleanup of the component code to make it more readable. 
  
Preview: http://browser-visualization-refactor.surge.sh


src/browser/modules/D3Visualization/lib/visualization/components/Renderer.ts
```diff
@@ -19,12 +19,12 @@
  */
```
FallingCeilingS, line 19: This file should go under `render` folder?

src/browser/modules/D3Visualization/lib/visualization/utils/arrays.ts
```diff
@@ -27,7 +27,7 @@ export default function cloneArray<T>(original: Array<T>): Array<T> {
   return clone
```
FallingCeilingS, line 27: I'm not sure if the `clone` method makes any sense> It's not even a deep copy of each element. And I checked the usage, it never manipulates the indexes. Can you get rid of it?

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -17,216 +17,113 @@
...
+        const isZoomClick = this.isZoomClick
+        this.draw = true
+        this.isZoomClick = false
+
+        return this.container
+          .transition()
+          .duration(isZoomClick ? 400 : 20)
+          .call(sel => (isZoomClick ? sel.ease(easeCubic) : sel))
+          .attr('transform', String(e.transform))
+      })
```
FallingCeilingS, line 105: Can this be extracted to an event handler file? It's not directly related to visualisation itself.

FallingCeilingS (changes requested):
I had some initial reviews. It mostly looks good to me! Much more clear by the reorganising. And it also benefits my work - I am able to cherry-pick things I want more easily. Thank you!

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -17,216 +17,113 @@
...
-    }),
-    zoomOutClick: () => ({
-      zoomInLimit: false,
-      zoomOutLimit: false
-    }),
-    boundingBox: noOp,
-    resize: noOp,
-    update: noOp
-  }
+export class Visualization<T extends Element = SVGElement> {
```
OskarDamkjaer, line 44: nice to match the name of the file

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -17,216 +17,113 @@
...
-    // Make the rect cover the whole surface, center of the svg viewbox is in (0,0)
-    .attr('x', () => -Math.floor(measureSize().width / 2))
-    .attr('y', () => -Math.floor(measureSize().height / 2))
-    .attr('width', '100%')
-    .attr('height', '100%')
-    .attr('transform', 'scale(1)')
-
-  const container = baseGroup.append('g')
-  const geometry = new GraphGeometry(style)
+  public forceSim: ForceSimulation
```
OskarDamkjaer, line 52: As far as I understand it class properties are public by default, so I'm not sure if it adds anything to explicitly annotate when we don't do it in the rest of the code base. But it's a matter of taste, so feel free to keep it

src/browser/modules/D3Visualization/lib/visualization/components/mouseEventHandlers.ts
```diff
@@ -0,0 +1,146 @@
...
+  selection: Selection<SVGGElement, unknown, BaseType, unknown>,
+  baseGroup: Selection<SVGGElement, unknown, BaseType, unknown>,
+  zoomBehavior: ZoomBehavior<SVGGElement, unknown>
+) => {
+  const handleZoomOnShiftScroll = (e: WheelEvent) => {
+    if (e.shiftKey) {
+      e.preventDefault()
+
+      const delta =
+        -e.deltaY * (e.deltaMode === 1 ? 0.05 : e.deltaMode ? 1 : 0.002)
```
OskarDamkjaer, line 24: here's some more magic numbers that we haven't moved to constants. I don't fully understand what the numbers mean, so I'm not sure what I would name the constants, so perhaps they should stay as inline numbers for now 🤷

src/browser/modules/D3Visualization/lib/visualization/components/mouseEventHandlers.ts
```diff
@@ -0,0 +1,146 @@
...
+    .on('wheel.zoom', handleZoomOnShiftScroll)
+    .on('mousewheel.zoom', handleZoomOnShiftScroll)
+}
+
+export const nodeEventHandlers = (
+  selection: Selection<SVGGElement, VizNode, BaseType, unknown>,
+  trigger: (event: string, node: VizNode) => void,
+  simulation: Simulation<VizNode, Relationship>
+) => {
+  let initialDragPosition: [number, number]
```
OskarDamkjaer, line 42: Does it matter that these don't have an initial/default values? It seems they'd always be defined by `dragStart` before they're used in `dragged`. But I'm not sure how the handlers work

src/browser/modules/D3Visualization/lib/visualization/components/mouseEventHandlers.ts
```diff
@@ -0,0 +1,146 @@
...
+    initialDragPosition = [event.x, event.y]
+    restartedSimulation = false
+  }
+
+  const dragged = (
+    event: D3DragEvent<SVGGElement, VizNode, any>,
+    node: VizNode
+  ) => {
+    // Math.sqrt was removed to avoid unnecessary computation, since this
+    // function is called very often when dragging.
```
OskarDamkjaer, line 84: nice to get a comment here 👍

src/browser/modules/D3Visualization/lib/visualization/components/mouseEventHandlers.ts
```diff
@@ -0,0 +1,146 @@
...
+    .on('click', onNodeClick)
+    .on('dblclick', onNodeDblClick)
+}
+
+export const relationshipEventHandlers = (
+  selection: Selection<SVGGElement, Relationship, BaseType, unknown>,
+  trigger: (event: string, rel: Relationship) => void
+) => {
+  const onRelationshipClick = (event: Event, rel: Relationship) => {
+    event.stopPropagation()
```
OskarDamkjaer, line 130: just curious, why do we do stopPropagation on relationship clicks, but not on node clicks?

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -235,129 +132,115 @@ const vizFn = function (
...
+      nodeGroups.call(renderer.onGraphChange, this)
+    )
 
-    const layers = container
-      .selectAll('g.layer')
-      .data(['relationships', 'nodes'])
-      .join('g')
-      .attr('class', d => `layer ${d}`)
+    for (const renderer of nodeMenuRenderer) {
+      nodeGroups.call(renderer.onGraphChange, this)
```
OskarDamkjaer, line 161: At first I thought you were doing the same as on line 157 again by mistake, but i see now that it's iterating over different renders. Could change this second loop to also be forEach to make is easier to see that it's not duplicated by mistake?

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -235,129 +132,115 @@ const vizFn = function (
...
-    if (updateViz) {
-      force.update(graph, options.precompute)
+    if (options.updateRelationships) {
+      this.updateRelationships()
+    }
 
-      viz.resize()
-      viz.trigger('updated')
+    if (options.rerender) {
+      this.forceSim.restart()
```
OskarDamkjaer, line 227: What's the reason you call the option rerender rather than restartSimulation or similar?

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -235,129 +132,115 @@ const vizFn = function (
...
-          restartedSim = true
-        }
+    this.forceSim.updateRelationships(this.graph)
+  }
 
-        node.hoverFixed = false
-        node.fx = event.x
-        node.fy = event.y
-      }
+  private zoomClick(isZoomingIn: any) {
```
OskarDamkjaer, line 191: Would you mind annotating this as a boolean rather than any?

OskarDamkjaer: Ideally it should be 
`private zoomClick(direction: "IN" | "OUT")`
imo, but it's fine to leave it as boolean as well

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -235,129 +132,115 @@ const vizFn = function (
...
-        }
-      }
+    this.zoomBehavior.scaleBy(this.baseGroup, isZoomingIn ? 1.3 : 0.7)
 
-      return d3Drag<SVGGElement, VizNode>()
-        .on('start', dragstarted)
-        .on('drag', dragged)
-        .on('end', dragended)
-    }
+    return limitsReached
```
OskarDamkjaer, line 198: Should we remove this return value if we don't know if we hit the limits anymore?

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -235,129 +132,115 @@ const vizFn = function (
...
       .join('g')
-      .attr('class', 'node')
-      // @ts-ignore
-      .call(dragHandler(force.simulation))
-      .on('mouseover', onNodeMouseOver)
-      .on('mouseout', onNodeMouseOut)
-      .on('click', onNodeClick)
-      .on('dblclick', onNodeDblClick)
-      .classed('selected', node => node.selected)
+      .attr('class', d => `layer ${d}`)
```
OskarDamkjaer, line 206: Not understanding d3 fully makes it a bit harder to review this, since I don't really "get" what it's doing. but it looks sensible 🤷

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -17,216 +17,113 @@
...
 
-const noOp = () => undefined
-const vizFn = function (
-  el: SVGElement,
-  measureSize: MeasureSizeFn,
-  graph: Graph,
-  layout: Layout,
-  style: GraphStyle
-): VizObj {
-  const viz: VizObj = {
```
OskarDamkjaer, line 59: Good that you got rid of this. This functionality really feels like it belongs in a class 👍

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -17,216 +17,113 @@
...
-
-  const onRelationshipClick = function (
-    event: Event,
-    relationship: Relationship
+  private draw = false
+  private isZoomClick = false
+
+  constructor(
+    element: T,
+    private measureSize: MeasureSizeFn,
```
OskarDamkjaer, line 61: What does it mean to define a constructor argument as private? never seen this pattern befor

src/browser/modules/D3Visualization/lib/visualization/components/ForceSimulation.ts
```diff
@@ -0,0 +1,103 @@
...
+      .force('centerY', forceY(0).strength(FORCE_CENTER_Y))
+      .on('tick', () => {
+        this.simulation.tick(TICKS_PER_RENDER)
+        render()
+      })
+      .stop()
+  }
+
+  updateNodes(graph: Graph) {
+    const nodes = cloneArray(graph.nodes())
```
OskarDamkjaer, line 70: is d3 still mutating the elements? why do we need to clone here

src/browser/modules/D3Visualization/lib/visualization/components/ForceSimulation.ts
```diff
@@ -0,0 +1,103 @@
...
+import cloneArray from '../utils/arrays'
+import circularLayout from '../utils/circularLayout'
+import Graph from './Graph'
+import Relationship from './Relationship'
+import VizNode from './VizNode'
+
+const oneRelationshipPerPairOfNodes = (graph: Graph) =>
+  Array.from(graph.groupedRelationships()).map(pair => pair.relationships[0])
+
+export class ForceSimulation {
```
OskarDamkjaer, line 52: neat that you break out the rel/node parts of the force layout. A bit tricky to compare with the old code without d3 skills, but it's worked well in my testing and looks clean

src/browser/modules/D3Visualization/lib/visualization/components/ForceSimulation.ts
FallingCeilingS: Same question here: https://github.com/neo4j/neo4j-browser/pull/1652#discussion_r791540892.

src/browser/modules/D3Visualization/lib/visualization/components/mouseEventHandlers.ts
nglgzz: I'm not entirely sure what they represent, I copied this from the D3 docs. I'll add a link to it in case we actually need to understand it better later on. Also, I tried using the method that supposedly returns a function that calculates this, but couldn't get it to work (it seems like it's requiring different arguments than the documented ones).

src/browser/modules/D3Visualization/components/Graph.tsx
```diff
@@ -52,141 +49,81 @@ type GraphProps = {
   offset: any
 }
 
-export class GraphComponent extends Component<GraphProps, GraphState> {
-  graph: Graph | undefined
-  graphEventHandler: GraphEventHandler | undefined
-  graphView: GraphView | undefined
-  svgElement: any
+export function GraphComponent(props: GraphProps) {
```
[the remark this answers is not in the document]
nglgzz: Visually I didn't notice any difference, but I didn't measure performance. Nice to know though, I didn't think about that before.

src/browser/modules/D3Visualization/lib/visualization/components/mouseEventHandlers.ts
eijawerner: Sounds good to add a comment and a link to make it clear that it is the default function we are using. I also tried to use the method that returns this default function and it is actually just typescript that complains, when I added ts-ignore it worked in runtime, so it seems to be a bug in the type definitions of this. Anyway I guess it is not good to have ts-ignore so I think it is better you add a comment and a link for now and we can update it later if the typescript problem is resolved.

src/browser/modules/D3Visualization/components/Graph.tsx
```diff
@@ -52,141 +49,81 @@ type GraphProps = {
...
-      this.graphEventHandler?.onItemMouseOut()
+    setGraphView(graphView)
+  }, [svgElement.current])
+
+  useEffect(() => graphView?.init(), [svgElement.current, props.styleVersion])
+  useEffect(() => graphView?.resize(), [svgElement.current, props.isFullscreen])
+
+  const zoomInClicked = (): void => {
+    if (graphView) {
+      graphView.zoomIn()
```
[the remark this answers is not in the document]
eijawerner: When I dig into the details of zoom while working on the zoom interactions card I actually found a way to get the current zoom level. However you could not get it at anytime, but when a zoom event just has happened. So it can be extracted in the function which is called on all zoom events (``.on('zoom', (e: D3ZoomEvent<SVGGElement, unknown>)...``), and get the current scale with e.transform.k. So I guess we could save that value to the state and from there we can access it whenever we want.

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
nglgzz: This is the same as defining a private class property with the same name and then assigning it in the constructor: https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
nglgzz: I tried but it was a bit tricky since it's referencing some class properties (draw and isZoomClick) that are changed by some of the other methods. I could in theory pass `this` to it, but it didn't feel like a clean solution, so I left as is for now.

src/browser/modules/D3Visualization/components/Graph.tsx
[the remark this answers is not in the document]
nglgzz: Nice! That's good to know. I think it would be best to keep it for a separate PR, given that there are already a lot of changes here.

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
nglgzz: Personally I prefer the more explicit version, but I didn't think about the rest of the codebase, will update the code to be more consistent with everything else.

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
nglgzz: Good spot, I missed this :+1:

OskarDamkjaer (approved):
All my points are addressed, happy to merge this when jun/eija are happy with it as well 👍

eijawerner (changes requested):
Really nice cleanup! ⭐ Only thing is that zoom buttons are not working, see comment. When you have fixed that I am ready to approve.

src/browser/modules/D3Visualization/components/Graph.tsx
```diff
@@ -44,149 +42,120 @@ type GraphProps = {
...
+    })
 
-  zoomOutClicked(el: any): void {
-    if (this.graphView) {
-      const limits = this.graphView.zoomOut(el)
-      this.setState({
-        zoomInLimitReached: limits.zoomInLimit,
-        zoomOutLimitReached: limits.zoomOutLimit
+    const graph = createGraph(nodes, relationships)
+    const graphView = new GraphView(
```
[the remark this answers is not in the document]
nglgzz: Good catch, updated!

eijawerner (approved):
⭐

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
FallingCeilingS: Same question here.

FallingCeilingS (approved):
LGTM! Thanks for the cleanups and great conversation in this PR. I have a small question below.
Also, it seems like `circumferentialRelationshipRouting.ts` and `circumferentialRelationshipDistribution.ts` are not used anywhere but they depend on each other. Can you double-check and probably remove them if they're not used anywhere else?

eijawerner:
> LGTM! Thanks for the cleanups and great conversation in this PR. I have a small question below. Also, it seems like `circumferentialRelationshipRouting.ts` and `circumferentialRelationshipDistribution.ts` are not used anywhere but they depend on each other. Can you double-check and probably remove them if they're not used anywhere else?

I am working in a branch based on this PR where I am adding the functionality of disabling zoom buttons when limit is reached, so we can leave the ``return limitsReached`` as it is in this PR and it will be cleaned up in that following PR.

PR #1653, opened 2022-01-28
Disable zoom buttons when limit reached

eijawerner:
Also bugfix issue where wheel zoom is not working in safari.


src/browser/modules/D3Visualization/components/Graph.tsx
```diff
@@ -123,6 +133,13 @@ export class GraphComponent extends React.Component<GraphProps> {
     }
   }
 
+  handleZoomEvent = (limitsReached: ZoomLimitsReached) => {
+    this.setState({
```
nglgzz, line 137: Since this is called every time the zoom changes, it can trigger loads of re-renders, especially when zooming with the scroll wheel. It might be worth setting the state only if it changed.

eijawerner: Nice catch! I'll fix!

nglgzz (approved):
Great that we can still have this piece of functionality :sparkles:

PR #1655, opened 2022-01-28
Use shift, cmd or ctrl as mod key for wheel zoom

eijawerner:
also in fullscreen no mod key is required for wheel zoom.


nglgzz (approved):
Looks good :sparkles:

PR #1657, opened 2022-01-31
Added button for zooming to fit to available space for graph and updated styles for zoom

eijawerner:
- Now possible to zoom to fit the whole graph
  - Avoid zooming in more than max limit for really small graphs, currently no limit for zooming out when pressing the zoom to fit button.
  (- NOT included: to be done in later PR, zoom min limit should be set dynamically to be the same as the scale needed to fit graph)
- Also added new icons for zoom in/out and fixed styling according to latest designs.
- Changed old icon buttons to have the same background states as in the new designs, however doesn't change the color yet, since the old icons are smaller and thinner it makes them more difficult to read. Remove the legacy color when all new icons are in place.
(- NOT included: to be done in later PR, get the icons and color from design library when that is released)

New:
Screenshot 2022-01-31 at 14 42 24
Old:
Screenshot 2022-01-31 at 14 42 05


src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -230,12 +235,42 @@ export class Visualization {
...
     this.zoomBehavior.scaleBy(this.root, isZoomingIn ? 1.3 : 0.7)
   }
 
+  private handleZoomToFitClick(): void {
+    this.draw = true
+    this.isZoomClick = true
+    this.zoomToFitWholeGraph()
+  }
+
+  private zoomToFitWholeGraph = () => {
```
nglgzz, line 250: Considering that they're always used together, wouldn't it make sense to combine `handleZoomToFitClick`, `zoomToFitWholeGraph` and `zoomToFitClick` into one function?

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -20,8 +20,13 @@
 import { easeCubic } from 'd3-ease'
 import { BaseType, Selection, select as d3Select } from 'd3-selection'
 import { D3ZoomEvent, ZoomBehavior, zoom as d3Zoom } from 'd3-zoom'
+import * as d3Zooming from 'd3-zoom'
```
nglgzz, line 23: Unsure if we're doing any tree shaking, but even just for consistency it would be best to import only the functions we're using.

```suggestion
import { D3ZoomEvent, ZoomBehavior, zoom as d3Zoom, zoomIdentity } from 'd3-zoom'
```

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
eijawerner: I see what you mean, however I like that it is clear connections to the corresponding zoom in/out functions and the common stuff needed to be done, so I will reduce the functions and make that connection clearer by combining those three clicks into one handler.

eijawerner:
> Is it possible to centre the graph taking the panel width into account? If it's expanded it looks a bit nicer having it be centred of the available space. https://user-images.githubusercontent.com/10564538/151823616-43c5eff6-447e-4a3d-83d5-a30058cf42fd.mp4

I agree that it would be nice, however I'd like to keep the scope down of this PR and this is current behavior in Bloom as well, so I keep it as it is for now.

nglgzz (approved):

PR #1660, opened 2022-02-02
bugfix center the graph on zoom to fit not always working

eijawerner:
Graphs that didn't have their midpoint in the centre (when no transforms are applied) didn't get centered when clicking zoom to fit. Now that is fixed by adjusting to have their mid point at the centre of the svg after zoom to fit is applied.

before:
Screenshot 2022-02-02 at 15 52 58
after:
Screenshot 2022-02-02 at 15 52 32

and here is how the graph looks in both cases before any transforms are applied, you can see that there is more space above than below:
Screenshot 2022-02-02 at 15 56 33


nglgzz (approved):
:+1:

PR #1667, opened 2022-02-09
Always run frame queries in web workers

nglgzz:
We currently have a setting to choose whether to run frame queries in web workers or in the main thread. By default we run queries on web workers, this setting doesn't seem to be surfaced to the user, and nobody has changed it in the past year, so it's a good time to remove it.


src/shared/services/bolt/bolt.ts
```diff
@@ -82,135 +77,104 @@ function routedWriteTransaction(
...
-      id,
-      payload,
-      onLostConnection
-    )
-    return workerPromise
-  } else {
-    return globalDirectTransaction(input, parameters, {
-      requestId,
-      cancelable,
+  const id = requestId || v4()
```
OskarDamkjaer, line 158: should we handle the case where the user is some really old browser that doesn't have window.Worker?

nglgzz: I could throw an error I guess. Otherwise I'm not sure. Do we have any fallback currently for unsupported browsers?

OskarDamkjaer: We don't have any other fallbacks, but what we could do is warn earlier in the application. I think it'd make sense to change `:play start` to a message warning about the browser being unsupported or something like that

OskarDamkjaer: It'd say it's not something you'd need to do

OskarDamkjaer (approved):
Nice to get rid of this old cruft!

PR #1669, opened 2022-02-09
Visualization: reduce jiggling of nodes when interacting with graph

nglgzz:
When dragging a node that has lots of neighbors, the neighboring nodes keep jiggling, making the visualization look worse than it did before. This PR is addressing this by increasing friction to make the nodes stabilize more quickly, and reducing the alpha value we use when dragging to reduce the movement of nodes that aren't being dragged.

The preview of this change is available here: http://browser-visualization-tweaks.surge.sh/


OskarDamkjaer (approved):
Worked well in my testing, nice to get some final tweaks in

PR #1672, opened 2022-02-10
Remove viz limit of 1000

eijawerner:
Now it is possible to select the graph view even if the results contains more than a 1000 record fields. The other limit which truncates the record fields data is still there.
 
Screenshot 2022-02-10 at 14 14 06


OskarDamkjaer:
I'm solving there flaky tests here: https://github.com/neo4j/neo4j-browser/pull/1674

PR #1673, opened 2022-02-10
Make sure can use zoom out button to fit whole graph

eijawerner:
... also for larger graphs

Screenshot 2022-02-10 at 14 33 03


src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -248,20 +249,43 @@ export class Visualization {
...
   }
 
-  private handleZoomClick(zoomType: ZoomType): void {
+  private handleZoomClick = (zoomType: ZoomType): void => {
     this.draw = true
     this.isZoomClick = true
 
     if (zoomType === ZoomType.IN) {
-      this.zoomBehavior.scaleBy(this.root, 1.3)
+      this.zoomTo(1.3)
```
OskarDamkjaer, line 257: Why not leave these functions as they are and just use the new function for when we have a point to zoom to?

OskarDamkjaer: Since the function seems to be doing two different things now

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -248,20 +249,43 @@ export class Visualization {
...
     } else if (zoomType === ZoomType.OUT) {
-      this.zoomBehavior.scaleBy(this.root, 0.7)
+      this.zoomTo(0.7)
     } else if (zoomType === ZoomType.FIT) {
-      this.zoomToFitWholeGraph()
+      const scaleAndOffset = this.getZoomScaleFactorToFitWholeGraph()
+      if (scaleAndOffset) {
+        const { scale, centerPointOffset } = scaleAndOffset
+        // Do not zoom in more than zoom max scale for really small graphs
+        this.zoomTo(Math.min(scale, ZOOM_MAX_SCALE), {
```
OskarDamkjaer, line 265: We could use `centerPointOffset` directly here right? No need to create the new object?

OskarDamkjaer: Like so: `this.zoomTo(Math.min(scale, ZOOM_MAX_SCALE), centerPointOffset))`

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
```diff
@@ -248,20 +249,43 @@ export class Visualization {
...
+          .scale(Math.min(scale, ZOOM_MAX_SCALE))
+          .translate(point.x, point.y)
+      )
+    } else {
+      this.zoomBehavior.scaleBy(this.root, scale)
     }
   }
 
-  private zoomToFitWholeGraph = () => {
+  private getZoomScaleFactorToFitWholeGraph = ():
```
OskarDamkjaer, line 286: Nice to reuse this logic 👍

src/browser/modules/D3Visualization/lib/visualization/components/Visualization.ts
eijawerner: Sure, I was thinking to have the same abstraction level of the different cases, but I can see that it doesn't really help here since the `this.zoomBehavior.scaleBy(this.root, 1.3)` is pretty easy to read as it is. I will create a zoomToFit() function instead and put all the logic for that there and removing the zoomTo function.

OskarDamkjaer (approved):

PR #1674, opened 2022-02-11
Fix flaky test caused by debouncing saving to localstorage

OskarDamkjaer:
What happens is we click to remove a guide and sometimes reload before it's been removed from local storage -> it's still there after refresh.


nglgzz (approved):
Nice catch :+1:

PR #1676, opened 2022-02-11
Bolt worker and work pool cleanup

nglgzz:
Follow up from https://github.com/neo4j/neo4j-browser/pull/1661. As in the previous PR, there are no functional changes here, only simplifying the code to make it easier to understand.


OskarDamkjaer (approved):
Nice changes! Much cleaner when the "work" doesn't have a reference to it's "worker"

PR #1678, opened 2022-02-14
Isolate graph visualization 

nglgzz:
This is the first step towards exporting the visualization as a component. There are no changes in behavior in this PR, just reorganizing the code. It moves the graph visualization into a separate directory, detaches it from the redux state, and updates the eslint rules to only allow importing this directory as if it was an external library (and also disallowing importing from adjacent directories in this folder). 

I also removed the GraphView file as it was a useless abstraction (each method was deferring to the Visualization class).


src/graphVisualization/utils/utils.test.ts
```diff
@@ -0,0 +1,40 @@
+import { selectorArrayToString, selectorStringToArray } from './utils'
```
OskarDamkjaer, line 1: no copyright header

OskarDamkjaer (approved):
A few nits, good work! I like the new names and the removed layer of abstraction with the GraphView

OskarDamkjaer (approved):

PR #1680, opened 2022-02-14
Handle null values in graph viz

OskarDamkjaer:
We stringify all node property values except `null`, which causes crashes later on. This PR makes we handle null, and that the correct type hint show up as well. Preview at http://handle_null_properties.surge.sh, easiest way to trigger the crash is run `call apoc.create.vNode(['stuff'], {id:5, blah:null})` and hover the created node. 

Here's a screen shot of how it looks now: 
 
 CleanShot 2022-02-14 at 19 03 49

### DIFF
--- a/src/browser/modules/D3Visualization/mapper.ts
+++ b/src/browser/modules/D3Visualization/mapper.ts
@@ -25,7 +25,9 @@ import { optionalToString } from 'services/utils'
 
 const mapProperties = (_: any) => Object.assign({}, ...stringifyValues(_))
 const stringifyValues = (obj: any) =>
-  Object.keys(obj).map(k => ({ [k]: optionalToString(obj[k]) }))
+  Object.keys(obj).map(k => ({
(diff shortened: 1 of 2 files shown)


src/browser/modules/D3Visualization/mapper.ts
```diff
@@ -25,7 +25,9 @@ import { optionalToString } from 'services/utils'
 
 const mapProperties = (_: any) => Object.assign({}, ...stringifyValues(_))
 const stringifyValues = (obj: any) =>
-  Object.keys(obj).map(k => ({ [k]: optionalToString(obj[k]) }))
+  Object.keys(obj).map(k => ({
+    [k]: obj[k] === null ? 'null' : optionalToString(obj[k])
```
nglgzz, line 29: Might want to get rid of the `optionalToString` function as it's used only here. Not necessary though

nglgzz (approved):

PR #1681, opened 2022-02-14
Remove unneeded listConfig call in sys info frame

OskarDamkjaer:
After the `metrics.prefix` and `metrics.namespaces.enabled` settings were added to `clientConfig`, we no longer need the work-around of calling `listConfig` directly in the sysinfo frame since it's already called when loading the server metadata. 

I've made sure it works even if the settings haven't been read yet as well. Tested on `4.4.0` with settings: 
metrics.prefix=abc
metrics.namespaces.enabled=true

Preview at http://remove_listConfig_call.surge.sh


nglgzz (approved):
Looks good to me :sparkles:

PR #1682, opened 2022-02-15
Improve causal cluster detection

OskarDamkjaer:
From 4.3 we can't use the presence of `dbms.cluster.overview` to know if we're connected to a cluster, but the same version added `SHOW DATABASES` which can.


src/shared/modules/features/featuresDuck.ts
```diff
@@ -41,8 +41,6 @@ export const getAvailableProcedures = (state: any) =>
 
 // having this procedure used to indicate that we're on a cluster
 // but from 4.3 stand alone instances also have it
```
nglgzz, line 43:
```suggestion
```

nglgzz (approved):
Looks good, just one comment

PR #1686, opened 2022-02-19
add connection related metrics event

spgandhi:
This PR adds DB connections related metrics events to be sent to Segment

 - when user submits the connection form
 - when the connection is successful

These events will help the Growth do Activation analysis better and be able to see drop off because of any connections related frictions


OskarDamkjaer (approved):
Looks good! Thanks for your patience in getting the tests through! :)

PR #1689, opened 2022-02-22
Remove relate by UI dependency

eijawerner:
Removed the dependency to relate-by-ui.
1. Copied the code for the relatable for now. 
2. Removed most of the dependencies to semantic-ui as well (icons switched to font awesome counterparts), but all the dependencies will be cleanup in other PRs when restyling the UI overall.
3. Removed the dependency "@trivago/prettier-plugin-sort-imports" since it reorders also imports with side effects, like the order of the import of the style files. When this is released (with an option to not reorder imports with side effects) https://github.com/trivago/prettier-plugin-sort-imports/pull/111 we can add the plugin again.
4. Reordered the style imports back to as it was in release 4.3.2.

to check how the Browser looks like in this PR you can visit: http://browser-remove-deps.surge.sh

Known changes:

- The arrows were removed in the links page, and approved by UX that it is ok.

Screenshot 2022-02-25 at 13 25 20
Screenshot 2022-02-25 at 13 25 11

- some spacing (padding/margins) and sizes (icons) are slightly different but are also OK from UX since no major issues, we will go through the whole UI when restyling the UI overall.


src/browser/modules/Stream/CypherFrame/RelatableView/relatable/global.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
+  const content: any
+  export default content
```
OskarDamkjaer, line 3: I think we already have this type in the our file called `custom.d.ts` so it's not needed here

src/browser/modules/Stream/CypherFrame/RelatableView/relatable-view.tsx
```diff
@@ -90,12 +91,12 @@ function getColumns(records: Record[], maxFieldItems: number) {
...
   return map(keys, key => ({
     Header: key,
-    accessor: (record: Record) => {
+    accessor: (record: Record): string => {
       const fieldItem = record.get(key)
 
       if (!Array.isArray(fieldItem)) return fieldItem
 
-      return slice(fieldItem, 0, maxFieldItems)
+      return slice(fieldItem, 0, maxFieldItems).join(',')
```
OskarDamkjaer, line 99: I remember this is something we tried to get the types to work properly. But we should remove it, as it causes the following bug:
CleanShot 2022-02-28 at 12 12 09@2x


How it should look: 
bild

OskarDamkjaer: (I tested and it works fine when we change it back, the :string type annotation also needs to be removed)

src/browser/modules/Stream/CypherFrame/RelatableView/relatable-view.tsx
```diff
@@ -80,6 +80,7 @@ export function RelatableViewComponent({
 
   return (
     <RelatableStyleWrapper>
+      {/* @ts-ignore */}
```
OskarDamkjaer, line 83: I played around a little bit with this because it was odd to get a new error 🤔 

It seemed like the error was from a mismatch in types -> you got the exact version of react-table from relate-by-ui, but the @types/react-table are version `7.0.16`, when I changed it to be the same as the react-table `7.0.4` the problem went away!

OskarDamkjaer (changes requested):
Nice work! The getting rid of this dependency had a long time coming, good to get all the unused code into the open...  Let's hope we can get rid of that as well soon! 

I left a few comments for you to have a look at! 💯

src/browser/modules/Stream/CypherFrame/RelatableView/relatable-view.tsx
eijawerner: ah yes, good catch, thanks!

package.json
```diff
@@ -217,6 +217,7 @@
     "react-resizable": "^3.0.4",
     "react-suber": "1.0.4",
     "react-svg-inline": "^2.1.1",
+    "react-table": "7.0.4",
     "react-timeago": "^4.1.9",
     "redux": "^3.7.2",
     "redux-observable": "^0.16.0",
```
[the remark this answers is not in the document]
eijawerner: ah thanks, the relate-by-ui deps should be removed yes. However the semantic-ui is still a dependency for the relatable code and for Popup when styling nodes/relationships (we could build our own but I didn't do it in this PR to keep the scope down).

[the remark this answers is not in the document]
OskarDamkjaer: Ah right!

src/browser/modules/Stream/CypherFrame/RelatableView/relatable-view.tsx
eijawerner: I get some other ts errors when updating to that version.

OskarDamkjaer: Looks like I was a bit quick to say it worked.. Let's keep it as is then

OskarDamkjaer (approved):
Changes look good! Let's merge on green ✨

PR #1690, opened 2022-02-25
New rectangle-shaped force.

apcj:
Makes better use of screen area.
Currently the aspect ratio is fixed, but should be
chosen more intelligently.
Also, the strength parameter will need some tuning.


OskarDamkjaer:
Here's a preview! http://rectangle-layout.surge.sh/

PR #1700, opened 2022-03-11
Add build date and revision

OskarDamkjaer:
When we switch to deploying more often in aura, it makes sense to be able to see when a version was built and what exact git revision it was as you can no longer rely only on the version number.

demo at http://add_build_date_and_revision.surge.sh

### DIFF
--- a/build_scripts/generate-manifest-helpers.js
+++ b/build_scripts/generate-manifest-helpers.js
@@ -37,6 +37,27 @@ function writeDataToFile(file, data) {
   }
 }
 
+function getCurrentGitRevision() {
+  try {
+    const rev = fs
+      .readFileSync('.git/HEAD')
+      .toString()
+      .trim()
+      .split(/.*[: ]/)
+      .slice(-1)[0]
+    if (rev.indexOf('/') === -1) {
+      return rev
+    } else {
+      return fs
+        .readFileSync('.git/' + rev)
+        .toString()
+        .trim()
+    }
+  } catch (e) {
+    throw new Error('Could not read git revision. Error: ' + e)
+  }
+}
+
 function buildTargetObject(data, dataProp) {
   const out = {}
   const keys = data[dataProp] || []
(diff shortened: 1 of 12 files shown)


build_scripts/generate-manifest-helpers.js
```diff
@@ -37,6 +37,33 @@ function writeDataToFile(file, data) {
...
+      .split(/.*[: ]/)
+      .slice(-1)[0]
+    if (rev.indexOf('/') === -1) {
+      return rev
+    } else {
+      return fs
+        .readFileSync('.git/' + rev)
+        .toString()
+        .trim()
+    }
```
eijawerner, line 59: Please add a comment or link to inspiration for this and remove support for detached head which I don't think we need, to simplify things.

src/shared/modules/app/appDuck.ts
```diff
@@ -46,6 +47,10 @@ export const inWebEnv = (state: any) => getEnv(state) === WEB
 export const inCloudEnv = (state: any) => getEnv(state) === CLOUD
 export const inWebBrowser = (state: any) => [WEB, CLOUD].includes(getEnv(state))
 export const inDesktop = (state: any) => getEnv(state) === DESKTOP
+export const getGitRevision = (state: any): string | null =>
```
eijawerner, line 50: Add GlobalState type?

src/shared/modules/app/appDuck.ts
```diff
@@ -46,6 +47,10 @@ export const inWebEnv = (state: any) => getEnv(state) === WEB
...
+  state[NAME].gitRevision ?? null
+export const getBuiltAt = (state: any): string | null =>
```
eijawerner, line 52: same

eijawerner (commented):
Nice! Added some comments.

src/shared/modules/app/appDuck.ts
OskarDamkjaer: App wasn't added properly on the global state, but I went ahead and added it now :+1:

eijawerner (approved):
⭐ Nice to have this info in the future!

PR #1710, opened 2022-03-22
 Fixes for integrating graph viz in nx

eijawerner:
Set default values for Explorer

Exchanged font awesome chevron icon for hero icon from ndl.

Fixed so resizing of node properties panel works when used in nx, by including missing styles in the library.


linuslundahl (approved):
Had a discussion with Eija around the code and we decided on some minor changes. Everything else looks good and go ahead and merge when tests are green.

PR #1719, opened 2022-04-01
Removed svg inline dependency from arc

eijawerner:
putting Browser legacy icons back in browser and making sure all icons used in arc library are from neeedle.


OskarDamkjaer (approved):
Great to get rid of that old dependency (at least in arc)

PR #1727, opened 2022-04-04
Publish monaco in arc

OskarDamkjaer:
- Added needed dependencies
- Renamed the Monaco component
- Added the more default props
- Switched resize observable dependency to one without type errors :+1:

### DIFF
--- a/src/neo4j-arc/cypher-language-support/cypher-editor/CypherEditor.tsx
+++ b/src/neo4j-arc/cypher-language-support/cypher-editor/CypherEditor.tsx
@@ -30,6 +30,7 @@ import {
 import { QueryResult } from 'neo4j-driver-core'
 import React from 'react'
 import styled from 'styled-components'
+import { ResizeObserver } from '@juggle/resize-observer'
 
 const shouldCheckForHints = (code: string) =>
   code.trim().length > 0 &&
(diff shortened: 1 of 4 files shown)


src/neo4j-arc/cypher-language-support/cypher-editor/CypherEditor.tsx
```diff
@@ -27,19 +27,21 @@ import {
...
   !code.trimLeft().startsWith(':') &&
   !code.trimLeft().toUpperCase().startsWith('EXPLAIN') &&
   !code.trimLeft().toUpperCase().startsWith('PROFILE')
 
-export type MonacoHandles = Monaco
+export type CypherEditorHandles = CypherEditor
 
+/*
+// Using this gives "invalid hook call" for some reason
+import styled from 'styled-components'
```
OskarDamkjaer, line 44: Forgot to fix this.. On it!

src/neo4j-arc/cypher-language-support/cypher-editor/CypherEditor.tsx
```diff
@@ -57,44 +55,68 @@ const MonacoStyleWrapper = styled.div`
...
 const EDITOR_UPDATE_DEBOUNCE_TIME = 300
-type MonacoDefaultProps = {
-  value: string
-  onDisplayHelpKeys: () => void
-  onChange: (value: string) => void
-  onExecute: (value: string) => void
-}
-
-export type MonacoProps = MonacoDefaultProps & {
+type CypherEditorDefaultProps = {
```
eijawerner, line 58: There is a Required util in typescript, which converts optionals to required, so suggest to ``type CypherEditorDefaultProps = Required<CypherEditorProps>`` to not have it defined twice.

package.json
```diff
@@ -173,9 +172,10 @@
   },
   "dependencies": {
     "@apollo/client": "^3.2.0",
-    "@neo4j/browser-lambda-parser": "1.0.5",
+    "@juggle/resize-observer": "^3.3.1",
```
eijawerner, line 175: can we remove this all together since all browsers that we support now don't need this?

eijawerner (commented):
Nice! ⭐ See comments.

eijawerner (approved):
💯

package.json
OskarDamkjaer: I removed this but then noticed our typescript version didn't have types for the resize-observer yet, started upgrading it which meant I had to fix new type errors, install a new version of ts-jest and so on. Turned out to be too much of a rabbit hole for now, so I'd prefer to keep the polyfill in for now :+1:

PR #1732, opened 2022-04-20
Repair autocomplete

OskarDamkjaer:
Fix bug causing proper autocomplete not to work. Adds a test to make sure it keeps working.


eijawerner (approved):
⭐ Great that you've added a test as well 👍 . LGTM on green.

PR #1733, opened 2022-04-22
Add a note that `+ssc` schemes are not supported

renetapopova:
Depends on https://github.com/neo-technology/neo4j-manual-modeling/pull/2774, which should be merged first.

### DIFF
--- a/docs/modules/ROOT/pages/visual-tour.adoc
+++ b/docs/modules/ROOT/pages/visual-tour.adoc
@@ -198,7 +198,7 @@ To access the list of all available editor keybindings, press F1 in the Cypher e
 You can click a block shown in the history result to bring it into the editor.
 The command history is persisted across Browser restarts.
 
-| `CALL db.schema.visualization()`
+| `CALL db..visualization()`
 | Show database schema.
 
 | `:sysinfo`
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/visual-tour.adoc
```diff
@@ -198,7 +198,7 @@ To access the list of all available editor keybindings, press F1 in the Cypher e
 You can click a block shown in the history result to bring it into the editor.
 The command history is persisted across Browser restarts.
 
-| `CALL db.schema.visualization()`
+| `CALL db..visualization()`
```
gregoryking, line 201: Is this a find and replace editing error?

gregoryking (changes requested):
Looks good overall, just noticed one file possibly has a mistaken edit in it.

docs/modules/ROOT/pages/visual-tour.adoc
renetapopova: Of course, good spot!!

renetapopova: Aah, yes, it is a replace error. Thanks!

renetapopova, line 201:
```suggestion
| `CALL db.schema.visualization()`
```

PR #1746, opened 2022-06-03
remove unneeded metadata-version check

OskarDamkjaer:
Before we had to be use that we were running against a 3.5+ database before we send metadata so we couldn't be sure to always send it. Since 3.4 is long out of support I'm removing the check


linuslundahl (approved):
LGTM

PR #1747, opened 2022-06-07
Use cluster role from show databases if available

OskarDamkjaer:
In modern version of neo4j the cluster role is available through `SHOW DATABASES` so we don't have to run `CALL dbms.cluster.role() YIELD role` to get it.


src/shared/modules/features/versionedFeatures.ts
```diff
@@ -25,6 +25,7 @@ import { guessSemverVersion } from './featureDuck.utils'
 import { GlobalState } from 'project-root/src/shared/globalState'
 
 const NEO4J_4_0 = '4.0.0-alpha01'
+const NEO4J_5_0 = '4.0.0-alpha01'
```
eijawerner, line 28: copy mistake :)

eijawerner (approved):

PR #1749, opened 2022-06-08
Settings 5.0

eijawerner:
Update settings for neo4j 5.0:
    
    Some neo4j settings renamed in 5.0
    - metrics.namespaces.enabled -> server.metrics.namespaces.enabled
    - metrics.prefix -> server.metrics.prefix
    - clients.allow_telemetry -> client.allow_telemetry
    
    Set authEnabled to true until setting is read to be on the safe side.
    
    Added support for duration that includes all durations that are supported for https://neo4j.com/docs/operations-manual/4.4/reference/configuration-settings/#config_browser.credential_timeout

  Renamed redux meta settings variables to easier distinguish from config names.
    
  Removed handling allowlist rename of whitelist since neo4j config does not support that yet.


src/shared/modules/dbMeta/dbMetaDuck.ts
```diff
@@ -90,6 +90,36 @@ export type Procedure = {
...
+ */
+export const initialClientSettings: ClientSettings = {
+  allowOutgoingConnections: false, // default is true, but set to false until settings read
+  credentialTimeout: 0,
+  postConnectCmd: '',
+  remoteContentHostnameAllowlist: 'guides.neo4j.com, localhost',
+  retainConnectionCredentials: false, // default is true, but set to false until settings read
+  retainEditorHistory: false, // default is true, but set to false until settings read
+  allowTelemetry: true, // default is true. Renamed to client.allow_telemetry after 5.0
+  authEnabled: true, // default is true, but set to false until settings read
```
OskarDamkjaer, line 119: This comment is outdated

OskarDamkjaer: We could probably leave out the comments about renaming here as well since you have renaming comments further down :+1:

src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -437,6 +416,41 @@ export const serverConfigEpic = (some$: any, store: any) =>
...
+      isConfigValFalsy(rawSettings['clients.allow_telemetry']) ||
+      isConfigValFalsy(rawSettings['client.allow_telemetry'])
+    ), // default true
+    authEnabled: !isConfigValFalsy(rawSettings['dbms.security.auth_enabled']), // default true
+    // Info: metrics... in versions < 5.0, server.metrics... in versions >= 5.0
+    metricsNamespacesEnabled:
+      isConfigValTruthy(rawSettings['metrics.namespaces.enabled']) ||
+      isConfigValTruthy(rawSettings['server.metrics.namespaces.enabled']), // default false
+    metricsPrefix:
+      rawSettings['metrics.prefix'] ||
```
OskarDamkjaer, line 446: In 5.0 this would just be "undefined" i presume?

eijawerner: yes it is undefined

OskarDamkjaer (approved):

PR #1752, opened 2022-06-13
Separate new & old queries frame, use SHOW TRANSACTIONS instead of dbms.listQueries

OskarDamkjaer:
Also updates a few e2e tests that were failing on 5.0


OskarDamkjaer:
Reviewed offline w/ eija :+1:

PR #1755, opened 2022-06-13
Add url param to preselect in connect frame

OskarDamkjaer:
Adds param `preselectAuthMethod` to preselect auth method in the connection form


OskarDamkjaer:
Reviewed offline w/ eija :+1:

PR #1765, opened 2022-07-05
feat: surface zoom to fit by default prop

jchen042:
This PR is intended to dogfood load preview feature in Data Importer. Graph visualisation should be able to zoom to fit in initial rendering so users can view the whole graph.

https://user-images.githubusercontent.com/26452483/177345014-7466943e-8be4-4bcf-b49a-e780b6f49edf.mov

PR #1773, opened 2022-07-07
fix: graph is no simulation in initial rendering

FallingCeilingS:
Based on #1765 , a bug is identified that no simulation is triggered when the graph is initially rendered until the user's first interaction.

https://user-images.githubusercontent.com/26452483/177783591-ac2cbf1c-e292-4fbb-b4fd-cba9cc0510db.mov

Had dig into the root issue, I found that's due to the `forceSimulation.restart()` is not triggered when initialising the graph so  no simulation ticks and that's why `forceSimulation.endSimulationCallback()` is not triggered.
`forceSimulation.restart()` is only triggered when calling `visualization.update({restartSimulation: true})`. However, `visualization.update()` is not called if no "Connect to result nodes" option is toggled. In the current logic, when initialising the graph, the logic will be conditionally triggered".
```ts
    if (getAutoCompleteCallback) {
      getAutoCompleteCallback((internalRelationships: BasicRelationship[]) => {
        graph.addInternalRelationships(
          mapRelationships(internalRelationships, graph)
        )
        onGraphModelChange(getGraphStats(graph))
        this.visualization?.update({
          updateNodes: false,
          updateRelationships: true
        }) // this will trigger forceSimulation.restart() and then forceSimulation.endSimulationCallback()
        graphEventHandler.onItemMouseOut()
      })
    }
```
 
Screenshot 2022-07-07 at 14 23 49

The simplest fix is to `restart` simulation in initial render rather than `precompute`.

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/VisualizationView/__snapshots__/VisualizationView.test.tsx.snap
+++ b/src/browser/modules/Stream/CypherFrame/VisualizationView/__snapshots__/VisualizationView.test.tsx.snap
@@ -39,7 +39,6 @@ exports[`Visualization renders with result and escapes any HTML 1`] = `
                 <g
                   aria-label="graph-node1"
                   class="node"
-                  transform="translate(0,0.9475741398926749)"
                 >
                   <circle
                     class="ring"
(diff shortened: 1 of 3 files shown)


FallingCeilingS:
[Load preview UI functionality](https://trello.com/c/3iqf8g43/304-load-preview-ui-functionality)

OskarDamkjaer (approved):

PR #1787, opened 2022-08-03
Remove dynamic import

OskarDamkjaer:
The dynamic import led to warnings in the console from styled components

### DIFF
--- a/src/browser/modules/Stream/FrameContainer.tsx
+++ b/src/browser/modules/Stream/FrameContainer.tsx
@@ -47,6 +47,7 @@ import StyleFrame from './StyleFrame'
 import SysInfoFrame from './SysInfoFrame/SysInfoFrame'
 import { Connection } from 'shared/modules/connections/connectionsDuck'
 import { FrameStack } from 'shared/modules/frames/framesDuck'
+import extras from './Extras/index'
 
 const nameToFrame: Record<string, React.ComponentType<any>> = {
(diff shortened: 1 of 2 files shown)


src/browser/modules/Stream/FrameContainer.tsx
```diff
@@ -78,14 +79,15 @@ const nameToFrame: Record<string, React.ComponentType<any>> = {
...
-      const command = cmd.replace(/^:/, '')
-      const Frame = command[0].toUpperCase() + command.slice(1) + 'Frame'
-      MyFrame = require('./Extras/index')[Frame] || nameToFrame['error']
-    } catch (e) {}
+  if (type === 'error') {
+    const command = cmd.replace(/^:/, '')
+    const frameName = command[0].toUpperCase() + command.slice(1) + 'Frame'
+    const isExtra = Object.keys(extras).includes(frameName)
+    if (isExtra) {
+      MyFrame = extras.SnakeFrame
```
noahmay, line 89: Should be the matching key right?

noahmay (changes requested):
See my comment

src/browser/modules/Stream/FrameContainer.tsx
OskarDamkjaer: Good catch, thanks!

noahmay (approved):

PR #1794, opened 2022-08-08
Makes sidebar resizable when at the guides tab

noahmay:
Adds the option to resize sidebar when the guides tab is open. Also replaces the library used for resizing (previously used with the node inspector). The new library has better performance since it directly modifies the DOM and then updates the state.


src/browser/components/TabNavigation/Navigation.tsx
```diff
@@ -172,10 +175,11 @@ class Navigation extends Component<NavigationProps, NavigationState> {
 
     const drawerIsVisible = this.state.transitionState !== Closed
 
-    const drawerWidth =
+    const guide_drawer_selected =
```
FallingCeilingS, line 178: Any reason why it's lower-kebab case?

FallingCeilingS (approved):
Nice improvement!
I only have a small question.

src/browser/components/TabNavigation/Navigation.tsx
noahmay: No, it's my mistake. Fixing!

FallingCeilingS:
@noahmay - please feel free to merge once all checks turn green!

PR #1796, opened 2022-08-10
Visualizer zooms to fit after initial render is default behaviour

noahmay:
Adds a wait until the initial graph render is done and then zooms to fit the graph. The wait is because of the initial wiggle of the nodes, if we do not wait the zoom to fit can get the incorrect scale.


noahmay:
Closed since #1801 solves the same issue

PR #1799, opened 2022-08-17
Make sure fit to graph always work

noahmay:
Add resize observer to visualizer to make sure zoom to fit works when window/container changes size.


noahmay:
[Fit to graph doesn't have correct bounds when frame is fullscreen](https://trello.com/c/76mbhjW6/77-fit-to-graph-doesnt-have-correct-bounds-when-frame-is-fullscreen)

OskarDamkjaer (approved):

PR #1804, opened 2022-08-22
Updating :help articles about cypher calls to CREATE/DROP INDEX/CONSTRAINT

BjornMagn-Neo4j:
The commands for CREATE/DROP INDEX/CONSTRAINT have been updated in newer releases of neo4j. Updated `:help` articles with these new commands. Also updated use of the commands in some play guides, sidebar guides and sample scripts -> Basic Queries.

I have used the following article to see what statements have been replaced with and from what versions: https://neo4j.com/docs/cypher-manual/current/deprecations-additions-removals-compatibility/

Some things I an unsure about:
* Should I write tests for this? I don't see that we test help pages or saved scripts anywhere but I might be wrong.
* The article about CREATE CONSTRAINT is rather long since it has 3 code examples for both new and old versions. See attached picture. Oskar had an Idea about only showing the relevant examples depending on which version the user is on. Do you have any idea how to do this? I was thinking maybe I could combine this with collapsable lists and having the correct list open by default.
* CREATE CONSTRAINT has changed syntax in both 4.3 and 4.4, and also minor changes in the versions before. I thought it would be too verbose to state the syntax of all possible combinations, so I just made examples with the newest and oldest syntax.
* I am not an expert at CYPHER so it would be nice if you could sanity check my statements.

Screenshot 2022-08-23 at 09 43 59


BjornMagn-Neo4j:
https://trello.com/c/zjaxoWVX/1999-create-index-on-deprecated

eijawerner (changes requested):
Nice to see docs updated! ⭐ Writing down our discussion here for reference: As we discussed we said we should only show the docs for the version that the user is on. Also the play guides needs check which version the user is on and adjust according to that. Regarding the tests I think it is fine for now to only add some test in static-scripts.test.tsx.

BjornMagn-Neo4j:
I have now fixed so that only the help doc for the user version is displayed. I did this using a new component VersionConditionalDoc which I also wrote tests for. Also renamed the help pages so that the are easier to find. 
Discussed with Oskar and apparently for play guides and sidebar guides a decision has been made to assume that the user is on the newest version, because we would like the guides to be possible to have as static markdown.
I don't think I need to add any tests for static-scripts because the functionality for version ranging I use is already tested.

eijawerner (approved):
Nice! ⭐

PR #1807, opened 2022-08-24
Fix credential timeout not working unless User interacts after :server connect

BjornMagn-Neo4j:
Fixes a bug in the automatic credential timeout, where if the user logs in to a server which has set `browser.credential_timeout=X` but then proceeds to do nothing, the user will never be logged out.

This fix manually triggers a credential timeout after the server settings have been loaded, meaning that the timer that counts down to log out the user will start as intended.


BjornMagn-Neo4j:
https://trello.com/c/JZOV0CSc/1889-credential-timeout-not-happening

OskarDamkjaer (approved):

PR #1808, opened 2022-08-24
Remove deprecated procedures

noahmay:
db.indexes() and db.constraints() have been deprecated in Neo4j 4.2 and will be removed in a future release. This pr replaces calls to those procedures with SHOW INDEXES and SHOW CONSTRAINTS.

### DIFF
--- a/src/browser/modules/Stream/SchemaFrame.tsx
+++ b/src/browser/modules/Stream/SchemaFrame.tsx
@@ -90,15 +90,43 @@ const Indexes = ({ indexes, neo4jVersion }: IndexesProps) => {
   )
 }
 
-const Constraints = ({ constraints }: any) => {
-  const rows = constraints.map((constraint: any) => [
-    replace(constraint.description, 'CONSTRAINT', '')
-  ])
+const Constraints = ({
+  constraints,
+  neo4jVersion
+}: {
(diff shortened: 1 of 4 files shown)


src/browser/modules/Stream/SchemaFrame.tsx
```diff
@@ -163,13 +192,17 @@ export class SchemaFrame extends Component<any, SchemaFrameState> {
...
-  fetchData() {
+  fetchData(neo4jVersion: string) {
     if (this.props.bus) {
       // Indexes
       this.props.bus.self(
         CYPHER_REQUEST,
         {
-          query: 'CALL db.indexes()',
+          query:
+            semver.valid(neo4jVersion) &&
```
OskarDamkjaer, line 202: There is a a `getSemanticVersion` getter in `dbMetaDuck` you can use instead of `getRawVersion`. You will still have to do some checks, but it calls `semver.coerce` which I think is could be needed for some odd version numbers that aura uses

OskarDamkjaer: I'll check that it works properly against my aura instance :+1:

OskarDamkjaer (approved):

PR #1823, opened 2022-09-13
Export parse functions from cypher editor support library in arc

OskarDamkjaer:


eijawerner (approved):

src/neo4j-arc/cypher-language-support/index.ts
```diff
@@ -26,6 +26,12 @@ export {
   initalizeCypherSupport,
   getText
 } from './cypher-editor/editorSupport'
+export {
+  parse,
```
eijawerner, line 30: parse is not really a describing name, maybe rename to something more descriptive?

PR #1824, opened 2022-09-15
bugfix missing tooltip on fa icons

eijawerner:
Tooltip was missing on question icon in sysinfo frame, now it works again.

Screenshot 2022-09-15 at 11 03 04


eijawerner:
[Sysinfo cluster members question icon is missing tooltip](https://trello.com/c/Rr8WAlYD/168-sysinfo-cluster-members-question-icon-is-missing-tooltip)

PR #1829, opened 2022-09-27
Better cypher stringify

OskarDamkjaer:
Only adds a new method for string-ification, doesn't remove the old "normal" stringifyMod in browser yet.

Our normalization of durations led to odd behaviour with Duration.inDays, so we opted to stop normalizing the values.

Another change is the display strings with "" so -> `"myString"` and stop doing it for temporal values.


eijawerner (approved):
⭐ Nice! Good with lots of test, and much simpler function. 👍

PR #1837, opened 2022-10-13
Add missing space between `connected` and `to` in connection confirma…

### LINKED ISSUE #1834: Missing space in "You are connected" message
I observe a missing space in the connection message:

Screenshot from 2022-10-10 14-29-41

this is due to a trailing space missing in https://github.com/neo4j/neo4j-browser/blob/master/src/browser/modules/Stream/Auth/ConnectedView.tsx#L44 , which when gets (possibly) concatenated with the host, it gets stitched together into `connectedto`. 

I can go on and submit a PR adding that space, unless there is a reason for that not being done? :)

stefano-ottolenghi:
Fix #1834 .

### DIFF
--- a/src/browser/modules/Stream/Auth/ConnectedView.tsx
+++ b/src/browser/modules/Stream/Auth/ConnectedView.tsx
@@ -41,7 +41,7 @@ const ConnectedView = ({
           <br />
         </span>
       ) : (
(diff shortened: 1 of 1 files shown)


OskarDamkjaer (approved):
Thanks!

PR #1840, opened 2022-10-24
Separate metadata count query

noahmay:
Suppose the metadata count query takes a long time then switch off the automatic refresh for that data and display a warning in the sidebar.  It also fixes the transitions when opening/closing the sidebar.

Screenshot 2022-10-24 at 10 08 30

### DIFF
--- a/src/shared/modules/dbMeta/dbMetaEpics.ts
+++ b/src/shared/modules/dbMeta/dbMetaEpics.ts
@@ -40,7 +40,7 @@ import {
   VERSION_FOR_CLUSTER_ROLE_IN_SHOW_DB,
   isOnCluster,
   updateCountAutomaticRefresh,
-  getCondAutomaticRefreshEnabled,
+  getCountAutomaticRefreshEnabled,
   DB_META_FORCE_COUNT,
   DB_META_COUNT_DONE,
   metaCountQuery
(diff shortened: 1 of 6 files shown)


src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -338,6 +368,35 @@ export const dbMetaEpic = (some$: any, store: any) =>
         .do(() => switchToRequestedDb(store))
         .mapTo({ type: DB_META_DONE })
     )
+export const dbCountEpic = (some$: any, store: any) =>
+  some$
+    .ofType(DB_META_DONE)
+    .filter(() => getCountAutomaticRefreshStatus(store.getState()))
```
OskarDamkjaer, line 374: This could perhaps be called something like `getCondAutomaticRefreshEnabled`? (no need to change it)

src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -338,6 +368,35 @@ export const dbMetaEpic = (some$: any, store: any) =>
...
+      Rx.Observable.fromPromise<void>(
+        new Promise(async resolve => {
+          store.dispatch(updateCountAutomaticRefresh({ loading: true }))
+          if (getCountAutomaticRefreshStatus(store.getState())) {
+            const startTime = performance.now()
+            await getNodeAndRelationshipCounts(store)
+            const timeTaken = performance.now() - startTime
+
+            if (timeTaken > 2000) {
+              console.log('turn off auto refresh')
```
OskarDamkjaer, line 387:
```suggestion
```

OskarDamkjaer (approved):

PR #1841, opened 2022-10-25
Composite db didn't show result graph

OskarDamkjaer:
The query to connect result nodes failed, causing the this.autoCompleteCallback never to be called. These changes improve the error handling and adds an e2e test to make sure composite databases results show


eijawerner (approved):
🌟

PR #1858, opened 2022-11-17
Fix relationship stylings not getting applied 

OskarDamkjaer:
The graph change handler did not repaint updated shaft widths or captions. The simplest change is to just call another tick on the visualization since the normal render will update with the new width/caption!

### DIFF
--- a/src/neo4j-arc/graph-visualization/GraphVisualizer/Graph/visualization/Visualization.ts
+++ b/src/neo4j-arc/graph-visualization/GraphVisualizer/Graph/visualization/Visualization.ts
@@ -234,6 +234,7 @@ export class Visualization {
     )
 
     this.forceSimulation.updateRelationships(this.graph)
+    this.render()
   }
 
   zoomByType = (zoomType: ZoomType): void => {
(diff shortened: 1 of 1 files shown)


src/neo4j-arc/graph-visualization/GraphVisualizer/Graph/visualization/Visualization.ts
```diff
@@ -234,6 +234,10 @@ export class Visualization {
     )
 
     this.forceSimulation.updateRelationships(this.graph)
+    // The onGraphChange handler does only repaint relationship color
+    // not width and caption. We work around that by doing an aditional full
+    // render to get the new stylings
```
eijawerner, line 239:
```suggestion
   // The onGraphChange handler does only repaint relationship color
    // not width and caption, since it requires taking into account surrounding data
    // since the arrows have different bending depending on how the nodes are 
    // connected. We work around that by doing an additional full render to get the 
    // new stylings
```

eijawerner (approved):

src/neo4j-arc/graph-visualization/GraphVisualizer/Graph/visualization/Visualization.ts
eijawerner, line 239:
```suggestion
   // The onGraphChange handler does only repaint relationship color
    // not width and caption, since it requires taking into account surrounding data
    // since the arrows have different bending depending on how the nodes are 
    // connected. We work around that by doing an additional full render to get the 
// new stylings
```

PR #1871, opened 2022-12-27
Don't disable automatic node counts unless the count succeeded.

OskarDamkjaer:


src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -400,23 +410,25 @@ export const dbCountEpic = (some$: any, store: any) =>
     .throttle(() => some$.ofType(DB_META_COUNT_DONE))
     .mergeMap(() =>
       Rx.Observable.fromPromise<void>(
-        new Promise(async resolve => {
+        (async () => {
```
OskarDamkjaer, line 413: This change was becuase we're technically creating a promise within a promise with the `async` marker here.

src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -400,23 +410,25 @@ export const dbCountEpic = (some$: any, store: any) =>
...
-            await getNodeAndRelationshipCounts(store)
+          const res = await getNodeAndRelationshipCounts(store)
+
+          const notAlreadyDisabled = getCountAutomaticRefreshEnabled(
+            store.getState()
+          )
+
+          if (
+            res.requestSucceeded &&
+            res.timeTaken > 1000 &&
```
OskarDamkjaer, line 424: I'm not sure whether this threshold makes sens, I think we could lower it even more actually. In my testing I land somewhere between 10-50m and it should be quick regardless of network and database size, since we only measure computing time and the operation should be a table lookup it

src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -198,8 +202,14 @@ async function getNodeAndRelationshipCounts(store: any) {
           relationships
         })
       )
+      const summary = res.summary as ResultSummary
+      return {
+        requestSucceeded: true,
+        timeTaken: summary.resultAvailableAfter.toNumber()
```
OskarDamkjaer, line 208: Before merging we need to test against a big dataset with the RBAC issue so we're sure it still works.

OskarDamkjaer: Tested //noah

PR #1876, opened 2023-01-09
Add mention of count behaviour in manual

OskarDamkjaer:
We've recently added some new functionality to keep browser running smoothly when RBAC is set up.


AlexicaWright (approved):
I've added a designated section for this, as I don't think it really belongs in the settings. If you agree, feel free to merge whenever. Thank you @OskarDamkjaer !

PR #1879, opened 2023-01-13
updates the text table and csv content for cypher frames

noahmay:
Adds labels and types to nodes and relationships respectively in the text table for a cypher frame. This look is more aligned with cypher shell than the previous one which only showed properties.

Screenshot 2023-01-13 at 09 42 49

Screenshot 2023-01-13 at 09 42 59


src/browser/modules/Stream/CypherFrame/helpers.ts
```diff
@@ -473,3 +426,91 @@ function isNeo4jValue(value: any) {
...
+      const properties = Object.entries(value.properties).map(
+        ([key, value]) => {
+          return `${key}: ${recursiveStringify(value)}`
+        }
+      )
+      const propertiesString = properties.length > 0 ? ` {${properties}}` : ''
+
+      return `[${type}${propertiesString}]`
+    } else {
+      console.log('value2', value)
```
eijawerner, line 492: console.log

eijawerner (changes requested):
Nice work! ⭐  See review comments for some details to look over. Also I think for these kind of functions with a lot of logic, and since unit tests are rather cheap to run, it would be good to have more unit tests, apart from the path with multiple nodes we talked about, make sure that all paths in the recordToStringArray and stringifyNodeRelationshipOrPath are covered.

noahmay:
@eijawerner thanks for the review everything is updated now

eijawerner (approved):
⭐

PR #1891, opened 2023-03-06
Upgrade dependencies without breaking changes 

OskarDamkjaer:


OskarDamkjaer:
[First round of dep upgrades](https://trello.com/c/P304aECB/404-first-round-of-dep-upgrades)

PR #1893, opened 2023-03-08
fix ascii table crash on empty array or object

noahmay:
The ascii table currently shows empty arrays or objects as ''. This PR changes this to [] and {} respectively. This also fixes a crash where ascii-data-table does not work with empty strings.

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/helpers.ts
+++ b/src/browser/modules/Stream/CypherFrame/helpers.ts
@@ -428,9 +428,10 @@ function isNeo4jValue(value: any) {
 }
 
 export const recordToStringArray = (record: Record): string[] => {
+  console.log('record', record)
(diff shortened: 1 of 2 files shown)


src/browser/modules/Stream/CypherFrame/helpers.ts
```diff
@@ -428,9 +428,10 @@ function isNeo4jValue(value: any) {
 }
 
 export const recordToStringArray = (record: Record): string[] => {
+  console.log('record', record)
```
OskarDamkjaer, line 431:
```suggestion

```

OskarDamkjaer (approved):

OskarDamkjaer:
[abc](https://trello.com/c/kvbnppjD/1-abc)

PR #1895, opened 2023-03-10
Add :debug-connectivity frame to help debugging bolt connections

OskarDamkjaer:
The error messages we give historically when attempting to connect to neo4j have not been very helpful. This PR is a step to surface errors earlier and attempt to help users resolve connection issues

### DIFF
--- a/src/browser/modules/Stream/Auth/ConnectForm.tsx
+++ b/src/browser/modules/Stream/Auth/ConnectForm.tsx
@@ -86,8 +86,8 @@ export type HttpReachablity =
   | { status: 'parsingJsonFailed'; error: Error }
   | { status: 'foundBoltPort' }
   | {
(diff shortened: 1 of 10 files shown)


src/browser/modules/Stream/Auth/ConnectForm.tsx
```diff
@@ -75,6 +79,70 @@ interface ConnectFormProps {
...
+  }
+
+  let json
+  try {
+    json = await res.json()
+  } catch (error) {
+    return { status: 'parsingJsonFailed', error: error as Error }
+  }
+
+  if ('auth_config' in json && 'oidc_providers' in json.auth_config) {
```
noahmay, line 117: I think it would be easier to read with early returns

noahmay (approved):

PR #1902, opened 2023-03-22
Reset expandedNodeMap when Graph is reset

Xxy1016:
When we [reset a Graph](https://github.com/neo4j/neo4j-browser/blob/master/src/neo4j-arc/graph-visualization/models/Graph.ts#L201), we see that the `expandedNodeMap` was not reset. We think we should clear all instance variables while resetting the graph. So this PR adds the reset to the `expandedNodeMap`

### DIFF
--- a/src/neo4j-arc/graph-visualization/models/Graph.ts
+++ b/src/neo4j-arc/graph-visualization/models/Graph.ts
@@ -203,6 +203,7 @@ export class GraphModel {
     this._nodes = []
     this.relationshipMap = {}
     this._relationships = []
+    this.expandedNodeMap = {}
   }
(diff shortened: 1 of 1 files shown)


Xxy1016:
Hi, @OskarDamkjaer, thanks for your attention. I have emailed my signed CLA as instruction in [here](https://neo4j.com/developer/cla/#sign-cla). Is there any thing I should do before getting this merged?

OskarDamkjaer (approved):

OskarDamkjaer:
@Xxy1016 Thanks! I'll rerun our tests and merge when they come back green!

PR #1908, opened 2023-04-05
Updating file with docs-base-uri and neo4j-docs-base-uri attribute

lidiazuin:
Some links were broken because this attributed wasn't set before.

### DIFF
--- a/docs/preview.yml
+++ b/docs/preview.yml
@@ -50,5 +50,7 @@ asciidoc:
     sectnums: true,
     sectnumlevel: 3,
     experimental: ''
-    copyright: 2021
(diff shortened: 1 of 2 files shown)


OskarDamkjaer (approved):

noahmay:
Not sure how the docs build is set up but I reckon this addition should be added in the `docs/preview.yml` as well. What do you think @OskarDamkjaer @lidiazuin?

lidiazuin:
You're right @noahmay . I pushed another commit updating the preview.yml file and also fixing the copyright year in both. Thanks for the review @OskarDamkjaer !

Issue #858, opened 2018-10-29
Save and restore a visualization state


mehdibalouchi:
I want to save position of nodes in a visualization somehow and later restore those positions.
consider the user gets a visualization and starts to move nodes on screen, he can `Export PNG` of the visualization. I want to make the user able to save the visualization he made somehow (in a file or something) and when he restores that, all nodes are in the position he saved.
Is there any straight solution you guys can help me with?

oskarhane:
Hi,
the position of elements in the viz is not something we track in neo4j-browser.
What is the user goal here, why do you want this feature?
It sounds like Neo4j Bloom is a better tool for you: https://neo4j.com/bloom/

mehdibalouchi:
> the position of elements in the viz is not something we track in neo4j-browser.

How can I do this? Where is the code that adjusts node positions? I cant't find that. 
I'm trying to find that part of code that adjusts the initial position of nodes in the visualization, so that maybe I can do some hacks and solve my problem.

oskarhane:
Thanks for the additional input here, and a use case @rsworden59!

@mehdibalouchi: The layout is using d3's force layout and happens in this file: https://github.com/neo4j/neo4j-browser/blob/master/src/browser/modules/D3Visualization/lib/visualization/components/layout.js

steveyang331:
it seems don't sloved yet ?

OskarDamkjaer:
This is an interesting idea, but I'm afraid it to be a quite complex feature and not something we can prioritize implementing. If anyone wants to give it a shot and hack together a demo, we could perhaps reconsider

PR #1928, opened 2023-07-24
NodeDrag by Petra Schanz on behalf of DB Systel GmbH

petra-schanz:
This fork implements the feature requested here: [https://github.com/neo4j/neo4j-browser/issues/858](https://github.com/neo4j/neo4j-browser/issues/858) and was written by petra.schanz@deutschebahn.com on behalf of DB Systel GmbH.

Svgs exported by neo4j contain the positions of the drawn nodes. If you show the same nodes in the neo4j-browser you can use the new upload-button (below the zoom buttons because this is all about visualization) and upload your svg: The nodes will be dragged automatically to the positions shown in the svg. You can interact with them and use them as you are used to.

If you need to re-import the database data you should set a uuid before exporting the svg and after importing the data by using something along "MATCH (n) SET n.uuid = n.myEternalAndUniqeAttribute RETURN n"

tempsnip

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,3 +77,16 @@ CYPRESS_BASE_URL=<url to reach the browser to test> (default http://localhost:80
 ```
 
 Example: `CYPRESS_E2E_TEST_ENV="local" CYPRESS_BASE_URL=http://localhost:30000 cypress open --env server=4.2`
+
+## Import svg-File with node positions
+This fork implements the feature requested here: https://github.com/neo4j/neo4j-browser/issues/858
+and was written by petra.schanz@deutschebahn.com
+
+Svgs exported by neo4j contain the positions of the drawn nodes.
+If you show the same nodes in the neo4j-browser you can use the new upload-button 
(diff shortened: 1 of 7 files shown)


petra-schanz:
Hi!
Could you, please, add my cla?
Thank you!

OskarDamkjaer:
@petra-schanz Hello and thanks for your contribution! I added your cla just now.

It was a few years since we discussed if/how we wanted this feature internally so I was unsure if it was still something we were interested in supporting and I forgot to bring it up with our product manager. I did so now and will get back to you! Sorry about the delay.

OskarDamkjaer:
Hello again @petra-schanz, I've spoken with product and I'm sorry to say it was decided we want to keep Browser more focused on quicker iterations rather than doing much work with the visualization results, in part because that use case is better covered by Neo4j Bloom. You can of course still use your work by deploying a your fork of Browser as seen in the [Neo4j Browser manual here](https://neo4j.com/docs/browser-manual/current/deployment-modes/dedicated-web-server/)

PR #1933, opened 2023-09-05
Ensure sso access token is always refreshed on expiry

OskarDamkjaer:
Make sure we catch throw connectivity errors from background calls and add a fallback to refresh the accesstoken in the reconnection flow. 

I also add end to end tests to make sure we don't accidentally break SSO in the future

In addition I solve a state where all queries fail with "client authorized", but Browser will still not trigger its lost connection flow. 

There was a filter put in 6 years ago to not needlessly trigger the connection flow before the user was connected (so don't trigger on authorization errors) and a more general fix was put in place 3 years ago superseded that (only trigger flow if we are supposed to be connected). The first fix is no longer needed, but didn't do any harm - except if your access_token expires and you don't catch the first "TokenExpiredError" and only see the "Unauthorized" errors.


src/shared/services/bolt/bolt.ts
```diff
@@ -182,7 +186,8 @@ function directTransaction(
 
 async function backgroundWorkerlessRoutedRead(
   input: string,
-  { useDb }: { useDb?: string }
+  { useDb }: { useDb?: string },
+  store: any
 ): Promise<QueryResult> {
   const session = getGlobalDrivers()
     ?.getRoutedDriver()
```
OskarDamkjaer, line 193: this also needs to trigger the "on lost connection flow"

PR #1939, opened 2023-09-21
Align documentation by allowing Graph export in JSON & CSV formats

QubitPi:
This PR aligns graph export behavior with [what's being documented](https://neo4j.com/docs/browser-manual/current/operations/export-results/), i.e. exporting in JSON and CSV

page

We like the feature of JSON export very much, because our research needs it a lot. Thanks a lot for developing this option, folks!

PS: I have also signed the CLA (jack20220723@gmail.com, QubitPi). Please let me know if I can help with anything else

### DIFF
--- a/src/browser/modules/Stream/CypherFrame/CypherFrame.tsx
+++ b/src/browser/modules/Stream/CypherFrame/CypherFrame.tsx
@@ -177,9 +177,12 @@ export class CypherFrame extends Component<CypherFrameProps, CypherFrameState> {
     const textDownloadEnabled = () =>
       this.getRecords().length > 0 &&
       this.state.openView &&
-      [ViewTypes.TEXT, ViewTypes.TABLE, ViewTypes.CODE].includes(
-        this.state.openView
-      )
+      [
+        ViewTypes.TEXT,
+        ViewTypes.TABLE,
+        ViewTypes.CODE,
+        ViewTypes.VISUALIZATION
+      ].includes(this.state.openView)
     const graphicsDownloadEnabled = () =>
       this.visElement &&
       this.state.openView &&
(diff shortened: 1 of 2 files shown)


QubitPi:
Hi Neo4J team, I have also signed the CLA (jack20220723@gmail.com). Please let me know if I can help with anything else

OskarDamkjaer:
Hi @QubitPi and thanks for opening this PR! I think it is a good change and I'll have a look at the code now 


I put your name in system now, so the CLA check should be green on when re-run :+1:

src/browser/modules/Stream/CypherFrame/CypherFrame.tsx
```diff
@@ -190,6 +190,8 @@ export class CypherFrame extends Component<CypherFrameProps, CypherFrameState> {
       { name: 'JSON', download: this.exportJSON }
     ]
     const downloadGraphics = [
```
OskarDamkjaer, line 192: This list is meant for the graphical / image downloads and I would like to keep this list with only those two items. To solve the problem of showing the text downloads I think it's better if you extend the `textDownloadEnabled` to say that text download is enabled for when the visualization is open too!

```ts
    const textDownloadEnabled = () =>
      this.getRecords().length > 0 &&
      this.state.openView &&
      [ViewTypes.TEXT, ViewTypes.TABLE, ViewTypes.CODE, ViewTypes.VISUALIZATION].includes(
        this.state.openView
      )
```

I think that should do the trick :)

OskarDamkjaer (changes requested):

QubitPi:
Hi @OskarDamkjaer, thank you very much for the quick feedback. Your comment makes sense and I have change it accordingly

OskarDamkjaer (approved):
Looks good, thanks! I'll merge when we get the tests to pass 💯

PR #1941, opened 2023-09-28
Attempt to refresh SSO token before expiry

OskarDamkjaer:


OskarDamkjaer:
The issue of needing to re-run a query doens't happen often enough to change how the refresh flow works, at least for now

PR #1968, opened 2024-05-22
resetting memory credentials after disconnect

nnecla:
This pull request is to reset the credentials stored in the application memory on disconnect from server.


src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -652,6 +656,7 @@ export const disconnectEpic = (action$: any, store: any) => {
     .merge(action$.ofType(USER_CLEAR))
     .do(() => bolt.closeConnection())
     .do(() => store.dispatch(resetUseDb()))
+    .do(() => resetMemoryCredentials())
```
OskarDamkjaer, line 659: If the user has the setting enabled to store credentials in localstorage, I think it would be fine to not clear the password in memory for convenience. At the same time though, I think it's reasonable if you've run `:server disconnect` (automatically from the connection dropping or manually) to expect the password to no longer be stored (when I tried it out it's cleared both in memory and localstorage after a `:server disconnect`)

I think it'd be good to ask Greg / PM how he wants it to work

nnecla: As far as I've seen, for the enabled localStorage credentials case, it really does not matter if we reset the memory credentials or not as it uses the data from the localStorage anyways to fill the connection form, if it exists.

We can of course prevent password to be removed if the setting is enabled both for the memory and for the localStorage, but this is different than the current prod experience. Let me check with Greg on this one as you've mentioned.

nnecla: Talked with Greg, and as in the feature enabled case it clears the password from the local storage on disconnect, we will be also clearing the password from the application memory, same as the not enabled case.

src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -193,7 +193,7 @@ export function getConnectionData(
     return { ...data, username: memoryUsername, password: memoryPassword }
   }
 
-  return data
+  return { ...data, username: data.username ? data.username : memoryUsername }
```
OskarDamkjaer, line 196:
```suggestion
  return { ...data, username: data.username ?? memoryUsername }
```

OskarDamkjaer: could be written like this too I think

OskarDamkjaer (approved):

src/shared/modules/connections/connectionsDuck.ts
nnecla: Unfortunately, there are cases where `data.username` is an empty string. So, it won't be working as expected as `??` checks the value being undefined or null.

PR #1974, opened 2024-07-29
Fix bug where visualizations were unable to load neighbours/relationships in a cluster environment

daveajrussell:
A client was experiencing an issue expanding a node's neighbours and relationships due to routing being disabled on the requests that these functions were executing.

In this change, we are updating the functions in the vizualization component to use routed requests to query for a node's relationships and neighbours.


OskarDamkjaer (approved):
@daveajrussell I looked at the `License Compliance` check, it was a false positive. You might need to push a new commit to get CI to rerun the check. 

Also the PR titles in Browser get used for the user facing changelogs, so it'd be good to re-write it so that it's clear what we did from a user perspective. Perhaps something like `Fix bug where visualization couldn't find neighbors in cluster environment` or similar

PR #1975, opened 2024-08-01
Ensure node expansion runs against the correct database

OskarDamkjaer:
Using the `database` field on the result ensures that the right database is used for node expansion, even if the query dynamically switched database, for example like this:

```
:use neo4j
USE movies MATCH (n) RETURN n;
```


nnecla (approved):
Looks good to me!

daveajrussell (approved):
I've repro-d on my neo4j desktop instance

Tested out the same steps on this and it's looking good

PR #1978, opened 2024-10-10
adding advertisement tile for the new browser ui

nnecla:
## Description

This pr is to add the `PreviewFrame` which contains the advertisement tile for the new browser UI. This frame will only be shown if the new UI is there, and the feature flag is on.

Also, if user prefers the new UI, we switch to the new UI before the app renders.

The look of the frame is not final and wip.


nnecla:
[add tile in :play start](https://trello.com/c/3IqNtThZ/1002-add-tile-in-play-start)

nnecla:
[Remember last used Query/Browser](https://trello.com/c/2bZZ8oK7/970-remember-last-used-query-browser)

src/browser/index.tsx
```diff
@@ -24,4 +24,19 @@ import AppInit, { setupSentry } from './AppInit'
 import './init'
 
 setupSentry()
-ReactDOM.render(<AppInit />, document.getElementById('mount'))
+;(async () => {
+  const doesPreferQuery = localStorage.getItem('prefersOldBrowser') === 'false'
+
+  const response = await fetch('/preview/manifest.json')
```
OskarDamkjaer, line 30: I think `fetch` can throw, so I think it'd be good with  try { } catch { } here

src/browser/index.tsx
```diff
@@ -24,4 +24,19 @@ import AppInit, { setupSentry } from './AppInit'
...
+  if (response.status === 200) {
+    if (doesPreferQuery) {
+      window.location.pathname = '/preview/'
```
OskarDamkjaer, line 33: In the bundled Browser case, the preview will be on `http://localhost:7474/browser/preview` and in the hosted case it's on `browser.neo4j.io/preview`.

So this would not work for bundled browser

src/browser/index.tsx
OskarDamkjaer: This also needs to be `fetch('./preview/manifest.json')` to handle the bundled case where it's under `/browser/preview`. And also to handle if users have set up their own custom hosting!

src/shared/modules/experimentalFeatures/experimentalFeaturesDuck.ts
```diff
@@ -9,13 +9,20 @@ export const showFeature = (state: any, name: any) =>
...
     displayName: 'Show experimental features',
     tooltip: 'Show feature section in settings drawer'
+  },
+  [experimentalFeaturePreviewName]: {
+    name: experimentalFeaturePreviewName,
+    on: false,
+    displayName: 'Advertise preview of new browser',
+    tooltip: 'Enable the advertisement tile of the new browser'
   }
 }
```
OskarDamkjaer, line 27: I think this is a user facing feature flag, I think we're better off to just have it on always if `isPreviewAvailable`

nnecla: Removed it!

src/browser/index.tsx
nnecla: Perfect point, changed it accordingly!

PR #1983, opened 2024-10-23
Bump neo4j-driver to latest version, and update all requisite dependencies

daveajrussell:
This PR bumps the version of the neo4j javascript driver to the latest version, 5.26.0.

I am also updating a number of requisite dependencies to accommodate the driver version bump, these I have annotated separately.


package.json
```diff
@@ -117,22 +117,22 @@
...
+    "eslint-plugin-react-hooks": "^4.6.2",
     "express": "^4.15.4",
     "file-loader": "^2.0.0",
     "fork-ts-checker-notifier-webpack-plugin": "^3.0.0",
     "fork-ts-checker-webpack-plugin": "^5.0.14",
     "html-loader": "^0.5.5",
     "html-webpack-plugin": "^4.5.2",
     "husky": "^0.14.3",
-    "jest": "^25.2.4",
+    "jest": "^26.6.3",
```
daveajrussell, line 135: v25 has a peer dependency of typescript <v4, so we best update the version 

however >v27, we run into a host of problems with jestdom and timers that I was unable to resolve

after v27, the jestdom environment is no longer included in the main jest package, and must be imported separately

as soon as this happens, many tests start throwing errors that `setImmediate` is undefined

there's a lengthy thread here https://github.com/prisma/prisma/issues/8558 with several workarounds

however, of the solutions I've tried, I did not get a successful test run

`global.setImmediate = jest.useRealTimers` caused tests to hang until they timed out
```
import { setImmediate } from 'timers;
global.setImmediate = setImmediate
```

caused any test that relied on RXJS to throw an error that `bind` was undefined

package.json
```diff
@@ -146,18 +146,18 @@
...
     "react-hot-loader": "^4.13.0",
     "react-refresh": "^0.11.0",
     "react-test-renderer": "^17.0.2",
     "redux-mock-store": "^1.2.3",
     "style-loader": "^0.23.1",
-    "ts-jest": "^25.0.0",
+    "ts-jest": "^26.5.6",
     "ts-loader": "^8.0.2",
-    "typescript": "^3.9.5",
+    "typescript": "^4.6.3",
```
daveajrussell, line 160: introducing the latest driver version also introduced some type errors at compile time

here, I've set the version of typescript to match that of the neo4j js driver

package.json
```diff
@@ -117,22 +117,22 @@
     "cross-env": "^6.0.3",
     "css-loader": "^1.0.0",
     "cypress": "^8.3.1",
-    "eslint": "^7.7.0",
+    "eslint": "^8.57.1",
```
daveajrussell, line 120: the version of eslint we were using similarly had a peer dependency on typescript <v4

v9 has some breaking changes with eslint config, again, something we should address, but to keep the change surface low, I've gone with v8

src/neo4j-arc/common/types/cypherDataTypes.ts
```diff
@@ -72,7 +72,7 @@ export type CypherBasicPropertyType =
   | number
   | string
   | Integer
-  | BigInt
+  | bigint
```
daveajrussell, line 75: `eslint --fix` introduced this change

.eslintrc.json
daveajrussell, line 1: following the eslint upgrade, the following rules started piping up - these additions add around 300 new warnings, for a total of ~3900

src/shared/utils/typeguards.ts
daveajrussell, line 1: adding these typeguards to satisfy the Error type requirements that began to error as a result of the typescript upgrade

package.json
OskarDamkjaer: If we'd want to invest in the build setup for browser I think it'd make sense to go to vite & vitest, but I don't think it's worth the time

OskarDamkjaer (approved):
Really nice work with all the version fiddling ⭐

PR #1984, opened 2024-10-25
Add analytics events for browser preview

daveajrussell:
Adds a number of events to track page loads and UI switch actions for the browser preview.

When the browser app is loaded, we track an analytics event, PREVIEW_PAGE_LOAD with properties { previewUI: boolean, hasTriedPreviewUI: boolean }

`hasTriedPreviewUI` is stored when we first switch to the preview UI.

When we use the welcome tile to switch to the preview, we track an additional analytics event, PREVIEW_UI_SWITCH with properties { switchedTo: 'browser' | 'preview'; timeSinceLastSwitch: number }.

We track timeSinceLastSwitch using a timestamp stored in localstorage, which we reset when switching between UIs.

There is a separate PR in the browser repo for tracking events within browser: https://github.com/neo4j/upx/pull/4086


src/shared/modules/udc/udcDuck.ts
```diff
@@ -324,3 +325,13 @@ export const trackErrorFramesEpic: Epic<Action, GlobalState> = (
...
+
+export const trackPreviewEpic: Epic<Action, GlobalState> = action$ => {
+  return action$.ofType(PREVIEW_EVENT).map((action: any) => {
+    return metricsEvent({
+      category: 'preview',
+      label: action.label,
+      data: action.data
+    })
+  })
+}
```
daveajrussell, line 337: adding this epic for tracking preview events here so that we can consume them in the existing subscriber in the App and dispatch to segment

src/shared/modules/preview/previewDuck.ts
daveajrussell, line 1: Adding this previewDuck module to create redux actions for our preview switch and page load events, containing logic for setting properties of each action

src/browser/modules/App/App.tsx
```diff
@@ -149,6 +150,11 @@ export function App(props: any) {
     props.bus && props.bus.send(initAction.type, initAction)
   }, [props.bus])
 
+  useEffect(() => {
+    const pageLoadAction = trackPageLoad()
+    props.bus && props.bus.send(pageLoadAction.type, pageLoadAction)
+  }, [props.bus, props.telemetrySettings.allowUserStats])
```
OskarDamkjaer, line 156: I think it shouldn't be a problem, but I still worry a little about this effect running extra times and giving us bad stats. I think we could be certain to avoid that if we run dispatch in `dbMetaEpics` instead, in the section where the server configuration finishes. There's a comment: `// side-effects` where it'd fit well

daveajrussell: thanks! will take a look

src/browser/modules/Stream/StartPreviewFrame.tsx
```diff
@@ -86,3 +101,15 @@ export const PreviewFrame = () => {
...
 }
+
+const mapDispatchToProps = (dispatch: Dispatch<Action>) => {
+  return {
+    executeTrackNavigateToPreview: () => dispatch(trackNavigateToPreview())
+  }
+}
+
+export default withBus(
+  connect(() => {
```
OskarDamkjaer, line 112: I think you can just pass null here instead of a placeholder function

src/shared/modules/dbMeta/dbMetaEpics.ts
```diff
@@ -546,6 +547,12 @@ export const serverConfigEpic = (some$: any, store: any) =>
             store.dispatch(triggerCredentialsTimeout())
           }
 
+          setTimeout(() => {
```
OskarDamkjaer, line 550: interesting that it works even with a timeout of `0`

OskarDamkjaer (approved):

src/shared/modules/dbMeta/dbMetaEpics.ts
daveajrussell: indeed - it's where I start to get a bit iffy with the eventloop/callback queue - but I guess the callback just going to the back of the queue is enough for react's internals to finish and resolve props in the app

PR #1990, opened 2024-12-16
Adding support for the calendar versioning

nnecla:
This pr is to add support for the calendar versioning, which Neo4j plans to use starting in January. Versions are expected to have zero paddings in the month, and no zero paddings in the patch numbers. Thus, the first calver version that neo4j will be using is expected to be `2025.01.0`.


nnecla:
[handle calendar based versions in old & new Browser (before jan)](https://trello.com/c/cujFH0hJ/1022-handle-calendar-based-versions-in-old-new-browser-before-jan)

OskarDamkjaer (approved):

PR #1992, opened 2025-02-04
Updates force password change flow to use Cypher first, falling back to dbms function call

daveajrussell:
This PR updates the force password change flow that users go through when their password requires resetting upon login. 

Currently, the flow will use the app's detected version to determine whether to use the dbms function call if the app is on a version lower than 4.0, otherwise it will use the Cypher query.

This is problematic since the server's version will not be in state if this is a first login attempt, so it will try and query the database to get the server version.

Only the user is not authorized to query the database, because their password requires a reset.. so we cannot use the server version to accurately determine which version of the query to use.

If we still don't have a version at this point, we fallback to using a function on the driver to determine if there is multidb support - if there is then we use the Cypher query, otherwise we will use the dbms function call.

If, for whatever reason, the server is running >4.0 and multidb returns false, the legacy query is used, and will fail since it does not exist.

In this refactored approach, we no longer attempt to determine the version, and simply attempt to execute the Cypher query, falling back to the dbms function call if a specific error is encountered executing the Cypher - a much more straightforward approach, with less margin of error.

There is also a small update in the bolt service to target the system database when verifying connectivity.

This will resolve a problem some users encounter upon logging in when the default `neo4j` database is unavailable.. deleted or otherwise.

**Repro**
- Run a version 3 server
- Drop the default neo4j database
- Create a user requiring a password change`CALL dbms.security.createUser('user', 'changeme', true)`
- Log in as this user - you will be prompted to change your password - change the password
- You should be able to log in successfully
---
- Run a version >4.0 server
- Drop the default neo4j database
- Create a user requiring a password change `CREATE USER user IF NOT EXISTS SET PASSWORD 'changeme' CHANGE REQUIRED`
- Log in as this user - you will be prompted to change your password - change the password
- You should be able to log in successfully

### DIFF
--- a/src/shared/modules/connections/connectionsDuck.ts
+++ b/src/shared/modules/connections/connectionsDuck.ts
@@ -951,7 +951,7 @@ export const handleForcePasswordChangeEpic = (some$: any) =>
         if (!action.$$responseChannel) return Rx.Observable.of(null)
 
         return new Promise(resolve => {
-          const resolveResponse = (error?: Error) => {
+          const resolveAction = (error?: Error | void) => {
             resolve({
               type: action.$$responseChannel,
               success: error === undefined,
@@ -976,28 +976,30 @@ export const handleForcePasswordChangeEpic = (some$: any) =>
             )
             .then(async driver => {
               try {
+                // Attempt to change the password using Cypher syntax
                 const result = await forceResetPasswordQueryHelper
                   .executeAlterCurrentUserQuery(driver, action)
-                  .then(() => resolveResponse())
+                  .then(resolveAction)
                   .catch(error => error)
 
                 if (isError(result)) {
                   if (result instanceof MultiDatabaseNotSupportedError) {
-                    // If we get a multi database not supported error, fall back to the legacy function
+                    // If we get a multi database not supported error,
+                    // fall back to the legacy dbms function
                     await forceResetPasswordQueryHelper
                       .executeCallChangePasswordQuery(driver, action)
-                      .then(() => resolveResponse())
-                      .catch(resolveResponse)
+                      .then(resolveAction)
+                      .catch(resolveAction)
                   } else {
                     // Otherwise, return the error for the UI to handle e.g. invalid password
-                    resolveResponse(result)
+                    resolveAction(result)
                   }
                 }
               } finally {
                 driver.close()
               }
             })
-            .catch(resolveResponse)
+            .catch(resolveAction)
         })
       }
     )
--- a/src/shared/modules/connections/connectionsDuck.ts
+++ b/src/shared/modules/connections/connectionsDuck.ts
@@ -958,10 +958,8 @@ export const handleForcePasswordChangeEpic = (some$: any) =>
             )
             .then(async driver => {
               try {
-                const supportsMultiDb = await driver.supportsMultiDb()
-
                 const res = await forceResetPasswordQueryHelper
-                  .executeAlterCurrentUserQuery(driver, action, supportsMultiDb)
+                  .executeAlterCurrentUserQuery(driver, action)
                   .then((res: any) => {
                     resolve({
                       type: action.$$responseChannel,
@@ -981,11 +979,7 @@ export const handleForcePasswordChangeEpic = (some$: any) =>
                     /(Invalid input 'A': expected <init>)/.test(res.message)
                   ) {
                     await forceResetPasswordQueryHelper
-                      .executeCallChangePasswordQuery(
-                        driver,
-                        action,
-                        supportsMultiDb
-                      )
+                      .executeCallChangePasswordQuery(driver, action)
                       .then((res: any) => {
                         resolve({
                           type: action.$$responseChannel,
--- a/src/shared/modules/connections/connectionsDuck.ts
+++ b/src/shared/modules/connections/connectionsDuck.ts
@@ -39,6 +39,8 @@ import {
 import { NEO4J_CLOUD_DOMAINS } from 'shared/modules/settings/settingsDuck'
 import { isCloudHost } from 'shared/services/utils'
 import { fetchMetaData } from '../dbMeta/dbMetaDuck'
+import { isError } from 'shared/utils/typeguards'
+import forceResetPasswordQueryHelper from './forceResetPasswordQueryHelper'
 
 export const NAME = 'connections'
 export const SET_ACTIVE = 'connections/SET_ACTIVE'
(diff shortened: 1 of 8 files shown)


src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -919,3 +922,106 @@ export const retainCredentialsSettingsEpic = (action$: any, store: any) => {
...
+                      }
+                    })
+                  })
+                  .catch(e => {
+                    return e
+                  })
+
+                if (isError(res)) {
+                  if (
+                    /(Invalid input 'A': expected <init>)/.test(res.message)
```
daveajrussell, line 981: this feels brittle.. is there a better way to determine if the Cypher call failed.. it'd be nice to have a GQL status code, but this currently returns the generic code

src/shared/services/bolt/globalDrivers.ts
```diff
@@ -62,7 +62,7 @@ export const buildGlobalDriversObject = async (
         opts,
         () => {}
       )
-      routed && (await routed.verifyConnectivity())
+      routed && (await routed.verifyConnectivity({ database: 'system' }))
```
daveajrussell, line 65: this resolves the issue upon login where the neo4j database does not exist.. verifyConnectivity will internally default to the neo4j database if the database arg is not provided.. here we use the system database to verify connectivity to the database.

src/shared/modules/connections/forceResetPasswordQueryHelper.ts
```diff
@@ -0,0 +1,85 @@
...
+      driver,
+      payload,
+      userActionTxMetadata.txMetadata,
+      supportsMultiDb ? { database: SYSTEM_DB } : undefined
+    )
+  },
+  /**
+   * Executes a query to change the user's password using legacy DBMS function
+   * for versions of Neo4j older than 4.0
+   */
```
OskarDamkjaer, line 67: versions under 4 can't have multi db

src/shared/modules/connections/forceResetPasswordQueryHelper.ts
```diff
@@ -0,0 +1,85 @@
...
+          throw e
+        })
+        .finally(() => {
+          session.close()
+        }))
+  },
+  /**
+   * Executes a query to change the user's password using Cypher available
+   * on Neo4j versions since 4.0
+   */
```
OskarDamkjaer, line 43: all databases after 4 have multidb

src/shared/modules/connections/forceResetPasswordQueryHelper.ts
```diff
@@ -0,0 +1,85 @@
...
+        {
+          metadata
+        }
+      )
+        .catch(e => {
+          throw e
+        })
+        .finally(() => {
+          session.close()
+        }))
```
OskarDamkjaer, line 38: since it's always writes and never session.run, could we just use `executeWrite` here?

src/shared/modules/connections/connectionsDuck.ts
```diff
@@ -919,3 +922,106 @@ export const retainCredentialsSettingsEpic = (action$: any, store: any) => {
...
+    .ofType(FORCE_CHANGE_PASSWORD)
+    .mergeMap(
+      (
+        action: Connection & { $$responseChannel: string; newPassword: string }
+      ) => {
+        if (!action.$$responseChannel) return Rx.Observable.of(null)
+
+        return new Promise(resolve => {
+          bolt
+            .directConnect(
```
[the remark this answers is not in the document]
daveajrussell: let me check into this

src/shared/modules/connections/forceResetPasswordQueryHelper.ts
daveajrussell: I will check in to this

src/shared/modules/connections/connectionsDuck.ts
daveajrussell: the reason I put this in is so that errors from the cypher call are presented, and we don't try the dbms call if the user has, for example, tried to set an invalid password.. we'd then show an error that a procedure doesn't exist, rather than the password not meeting requirements

src/shared/modules/connections/forceResetPasswordQueryHelper.ts
daveajrussell: yes - `executeWrite` works here, I've replaced it.

src/shared/modules/connections/forceResetPasswordQueryHelper.ts
daveajrussell: I am now always passing undefined for queries using the `CALL dbms` method

src/shared/modules/connections/connectionsDuck.ts
daveajrussell: so something that I've thought about making other changes, is now that we're attempting the Cypher query assuming that it will always support multi databases, is we get back a specific error from the JS driver..

`Driver is connected to the database that does not support multiple databases. Please upgrade to neo4j 4.0.0 or later in order to use this functionality`

I was thinking, we could create an error class for this in the query helper (e.g. `class MultiDatabaseNotSupportedError extends Error`, then check for this message when we try to execute a query, throwing an instance of `MultiDatabaseNotSupportedError` if we get this error message.

Then when we check the response we could do

```
if (isError(res)) {
   if (res instanceof MultiDatabaseNotSupportedError) {
      // fallback
   } else {
      // show other errors in the UI, e.g. password requirement error
   }
}
```

wdyt?

src/shared/modules/connections/connectionsDuck.ts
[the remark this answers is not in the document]
OskarDamkjaer: Using the protocol selected is good, thanks for confirming!

src/shared/modules/connections/connectionsDuck.ts
OskarDamkjaer: >the reason I put this in is so that errors from the cypher call are presented, and we don't try the dbms call if the user has, for example, tried to set an invalid password.. we'd then show an error that a procedure doesn't exist, rather than the password not meeting requirements

ah of course. Interesting. I think the approach you outlined makes sense. The other option is to do a `supportsMultidatabase` call before and use the old/new syntax depending on that. It's a little problematic though because some old version (4.1, 4.2) are reported as not supporting multi database after they went out of support...

daveajrussell: 😬 

okay cool, thanks for confirming! 

I've pushed the change and streamlined the epic slightly, less repetition of certain bits and added some accompanying comments

OskarDamkjaer (approved):

PR #1994, opened 2025-02-04
passing dbms and db info while passing to the preview experience

nnecla:
### Description
This pr is to pass the db and the dbms connection information to the preview experience.


nnecla:
[Add `db` url param when switching to new preview](https://trello.com/c/ZaK4pMzO/1183-add-db-url-param-when-switching-to-new-preview)

daveajrussell (approved):
I wonder if there's anything we can do anything for the automatic redirect that we do from `index.tsx` if the user has the `prefersOldBrowser` flag set to false 🤔 

I don't think the store has been initialized at that point.. but I wonder if we could manually look into localstorage at the `neo4j.connections` data, if there's an active connection, we could try and extract this and use it when redirecting? (I don't know if this is a horrible idea or not)

Screenshot 2025-02-05 at 10 05 39

PR #2000, opened 2025-03-27
Update @neo4j-devtools/word-color package version

daveajrussell:
In this change, we update the `@neo4j-devtools/word-color` package version from 0.0.7 to 0.0.8.

The 0.0.7 version is missing license information.

There are a few small changes required to facilitate this update;
- configure word-color to use babel-loader (in 0.0.8, there is a change that includes a nullish coalescing operator which needs transpiling).
- change import path for `DetailsPaneProps` to resolve a type mismatch.


nnecla (approved):
LGTM!

PR #2001, opened 2025-03-28
Rollout of new Browser experience as default.

daveajrussell:
We wish to execute a controlled rollout of the new Browser as the default experience.

This change will opt in all users by default, excluding those who have previously opted _in_ and then reverted to the old Browser.


src/browser/services/preview-optin-service.ts
```diff
@@ -0,0 +1,55 @@
...
+  'uk-ua' // Ukrainian (Ukraine)
+]
+
+const OPTED_IN_LOCALES = [...EU_LOCALES]
+
+const userHasNotOptedOutOfPreview = (): boolean => {
+  const prefersOldBrowser = localStorage.getItem('prefersOldBrowser')
+  const doesPreferQuery = prefersOldBrowser === 'false'
+  return doesPreferQuery || prefersOldBrowser === null
+}
```
daveajrussell, line 48: ensure that users who have tried the preview, but have clicked the 'Go back to old Browser' link are not opted in again.

OskarDamkjaer (approved):

PR #2002, opened 2025-03-28
Remove neo4j-arc build and release configs, workflows

daveajrussell:
Removing package.json/yarn.lock/tsconfig.json/README and workflow files since this project is no longer released, nor does it need to be built separately for Browser.


nnecla (approved):
Looks good to me! Also, do we still need the license under neo4j-arc?

PR #2004, opened 2025-04-07
Add localhost preview-as-default rollout service

daveajrussell:
We wish to execute a controlled rollout of the new Browser as the default experience.

This change will opt in all users on **localhost** by default, excluding those who have previously opted in and then reverted to the old Browser.


OskarDamkjaer (approved):
Exciting!

PR #2011, opened 2025-07-24
Add link to production version of the new Browser

OskarDamkjaer:


src/browser/modules/Stream/StartPreviewFrame.tsx
```diff
@@ -18,39 +18,28 @@
...
+  if (db) {
     url.searchParams.set('db', db)
   }
 
-  const previewPath = '/preview/'
-  if (!url.pathname.endsWith(previewPath)) {
-    url.pathname = url.pathname.replace(/\/?$/, previewPath)
-    window.location.href = decodeURIComponent(url.toString())
-  }
+  window.open(url.toString(), '_blank')
```
daveajrussell: perhaps not completely necessary since we define `url` ourselves - but we could add `noreferrer` to here to the options parameter as a general sanitary measure

src/browser/modules/Stream/StartPreviewFrame.tsx
```diff
@@ -71,10 +60,8 @@ const PreviewFrame = ({
       <div className="teasers">
         <div className="teaser teaser-advertise teaser-3">
           <img src="./assets/images/clusters.svg" className="img-advertise" />
-          <h3>🚀 Try the new Browser preview!</h3>
-          <p>
-            Switch to the preview experience to access all the latest features.
-          </p>
+          <h3>🚀 Try the new hosted Browser!</h3>
+          <p>Switch to the hosted Browser to access all the latest features.</p>
```
daveajrussell:
```suggestion
          <p>Switch to the hosted Browser to access all of the latest features.</p>
```

daveajrussell (approved):
👍

src/browser/modules/Stream/StartPreviewFrame.tsx
OskarDamkjaer: it's good practice so might as well

dpozinen:
hi @OskarDamkjaer I'm a little bit confused - how can I open the preview ui on a standalone local instance after this change?

daddycocoaman:
> hi @OskarDamkjaer I'm a little bit confused - how can I open the preview ui on a standalone local instance after this change?

+1 on this question @OskarDamkjaer. Is the preview not intended to be used in a local instance? That seems like a step backwards.

